PR #7, opened 2021-03-04
feat(examples): added example for Nhost with React-Query

alveshelio:
This PR adds a new example for Nhost with React-Query


elitan:
Fanstastic!

PR #50, opened 2021-06-18
Change Nextjs Template Directory Name

mrinalwahal:
Moved `templates/nextjs` to `templates/next` as per convention followed with `nuxt` and `react` directories, since the CLI directly parses the repo tree and clones the folder directly.

Small change. Shouldn't break anything.


subatuba21:
Looks good, merging.

PR #53, opened 2021-07-01
adds a self managed, react-app template

yashank09:
This commit renames the current React template to `cra-react`, and adds a new template `react` that provides a self managed app without any workflow dependencies.

The logic behind this change is to provide the users with a managed workflow(CRA), as well as a self managed one for custom configurations(React, Webpack5 and Babel).

One dependency which would need to be changed after this is the CLI, where the URLS will have to be adjusted. I can send a PR for this on your CLI fork if okay @mrinalwahal.

TODO
- [x] Add Nhost-auth


mrinalwahal:
We specifically renamed it from `cra-react` to `react`, because the CLI fundamentally does nothing except parse the git tree for templates directory and clone the same directory based on the framework choice.

Simply using `react` follows the convention by `nuxt` and `next`.

Otherwise if users want to bypass the prompt on CLI they will have to choose `nhost templates -c cra-react` instead of `nhost templates -c react`.

We want to keep the latter since it follows the format with other frameworks. Example: `nhost templates -c nuxt`

yashank09:
Ah I see. The issue is we would like to support a CRA-less React template as well, since most apps need to be 'ejected' at one time anyway. What should we do if we decide to use both the pure React, as well as the CRA managed template?

mrinalwahal:
If you need both pure react and a CRA managed template, then we have no other option than using both `react` and `cra-react` folders.

I'll let @nunopato take the call. And will make the required change on the CLI accordingly.

yashank09:
Maybe we could display a sub-menu on `nhost templates -c react` and let the user decide there instead? I personally think it will be good to support a CRA less template as well.

mrinalwahal:
I don't think that's a good idea. CLIs have their own design philosophy. It's not good practice to make them unpredictable by adding recursive sub-options for a single option. Also, we'll need to handle those sub-options through flags as well, to allow users to bypass those prompts in cases where we don't have access to their host machines.

Using two different folders is a better idea, than this one.

yashank09:
Closing this MR - it will be implemented in future if requested by users.

PR #150, opened 2022-02-07
Monorepo

plmercereau:


examples/react-apollo-crm/package.json
```diff
@@ -1,15 +1,15 @@
...
     "@apollo/client": "^3.4.16",
     "@craco/craco": "^6.4.0",
     "@headlessui/react": "^1.4.2",
     "@heroicons/react": "^1.0.5",
-    "@nhost/nhost-js": "^0.3.4",
-    "@nhost/react-apollo": "^2.0.7-0",
-    "@nhost/react-auth": "^2.0.3",
+    "@nhost/nhost-js": "workspace:*",
+    "@nhost/react-apollo": "workspace:*",
+    "@nhost/react-auth": "workspace:*",
```
elitan, line 12: How will this work for developers who copy the `react-apollo-crm` folder to a separate folder and runs this example independently?

elitan (changes requested):

examples/react-apollo-crm/package.json
plmercereau: corrected. this syntax is only useful to generate version range when publishing packages to npm. Examples are not published as packages therefore don't need this.

package.json
```diff
@@ -0,0 +1,78 @@
...
+  "author": "Nhost",
+  "license": "MIT",
+  "homepage": "https://nhost.io",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/nhost/nhost-js.git"
+  },
+  "bugs": {
+    "url": "https://github.com/nhost/nhost-js/issues"
+  },
```
elitan, line 15: Do we need to update the name, description, and URLs for this package.json?

elitan (changes requested):
The folder name of e.g. `@nhost/hasura-auth-js` is `hasura-auth` (with out the `-js`). 

Should we instead try to keep the folder and package names in sync?

Like this:

`@nhost/nhost-js` -> `packages/nhost-js`
`@nhost/hasura-auth-js` -> `packages/hasura-auth-js`

etc

elitan:
Will packages' versions be updated individually with this monorepo setup?

elitan:
There are a lot of config files in the root directory of the repo.

Is there any chance we can reduce those? Fewer files in the root directory mean less scrolling to the README file :)

CleanShot 2022-02-07 at 09 43 48@2x

package.json
plmercereau: I'm not fond on this `-js` suffix to be honest. Any npm package is meant to be Javascript, this suffix is redundant. I did not want to change the package name as we are not talking about renaming our existing npm packages, although I think we should

plmercereau:
> 


> Will packages' versions be updated individually with this monorepo setup?

Yes, that's the spirit

plmercereau:
> There are a lot of config files in the root directory of the repo.
> 
> Is there any chance we can reduce those? Fewer files in the root directory mean less scrolling to the README file :)
> 
> CleanShot 2022-02-07 at 09 43 48@2x

yes, that's too much. I moved this to a `.config` folder that is populated when the `prepare` script is executed, and added the root files to `.gitignore`

package.json
elitan: Yep I see your point and I agree.

The question is if we want to introduce the mismatch now? Or keep things in sync for now until we make such name change?

plmercereau: I haven't changed the package name and can keep add `-js` suffixes to the folders although it is hurting my eyes :)

elitan: I don't have a particular hard opinion here tbh. Maybe @nunopato wants to chime in?

elitan: > I haven't changed the package name and can keep add `-js` suffixes to the folders although it is hurting my eyes :)

At least we're being consistent! :D

elitan (approved):

package.json
plmercereau: ah ok, I saw your second comment about the root package.json which is somehow incorrect. Pushing changes

PR #214, opened 2022-02-17
Next.js and React rewrite

plmercereau:
- [Documentation preview](https://docs-v2-git-react-next-nhost.vercel.app/get-started)
- [React demo app](https://nhost-react-apollo-demo.surge.sh) (Oauth providers not set yet)
- [Changesets](#issuecomment-1042848739)


docs/content/docs/reference/nextjs/index.mdx
```diff
@@ -0,0 +1,116 @@
+---
+title: 'Introduction'
+---
+
+It is possible to use [`@nhost/react`](/reference/react) in any Next.js page that would be configured to render on the client side.
+
+When rendering a page from the server side, Next.js needs to get some information from the client to determine their authentication status. Such a communication is only available from cookies, and the the Nhost client has been designed to enable such a mechanism.
+
+The cookie is `same-site` and is used only stores the refresh cookie and the expiration date of the JWT so the browser can proactively ask for a new JWT before it expires.
```
szilarddoro, line 9: There is probably a typo here: "is used only stores" should be "is used to store" / "is used only to store"

docs/content/docs/reference/nextjs/index.mdx
```diff
@@ -0,0 +1,116 @@
...
+    activate auth
+    auth-->>browser: New refresh token + JWT + user data
+    deactivate auth
+    browser->>browser: Set the new refresh token in the cookie
+    deactivate browser
+```
+
+## Refreshing the JWT token
+
+For security considerations, the Nhost JS tools never persist the JWT anywhere in the browser either in `localStorage` or in a cookie, and the JWT expires regularily (15 minutes by default). As a result, the Nhost client regularly call the `hasura-auth` service to get a new JWT from a valid refresh tooken. The new JWT is emitted together with a new refresh token, while the old one is invalidated.
```
szilarddoro, line 33: I found some typos here:

"expires regularily" -> "expires regularly"
"As a result, the Nhost client regularly call" -> "As a result, the Nhost client regularly calls"
"refresh tooken" -> "refresh token"

docs/content/docs/reference/nextjs/index.mdx
```diff
@@ -0,0 +1,116 @@
...
+    activate hasura
+    hasura-->>browser: Updated data
+    deactivate hasura
+    browser->>browser: Re-render data
+    deactivate browser
+```
+
+## Loading a static page
+
+It is still possible to use the Nhost SDK in a more classic 'React' mode. It will then be handlded in a very similar way, except that the refresh token will be stored in the cookie instead of `localStorage`.
```
szilarddoro, line 96: I found a typo here: "handlded" should be "handled"

packages/nextjs/src/utils.ts
```diff
@@ -0,0 +1,16 @@
...
+): Promise<NhostSession | null> => {
+  const result = await fetch(`${nhostUrl}/v1/auth/token`, {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json'
+    },
+    body: JSON.stringify({ refreshToken })
+  })
+  if (result.status >= 400) return null
+  return await result.json()
```
szilarddoro, line 15: `await` is unnecessary here, because the function is asynchronous and its result would be awaited by users of the lib anyway.

packages/nextjs/src/utils.ts
```diff
@@ -0,0 +1,16 @@
...
+  refreshToken: string
+): Promise<NhostSession | null> => {
+  const result = await fetch(`${nhostUrl}/v1/auth/token`, {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json'
+    },
+    body: JSON.stringify({ refreshToken })
+  })
+  if (result.status >= 400) return null
```
szilarddoro, line 14: Can we use `result.ok` here instead? Just asking because I don't really know what status codes should result in a nullish result.

elitan:
Very interesting to see!

Here are a few ideas. Let me know what you think:

- Always provide:
  - `isLoading`, `isError`, and `isSuccess`. Similar to [React Query](https://react-query.tanstack.com/guides/queries#query-basics).
- rename `change` to `changeEmail` for `useChangeEmail`
- rename `chenge` to `changePassword` for `useChangePassword`

To add parameters to a hook like this feels strange:

```js
const [email, setEmail] = useState('')
  const { change, loading, success, hasError, error } = useChangeEmail(password)
```

It feels more natural (for me) to just initiate the hook and then use `change` (or `changeEmail`) in the onSubmit handler.


> When the authentication status is known, this `useReady` will return true. `useLoading` does the exact opposite things: it returns true when the authentication status is not known yet.

Having both feels redundant if they always return the opposite of each other.

(I didn't read all, I'll add more later.)

packages/react-apollo/src/provider.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+// TODO check this type
+type Props = Omit<NhostApolloClientOptions, 'backendUrl' | 'authService'>
+const Wrapper: React.FC<Props> = ({ children, ...options }) => {
+  const { interpreter, backendUrl } = useContext(NhostReactContext)
+
+  const [client, setClient] = useState<ApolloClient<InMemoryCache>>()
+
+  useEffect(() => {
+    setClient(createApolloClient({ interpreter, backendUrl, ...options }))
```
szilarddoro, line 15: If we just simply want to create an Apollo client when the provider component is mounted it would be enough to create this client outside of the `useEffect` call. That would improve the readability of this component. But let's discuss it, probably I'm missing something that makes this hook necessary.

If the `useEffect` can be removed, then we know for sure that there is no such case when a `client` is not available, thus the `return <div>no Apollo client</div>` can be removed. If there is a case where `client` is not available, then it can stay here to provide a fallback.

I'd also convert it to a more readable format then with an early-exit / if-guard approach:

```jsx
if (client) {
  return <ApolloProvider ...>...</ApolloProvider>
}

return <div>no Apollo client</div>
```

packages/react/src/hooks/registration.ts
```diff
@@ -0,0 +1,34 @@
+import { useMemo } from 'react'
+
+import { useSelector } from '@xstate/react'
+
+import { useAuthenticated, useNhostInterpreter, useReady } from './common'
+
+export const useSignUpEmailPassword = (stateEmail?: string, statePassword?: string) => {
```
szilarddoro, line 7: To maintain consistency here, we should name this hook as `useEmailPasswordSignUp` instead, since `useSignIn` hooks are following this pattern.

plmercereau:
Just pushed modifications according to your comments and @szilarddoro's.

> To add parameters to a hook like this feels strange:
> 
> ```js
> const [email, setEmail] = useState('')
>   const { change, loading, success, hasError, error } = useChangeEmail(password)
> ```
> 
> It feels more natural (for me) to just initiate the hook and then use `change` (or `changeEmail`) in the onSubmit handler.
> 

both works:
```js 
const [email, setEmail] = useState('')
const { change, loading, success, hasError, error } = useChangeEmail(password)
return <button onClick={change}>Change email</button>
```

```js 
const [email, setEmail] = useState('')
const { change, loading, success, hasError, error } = useChangeEmail(password)
return <button onClick={() => change(password)}>Change email</button>
```
we may somehow want to explain both syntaxes in the documentation without having readers getting lost

packages/react-apollo/src/provider.tsx
plmercereau: it was related to an attempt to make Apollo and Nextjs work together, but I set this aside (it's a real pain), so I simplified in the latest commit. Thanks!

docs/content/docs/reference/react/hooks.mdx
```diff
@@ -0,0 +1,393 @@
...
+- `isSuccess: boolean` returns `true` if the registration suceeded. Returns `false` if the new email needs to be verified first, or if an error occurred.
+- `isError: boolean` returns `true` if an error occurred.
+- `error?: {status: number, error: string, message: string}` provides details about the error.
+- `SignUpOption`:
+  - `locale?: string` locale of the user, in two digits, for instance `en`.
+  - `allowedRoles?: string[]`: allowed roles of the user. Must be a subset of the default allowed roles defined in Hasura Auth.
+  - `defaultRole?: string`: default role of the user. Must be part of the default allowed roles defined in Hasura Auth.
+  - `displayName?: string`
+  - `metadata?: Record<string, unknown>`: Custom additional user information stored in the `metadata` column.
+  - `redirectTo?: string`: redirection path in the client application that will be used in the link in the verification email. For instance, if you want to redirect to `https://myapp.com/success`, the `redirectTo` value is `'/success'`.
```
elitan, line 26: Is this accurate? I thought you would have to specify the full URL for `redirectTo`.

docs/content/docs/reference/react/hooks.mdx
```diff
@@ -0,0 +1,393 @@
+---
+title: 'Hooks'
+---
+
+## Authentication Hooks
+
+### Email + password registration
```
elitan, line 7: Use "sign-up" instead of "registration".

docs/content/docs/reference/react/hooks.mdx
```diff
@@ -0,0 +1,393 @@
...
+
+```js
+const { signUp, isLoading, isSuccess, needsVerification, isError, error } =
+  useEmailPasswordSignUp(email?: string, password?: string, options?: SignUpOptions )
+```
+
+- `signUp: (email?: string, password?: string) => void` triggers the user registration. The email/password arguments will take precedence over the possible state values used when creating the hook.
```
elitan, line 14: "triggers the user registration" => "is used to sign up a user" or something similar. Effectively changing "registration" to "sign up".

docs/content/docs/reference/react/hooks.mdx
```diff
@@ -0,0 +1,393 @@
...
+- `needsVerification: boolean` returns `true` if the registration has been accepted, but a verificaiton email has been sent and is awaiting.
+- `isSuccess: boolean` returns `true` if the registration suceeded. Returns `false` if the new email needs to be verified first, or if an error occurred.
+- `isError: boolean` returns `true` if an error occurred.
+- `error?: {status: number, error: string, message: string}` provides details about the error.
+- `SignUpOption`:
+  - `locale?: string` locale of the user, in two digits, for instance `en`.
+  - `allowedRoles?: string[]`: allowed roles of the user. Must be a subset of the default allowed roles defined in Hasura Auth.
+  - `defaultRole?: string`: default role of the user. Must be part of the default allowed roles defined in Hasura Auth.
+  - `displayName?: string`
+  - `metadata?: Record<string, unknown>`: Custom additional user information stored in the `metadata` column.
```
elitan, line 25: Maybe mention that this can be any JSON object?

docs/content/docs/reference/react/hooks.mdx
```diff
@@ -0,0 +1,393 @@
...
+      <input
+        value={password}
+        onChange={(event) => setPassword(event.target.value)}
+        placeholder="Password"
+      />
+      <button onClick={signUp}>Register</button>
+      {isSuccess && <div>Your account have beed created! You are now authenticated</div>}
+      {needsVerification && (
+        <div>
+          Please check your mailbox and follow the verification link to confirm your registration
```
elitan, line 51: Suggetion: "confirm your registration" => "verify your email".

docs/content/docs/reference/react/hooks.mdx
```diff
@@ -0,0 +1,393 @@
...
+        </div>
+      )}
+    </div>
+  )
+}
+```
+
+### Email + password authentication
```
elitan, line 59: Suggestion: 

### Email + password authentication

to

### Email and Password Sign-In

docs/content/docs/reference/react/hooks.mdx
```diff
@@ -0,0 +1,393 @@
...
+
+```js
+const { signIn, isLoading, needsVerification, isSuccess, isError, error } =
+  useEmailPasswordSignIn(email?: string, password?: string)
+```
+
+- `signIn: (email?: string, password?: string) => void` will try to authenticate. The email/password arguments will take precedence over the possible state values used when creating the hook.
```
elitan, line 66: What do you think about changing this `signIn` to `emailPasswordSignIn` to make it explicit?

docs/content/docs/reference/react/hooks.mdx
```diff
@@ -0,0 +1,393 @@
...
+        value={password}
+        onChange={(event) => setPassword(event.target.value)}
+        placeholder="Password"
+      />
+      <button onClick={signUp}>Register</button>
+      {isSuccess && <div>Authentication suceeded</div>}
+      {needsVerification && (
+        <div>
+          Your email is still pending verification. Check your mailbox and follow the instructions
+          to activate your account.
```
elitan, line 97: Suggestion: "You must verify your email to sign in. Check your mailbox and follow the instructions to verify your email".

docs/content/docs/reference/react/hooks.mdx
```diff
@@ -0,0 +1,393 @@
...
+        </div>
+      )}
+    </div>
+  )
+}
+```
+
+### Oauth providers
```
elitan, line 105: "Oauth providers" -> "OAuth Providers"

docs/content/docs/reference/react/hooks.mdx
```diff
@@ -0,0 +1,393 @@
...
+const { signIn, isLoading, isSuccess, isError, error } =
+  useEmailPasswordlessSignIn(email?: string, options?: PasswordlessOptions)
+```
+
+- `signIn: (email?: string) => void` will send a magic link to the given email The email argument will take precedence over the the possible state value used when creating the hook.
+- `isLoading: boolean` returns `true` when the action is executing, `false` when the action result is available.
+- `isSuccess: boolean` returns `true` if the magic link email user has successfully send.
+- `isError: boolean` returns `true` if an error occurred.
+- `error?: {status: number, error: string, message: string}` provides details about the error.
+- `PasswordlessOptions`:
```
elitan, line 128: Are these options specific to "passwordless" (given the name `PasswordlessOptions`). I think they are generic to other sign-up methods too?

docs/content/docs/reference/react/hooks.mdx
```diff
@@ -0,0 +1,393 @@
...
+          An email has been sent to {email}. Please check your mailbox and click on the
+          authentication link.
+        </div>
+      )}
+    </div>
+  )
+}
+```
+
+### Sign out
```
elitan, line 160: Sign out => Sign Out

docs/content/docs/reference/react/hooks.mdx
```diff
@@ -0,0 +1,393 @@
...
+        {isSuccess && <div>You have successfully signed out!</div>}
+      </div>
+    )
+  else return <div>Not authenticated</div>
+}
+```
+
+## Authentication status
+
+### `useLoading`
```
elitan, line 193: Do we want a more explicit name for this hook? Like `useAuthLoading` or something similar?

docs/content/docs/reference/react/hooks.mdx
```diff
@@ -0,0 +1,393 @@
...
+
+<!-- TODO better documentation -->
+
+During startup, the client may need some initial steps to determine the authentication status, like fetching a new JWT from an existing refresh token.
```
elitan, line 197: "The client may need some initial steps to determine the authentication status during startup, like fetching a new JWT from an existing refresh token."

docs/content/docs/reference/react/hooks.mdx
```diff
@@ -0,0 +1,393 @@
...
+```js
+const jwt = useAccessToken()
+```
+
+## User management
+
+### Change email
+
+```js
+const { change, isLoading, isSuccess, needsVerification, isError, error } =
```
elitan, line 228: `change` => `changeEmail`

docs/content/docs/reference/react/hooks.mdx
```diff
@@ -0,0 +1,393 @@
...
+```
+
+### Change password
+
+```js
+const { changePassword, isLoading, isSuccess, isError, error } = useChangePassword(password?: string)
+```
+
+- `changePassword: (password?: string) => void` requests the password change. The arguement password will take precedence over the the possible state value used when creating the hook.
+- `isLoading: boolean` returns `true` when the action is executing, `false` when the action result is available.
```
elitan, line 269: What is "action result" and where is it available?

docs/content/docs/reference/react/hooks.mdx
```diff
@@ -0,0 +1,393 @@
...
+      <input value={password} onChange={(event) => setPassword(event.target.value)} />
+      <button onClick={changePassword}>Change password</button>
+    </div>
+  )
+}
+```
+
+### Reset password
+
+When an user lost their password, we can resend them an email to authenticate, so they can change their password:
```
elitan, line 294: "If a user loses their password, we can resend them an email to authenticate so that they can change their password:"

docs/content/docs/reference/react/protecting-routes.mdx
```diff
@@ -0,0 +1,44 @@
+---
+title: 'Protecting routes'
+---
+
+## React Router
+
+When using [React Router](https://reactrouter.com/web/guides/quick-start) with `@nhost/react`, you can protect routes by creating an `AuthGate` component:
```
elitan, line 7: "You can protect routes by creating an `AuthGate` component when using `@nhost/react` with [React Router](https://reactrouter.com/web/guides/quick-start)."

docs/content/docs/reference/nextjs/configuration.mdx
```diff
@@ -0,0 +1,101 @@
+---
+title: 'Configuration'
+---
+
+## Configuration
+
+To configure Nhost with Next.js follows the same logic as with React, except we are intialising with `NhostSSR` instead of `Nhost`.
+Under the hood, `NhostSSR` uses cookies to store the refresh token, and disables auto-refresh and auto-login when running on the server side.
```
elitan, line 8: "Configuring Nhost with Next.js follows the same logic as React, except we are initializing with `NhostSSR` instead of `Nhost`.
Under the hood, `NhostSSR` uses cookies to store the refresh token, and disables auto-refresh and auto-login when running on the server-side."

docs/content/docs/reference/nextjs/index.mdx
```diff
@@ -0,0 +1,116 @@
+---
+title: 'Introduction'
+---
+
+It is possible to use [`@nhost/react`](/reference/react) in any Next.js page that would be configured to render on the client side.
+
+When rendering a page from the server side, Next.js needs to get some information from the client to determine their authentication status. Such a communication is only available from cookies, and the the Nhost client has been designed to enable such a mechanism.
```
elitan, line 7: "It is possible to use [`@nhost/react`](/reference/react) in any Next.js page that would be configured to render on the client-side.

When rendering a page from the server-side, Next.js needs to get some information from the client to determine their authentication status. Such communication is only available from cookies, and the Nhost client is designed to enable such a mechanism."

docs/content/docs/reference/nextjs/index.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+The cookie is `same-site` and is used only to store the refresh cookie and the expiration date of the JWT so the browser can proactively ask for a new JWT before it expires.
+
+## Signing in
+
+There is no need to activate SSR in the sign-in page as there is not yet any session information to capture. As a result, it is recommended to code it as a classic static page.
+
+The intitial authentication would then work like this diagram:
```
elitan, line 15: "There is no need to activate SSR on the sign-in page as there is no session information to capture. As a result, it is recommended to code it as a classic static page.

The initial authentication would then work like this diagram:"

elitan (changes requested):
Wow. This was a huge PR. Great job! 

I took 30 min to review but now I need to go for lunch :D But I'll submit this one for now.

## Grammarly

I added a bunch of edits I did together with Grammarly. Feel free to adapt them if you think they make sense.


## Consistency

I also think we should be consistent in using:

- sign up / sign in (verb)
- sign-up / sign-in (noun)

instead of register, registration, or login.

## Headers

I think we should always use capitalization for headers (h1/2/3/4/etc)

docs/content/docs/reference/react/hooks.mdx
plmercereau: yes, they are the same as for email+password, I renamed to `Options` so it simplifies the documentation

docs/content/docs/reference/react/hooks.mdx
plmercereau: I find it too verbose and explicit enough with the hook name

examples/nextjs/pages/second.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+const SecondPage: React.FC<{ initial: NhostSession }> = () => {
+  const isAuthenticated = useAuthenticated()
+  const user = useUserData()
+  const jwt = useAccessToken()
+  if (!isAuthenticated) return <div>User it not authenticated</div>
+  return (
+    <div>
+      <h1>
+        SECOND User {user?.displayName} is authenticated {isAuthenticated ? 'OUI' : 'NON'}
```
elitan, line 26: OUI / NON => English? :)

examples/nextjs/pages/second.tsx
```diff
@@ -0,0 +1,33 @@
...
+    props: {
+      nhostSession
+    }
+  }
+}
+
+const SecondPage: React.FC<{ initial: NhostSession }> = () => {
+  const isAuthenticated = useAuthenticated()
+  const user = useUserData()
+  const jwt = useAccessToken()
```
elitan, line 21: I think we should always use "access token" instead of "jwt" for consistency. This has to be changed in more places than here.

.changeset/react.md
```diff
@@ -0,0 +1,37 @@
...
+- `useAuthLoading`
+- `useNhost`
+- `useNhostBackendUrl`
+- `useNhostInterpreter`
+- `useResetPassord`
+- `useSignOut`
+- `useUserData`
+- `useUserId`
+- `useUserLocale`
+- `useUserRoles`
```
elitan, line 34: Can these hooks be combined with `useUser`?

.changeset/react.md
```diff
@@ -0,0 +1,37 @@
...
+- `useEmail`
+- `useEmailPasswordlessSignIn`
+- `useEmailPasswordSignIn`
+- `useEmailPasswordSignUp`
+- `useIsAnonymous`
+- `useAuthLoading`
+- `useNhost`
+- `useNhostBackendUrl`
+- `useNhostInterpreter`
+- `useResetPassord`
```
elitan, line 29: typo "passord" => "password"

.changeset/react.md
```diff
@@ -0,0 +1,37 @@
...
+
+This release brings a full rewrite of the React client, to make it tree shakable and fully reactive.
+See the [documentation](https://docs.nhost.io/reference/react) for further information.
+
+### Hooks
+
+- `useAccessToken`
+- `useAnonymousSignIn`
+- `useAuthenticated`
+- `useAvatarUrl`
```
elitan, line 15: Feels a bit verbose to have a hook just for the avatarUrl. How about a `useUser(Data)` hook with this, and more, user data?

.changeset/react.md
```diff
@@ -0,0 +1,37 @@
...
+- `useChangeEmail`
+- `useChangePassword`
+- `useDefaultRole`
+- `useDisplayName`
+- `useEmail`
```
elitan, line 20: Same as avatarUrl above

docs/components/MDX/components.tsx
```diff
@@ -144,4 +144,7 @@ const components = {
   }
 }
 
+let currentId = 0
+const uuid = () => `mermaid-${(currentId++).toString()}`
+
```
elitan, line 149: This does not seem to be used anywhere.

packages/react/src/hooks/common.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export const useAuthLoading = () => {
+  const service = useNhostInterpreter()
+  return useSelector(service, (state) => !state.hasTag('ready'))
+}
+
+export const useAuthenticated = () => {
+  const service = useNhostInterpreter()
+  return useSelector(service, (state) => state.matches({ authentication: 'signedIn' }))
+}
```
elitan, line 33: Will this hook automatically set a new value if I'm signed out and then I successfully do `nhost.auth.signIn({email, password});`?

docs/content/docs/reference/react/hooks.mdx
elitan: Every other hook match the pattern: `const { xxx } = useXxx();`. So I think we should follow that pattern even for `useEmailPasswordSignIn()`.

Like: `const { changePassword } = useChangePassword();`

Thoughts @nunopato @szilarddoro?

docs/content/docs/reference/react/hooks.mdx
elitan: I don't think so.

In Hasura Auth we grab the `redirectTo` from the query parameter [here](https://github.com/nhost/hasura-auth/blob/main/src/routes/verify/verify.ts#L12) and then send a redirect [here](https://github.com/nhost/hasura-auth/blob/main/src/routes/verify/verify.ts#L89-L91).

docs/content/docs/reference/react/index.mdx
```diff
@@ -0,0 +1,58 @@
...
+| ------------------ | ----------------------------------- | ---------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `backendUrl`       | string                              |                  | The Nhost app url, for instance `https://my-app.nhost.run`. When using the CLI, its value is `http://localhost:1337`                                                                                                                           |
+| `autoLogin`        | boolean                             | `true`           | If set to `true`, the client will detect credentials in the current URL that could have been sent during an email verification or an Oauth authentication. It will also automatically authenticate all the active tabs in the current browser. |
```
elitan, line 55: `autoLogin` should be `autoSignIn` if we want to be consistent. Which I think we should.

docs/content/docs/reference/react/hooks.mdx
plmercereau: the xstate-based client prepends any route starting with a `/` with the client URL, which is by default `window.location.origin`. the full path is always send to hasura-auth

examples/nextjs/pages/second.tsx
plmercereau: ok

packages/react/src/hooks/common.ts
plmercereau: yes

docs/components/MDX/components.tsx
plmercereau: the Mermaid component has been quashed in one of the merge from main. Added it again

docs/content/docs/reference/react/index.mdx
plmercereau: I used the same name as in nhost-js and hasura-auth-js. Changed to autoSignIn

.changeset/react.md
plmercereau: we could. But in terms of typings, userData is `User | null`, meaning devs won't be able to destructure.
so something like `const { email } = useUserData()` will raise a typescript error.
whereas `const email = useEmail()` won't.

elitan (approved):

packages/react-apollo/src/provider.tsx
plmercereau: When using Next, if `createClient` is not wrapped into `useEffect`, then it seeks to render the page on the server side. This is to avoid as we don't want SSR if no `getServerSideProps` is present.
Plus, for the moment, Apollo is not correctly set with the sdk to run on the server side, then to hydrate the apollo cache.
Wrapping again in an `useEffect` until we found a better solution

PR #239, opened 2022-03-02
Fix useNhostAuth typings

### LINKED ISSUE #187: The TypeScript type for `user` returned from the `useNhostAuth` object is always `null`
The correct type should be `User | null`.

plmercereau:
Close #187 

Another minor point: `pnpm run clean:all` now also removes the `.nhost` folders

### DIFF
--- a/packages/react-auth/src/index.tsx
+++ b/packages/react-auth/src/index.tsx
@@ -1,8 +1,14 @@
 import React, { createContext, ReactNode, useContext, useEffect, useState } from 'react'
 
+import type { User } from '@nhost/hasura-auth-js'
 import { NhostClient } from '@nhost/nhost-js'
 
-export const AuthContext = createContext({
+type NhostContext = {
+  isLoading: boolean
+  isAuthenticated: boolean
+  user: User | null
+}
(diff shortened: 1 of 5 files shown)


elitan (approved):

szilarddoro (approved):

packages/react-auth/src/index.tsx
```diff
@@ -1,8 +1,14 @@
 import React, { createContext, ReactNode, useContext, useEffect, useState } from 'react'
 
+import type { User } from '@nhost/hasura-auth-js'
 import { NhostClient } from '@nhost/nhost-js'
 
-export const AuthContext = createContext({
+type NhostContext = {
```
szilarddoro, line 6: I think we could export this type to make it available to users of the library.

One scenario where this could come particularly useful is when they try to create their own context which would extend our `AuthContext` in some way.

The same is also true for our `NhostAuthProvider` which could have its props exported for the same reason.

szilarddoro: I'm leaving this here for future reference.

PR #242, opened 2022-03-02
docs: add custom-email page

plmercereau:


szilarddoro:
Screenshot 2022-03-03 at 9 24 17

The corners of the dialog are looking a bit weird. If I export it manually from Figma, that looks the same, so perhaps @jerryjappinen could check the asset and help with this.

PR #244, opened 2022-03-03
refactor(hasura-storage-js): rename `storage.getUrl` to `storage.getPublicUrl`

### LINKED ISSUE #178: Rename `getUrl` to `getPublicUrl`
There are two ways of getting a file URL:

1. Get the direct (or public) URL. For users to access the file using this URL the `public` role in Hasura must have access to the file in the GraphQL API.
2. Pre signed URL. For users to access this file the user has to have permission to the file in the GraphQL API.

We could keep the `getUrl`, deprecate it, and use it as an alias for `getPublicUrl`.

With the new naming we would have two ways of getting a file URL:

- `getPublicUrl`
- `getPresginedUrl`

plmercereau:
It aims to make a clear distinction between `storage.getPublicUrl` and `storage.getPresginedUrl`
`storage.getUrl` is now deprecated.

fix #178

### DIFF
--- a/.changeset/long-cars-melt.md
+++ b/.changeset/long-cars-melt.md
@@ -1,0 +1,7 @@
+---
+'@nhost/hasura-storage-js': patch
+---
+
+Rename `storage.getUrl` to `storage.getPublicUrl`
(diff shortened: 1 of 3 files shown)


elitan (approved):
Nice!

PR #266, opened 2022-03-10
feat: correct available providers (Discord & Twitch added)

elitan:


plmercereau (commented):
why apple, twitter and windowslive are removed?

elitan:
> why apple, twitter and windowslive are removed?

My idea was since those providers are not available on Nhost. But they are available in Hasura Auth so I'll put them back.

plmercereau (approved):

PR #273, opened 2022-03-11
Unify @nhost/react and @nhost/nhost-js

plmercereau:


examples/nextjs/pages/index.tsx
```diff
@@ -25,9 +25,9 @@ const Home: NextPage = () => {
   const [newPassword, setNewPassword] = useState('')
   const accessToken = useAccessToken()
   const { signOut } = useSignOut()
-  const { signUp, ...signUpResult } = useEmailPasswordSignUp(email, password)
-  const { signIn } = useEmailPasswordSignIn(email, password)
-  const { signIn: passwordlessSignIn } = useEmailPasswordlessSignIn(email)
+  const { signUpEmailPassword, ...signUpResult } = useSignUpEmailPassword(email, password)
+  const { emailPasswordSignIn } = useEmailPasswordSignIn(email, password)
+  const { emailPasswordlessSignIn } = useEmailPasswordlessSignIn(email)
```
elitan, line 30: Either we should name the variable name and hooks `sign[Up|In]...` or `...Sign[Up|In]` and I think I would prefer to start with `sign[Up|In]...` just like the current `useSignUpEmailPassword` already uses.

examples/react-apollo/src/App.tsx
```diff
@@ -57,8 +57,22 @@ function App() {
...
+                <SignInPage />
+              </PublicGate>
+            }
+          />
+          <Route
+            path="/sign-up/*"
+            element={
+              <PublicGate>
+                <SignUpPage />
+              </PublicGate>
```
elitan, line 73: Do we really need a `PublicGate` wrapper?

examples/react-apollo/src/components/auth-gates.tsx
```diff
@@ -25,8 +25,7 @@ export const PublicGate: React.FC = ({ children }) => {
   }
 
   if (isAuthenticated) {
-    // ? stay on the same route - is it the best way to do so?
-    return <Navigate to={location} state={{ from: location }} replace />
+    return <Navigate to={'/'} state={{ from: location }} replace />
```
elitan, line 28: I'm not sure I understand the value of this `PublicGate` component.

examples/react-apollo/src/profile/mfa.tsx
```diff
@@ -0,0 +1,28 @@
+import { useConfigMfa } from '@nhost/react'
+import { useState } from 'react'
+import { Button, Input, Panel } from 'rsuite'
+
+export const Mfa: React.FC = () => {
+  const [code, setCode] = useState('')
+  const { generate, activate, isActivated, isGenerated, qrCode } = useConfigMfa(code)
```
elitan, line 7: I feel these variable names (`generate`, `activate`, `qrCode`) could be a bit more descriptive.

Generate what?
Activate what?
qrCode - is this a base64 string? URL? or something else?

elitan (changes requested):
Wow. So much code. This looks really exciting.

I mostly had naming feedback.

examples/nextjs/pages/index.tsx
plmercereau: Corrected in https://github.com/nhost/nhost/pull/273/commits/22424a5f35246c78d88fa636b560647997216ca7

examples/react-apollo/src/App.tsx
plmercereau: Not much. It redirects authentication-related routes to index so we never show authentication routes (e.g. `/sign-in`) to an already authenticated user.
It's only in the example, but I can remove it if you want

examples/react-apollo/src/components/auth-gates.tsx
plmercereau: see the above comment

examples/react-apollo/src/profile/mfa.tsx
plmercereau: - `generate` the QR code
- `activate` MFA for the user from the code that we get after scanning the QR code
- `qrCode` is the base64 image that is generated by hasura-auth and that can be used as a `src` property in `` like ``

What about:
- `generate` -> `generateQrCode`
- `activate` -> `activateMfa`

elitan: generate -> generateQrCode
activate -> activateMfa

looks good 👍 

About `qrCode`. What about something like `qrCodeDataUrl` (https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs)?

Or does the `qrCode` includes the `data:...`?

examples/react-apollo/src/App.tsx
elitan: At first, I thought it looked strange but I tried GitHub and Vercel and they redirect users from the login page to `/` so I think we should keep it 👍

examples/react-apollo/src/profile/mfa.tsx
plmercereau: It does - we're using the [`toDataURL`](https://github.com/soldair/node-qrcode#nodejs) function on the hasura-auth side.

Changed names in https://github.com/nhost/nhost/pull/273/commits/9b942a59bc33d30fe5fec05fd97f26d710295e7a
I also renamed to `qrCodeDataUrl`

Generally speaking, when I'm picking names, I ask myself the question "could it be something else" For instance can `qrCode` be something else than a data uri? If yes, then we need to be explicit. Can it generate something else than a qrCode, can it activate something else than MFA, etc.

elitan: Yeah, maybe I'm picky about the names. And there is no "right" or "wrong" really. Naming and cache. Very hard.

I could argue for both your initial naming and mine.

.changeset/long-rockets-prove.md
```diff
@@ -0,0 +1,16 @@
+---
+'@nhost/react-auth': major
+---
+
+Use `@nhost/react`
```
elitan, line 5: Incomplete sentence.

elitan (changes requested):

elitan (approved):

packages/react/src/hooks/authentication.ts
```diff
@@ -1,83 +1,136 @@
...
-export const useEmailPasswordSignIn = (stateEmail?: string, statePassword?: string) => {
-  const service = useNhostInterpreter()
-  const signIn = (valueEmail?: string | unknown, valuePassword?: string | unknown) =>
+export const useSignInEmailPassword = (
+  stateEmail?: string,
+  statePassword?: string,
+  stateOtp?: string
+) => {
+  const service = useAuthInterpreter()
+  const signInEmailPassword = (valueEmail?: string | unknown, valuePassword?: string | unknown) =>
```
elitan, line 22: How about we use `email` and `password` instead of `valueEmail` and `valuePassword` to make it look better in an editor?

image

elitan: In the `@nhost/nhost-js` SDK we always used objects as parameters. Maybe it's worth doing that here too?

```
signInEmailPassword({email, password});
```

PR #274, opened 2022-03-11
fix: current options when sign in with a provider

elitan:
We currently only support setting the `redirectTo` option for providers.

This PR removed the options that do not work and adds the `redirectTo` option correctly to the provider sign-in URL.

See code from Haura Auth here: https://github.com/nhost/hasura-auth/blob/main/src/routes/signin/providers/utils.ts#L255-L268

### DIFF
--- a/packages/hasura-auth-js/src/utils/types.ts
+++ b/packages/hasura-auth-js/src/utils/types.ts
@@ -92,12 +92,12 @@ export type Provider =
 export interface SignInWithProviderOptions {
   provider: Provider
   options?: {
-    locale?: string
-    allowedRoles?: string[]
-    defaultRole?: string
-    displayName?: string
+    // locale?: string
+    // allowedRoles?: string[]
+    // defaultRole?: string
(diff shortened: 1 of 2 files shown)


plmercereau (approved):

elitan:
I think we can wait for https://github.com/nhost/hasura-auth/pull/108 before we update and merge this PR.

packages/hasura-auth-js/src/utils/types.ts
```diff
@@ -92,12 +92,12 @@ export type Provider =
...
-    allowedRoles?: string[]
-    defaultRole?: string
-    displayName?: string
+    // locale?: string
+    // allowedRoles?: string[]
+    // defaultRole?: string
+    // displayName?: string
     redirectTo?: string
-    metadata?: Record<string, unknown>
+    // metadata?: Record<string, unknown>
```
nunopato, line 100: why are we commenting these?

nunopato (approved):

Issue #302, opened 2022-03-28
Protected Routes not working as intended. isLoading and isAuthenticated are both false even when user is authenticated


alveshelio:
Hi,

Looking at the example to protect private routes here: https://docs.nhost.io/reference/nextjs/protecting-routes with the new package `@nhost/nextjs`

```ts
import { useRouter } from 'next/router'
import { useAuthLoading, useAuthenticated } from '@nhost/nextjs'

export function authProtected(Comp) {
  return function AuthProtected(props) {
    const router = useRouter()
    const isLoading = useAuthLoading()
    const isAuthenticated = useAuthenticated()

    if (isLoading) {
      return <div>Loading...</div>
    }

    if (!isAuthenticated) {
      router.push('/login')
      return null
    }

    return <Comp {...props} />
  }
}
```
And this isn't working, the thing is that when `isLoading` is set to false, `isAuthenticated` is still set to false and it creates a redirect to `/login`. In my `Login.tsx` component I check if the user is logged, if so I redirect him to the home page. However, when the user arrives at the login page coming from the redirect of `authProjected` `isAuthenticated` is `true` and it then redirects the user to the root `/`

Somehow, when you set `isLoading` to `false`, the verifications hasn't finished yet and so we get the case where `isLoading` is `false` and `isAuthenticated` is `false` which will cause the user to get redirected.

There's probably a state that is not properly set in your state machine. I don't think we should have this race condition where we finish checking if `isLoading` and `isAuthenticated` is `false` and the next update `isAuthenticated` is `true`.

PR #303, opened 2022-03-29
Deprecate `useAuthLoading` and introduce `useAuthenticationStatus`

plmercereau:
When using both `useAuthLoading` and `useAuthenticated` together, the hooks rerender independently from each other.
As a result, when a user loads the page while he previously authenticated, the hooks values were chronologically:

| isLoading | isAuthenticated |
| --------- | --------------- |
| `true`    | `false`         |
| `false`   | `false`         |
| `false`   | `true`          |

The intermediate (`false`, `false`) is incorrect and is causing issues when using an authentication gate.

It is therefore recommended to stop using `useAuthLoading`, and to use `useAuthenticationStatus` instead, in order to keep the loading state and the authentication in sync within the same hook.

Usage:

```js
const { isLoading, isAuthenticated } = useAuthenticationStatus()
```

Fixes [this issue](https://github.com/nhost/nhost/issues/302)

### DIFF
--- a/.changeset/red-tables-punch.md
+++ b/.changeset/red-tables-punch.md
@@ -12,7 +12,7 @@ As a result, when a user loads the page while he previously authenticated, the h
 | `false`   | `false`         |
 | `false`   | `true`          |
 
-The interpediate (`false`, `false`) is incorrect and is causing issues when using an authentication gate.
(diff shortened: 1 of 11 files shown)


szilarddoro (approved):
looks good to me! 🎉

PR #310, opened 2022-03-31
fix: check if `window.location` exists

plmercereau:


packages/core/src/machines/index.ts
```diff
@@ -635,10 +635,10 @@ export const createAuthMachine = ({
           }),
 
         autoSignIn: async () => {
```
szilarddoro, line 637: I'd try to make an early exit here (the error message is just a sample):
```js
if (typeof window === undefined || !window.location) {
  throw new Error('window is undefined or location does not exist')
}
```

packages/core/src/machines/index.ts
```diff
@@ -635,10 +635,10 @@ export const createAuthMachine = ({
...
-          if (typeof window !== 'undefined') {
-            const location = window.location
-            if (location.hash) {
-              const params = new URLSearchParams(location.hash.slice(1))
+          if (typeof window !== 'undefined' && window.location) {
+            const hash = window.location.hash
```
szilarddoro, line 639: this could be converted to:
```js
const { hash } = window.location
```

packages/core/src/machines/index.ts
```diff
@@ -635,10 +635,10 @@ export const createAuthMachine = ({
...
+            if (hash) {
```
szilarddoro, line 640: similar to the early exit where we throw the error, we could make one here too by simply returning when `hash` is `undefined`

packages/core/src/machines/index.ts
```diff
@@ -635,10 +635,10 @@ export const createAuthMachine = ({
...
+              const params = new URLSearchParams(hash.slice(1))
               const refreshToken = params.get('refreshToken')
               if (refreshToken) {
```
szilarddoro, line 643: We can make an early exit here similar to how we would return if `hash` is `undefined`.

The only reason I'd like to convert this is because we have too many nested if statements here.

packages/core/src/machines/index.ts
plmercereau: it doesn't change the existing logic I think:
!(a && b) <=> (!a || !b)

packages/core/src/machines/index.ts
plmercereau: ok

packages/core/src/machines/index.ts
```diff
@@ -635,27 +635,25 @@ export const createAuthMachine = ({
...
-                // TODO remove the hash. For the moment, it is kept to avoid regression from the current SDK.
-                // * Then, only `refreshToken` will be in the hash, while `type` will be sent by hasura-auth as a query parameter
-                // window.history.pushState({}, '', location.pathname)
-                const channel = new BroadcastChannel('nhost')
-                // TODO broadcat session instead of token
-                channel.postMessage(refreshToken)
-                return { session }
-              }
-            }
+          if (typeof window === 'undefined' || !window.location)
```
szilarddoro, line 638: I'd say we should wrap this in curly braces so we could visually separate our if statements a bit.

We are using the [curly](https://eslint.org/docs/rules/curly) ESLint rule against this in the console.

packages/core/src/machines/index.ts
```diff
@@ -635,27 +635,25 @@ export const createAuthMachine = ({
...
+            throw Error('window is undefined or location does not exist')
+          const { hash } = window.location
+          if (!hash) return
```
szilarddoro, line 641: Same as above.

szilarddoro (approved):
Approved ✅  Just wanted to mention a small thing we should consider.

packages/core/src/machines/index.ts
plmercereau: I'm fine with it, although we will have to change it in lots of place as I have this (bad) habit :)

PR #317, opened 2022-03-31
Add custom endpoint options to the NhostClient constructor, and document NhostClient options

### LINKED ISSUE #199: Custom auth and graphql endpoints
Hi, the new sdk (nhost-js) accepts only backendUrl string 

```
export const nhost = new NhostClient({ backendUrl })
```
as the base for auth, storage and graphql endpoints:
```
    this.auth = new HasuraAuthClient({
      url: `${backendUrl}/v1/auth`,
      refreshIntervalTime,
      clientStorage,
      clientStorageType,
      autoRefreshToken,
      autoLogin
    })

    this.storage = new HasuraStorageClient({
      url: `${backendUrl}/v1/storage`
    })

    this.functions = new NhostFunctionsClient({
      url: `${backendUrl}/v1/functions`
    })

    this.graphql = new NhostGraphqlClient({
      url: `${backendUrl}/v1/graphql`
    })
```

https://github.com/nhost/nhost-js/blob/5ad91d231a8c9c6a67d8060bb7edae941f5c64ff/packages/nhost/src/core/nhost-client.ts

perhaps we could allow for custom endpoints, as is the case with:

```
import { NhostApolloProvider } from '@nhost/react-apollo'
...
<NhostApolloProvider nhost={nhost} graphqlUrl={graphqlUrl}>
...
```

plmercereau:
Closes #199


packages/nhost-js/src/core/nhost-client.ts
```diff
@@ -18,7 +39,7 @@ export class NhostClient {
    * Nhost Client
    *
    * @example
-   * const nhost = new NhostClient({ url });
+   * const nhost = new NhostClient({ backendUrl: 'https://my-app.nhost.run' });
```
szilarddoro, line 42: Could you please wrap this example in a code block? This way our DocGen tool will be able to visualize the example for this constructor correctly.

packages/nhost-js/src/core/nhost-client.ts
```diff
@@ -4,10 +4,31 @@ import { HasuraStorageClient } from '@nhost/hasura-storage-js'
...
+  /** Functions endpoint. If given, it will take precedence over `backendUrl` */
+  functionsUrl?: string
 }
 
+type RequireAtLeastOne<T, Keys extends keyof T = keyof T> = Pick<T, Exclude<keyof T, Keys>> &
+  {
+    [K in Keys]-?: Required<Pick<T, K>> & Partial<Pick<T, Exclude<Keys, K>>>
+  }[Keys]
+
+export type NhostClientConstructorParams = RequireAtLeastOne<Params, 'backendUrl' | 'graphqlUrl'> &
```
szilarddoro, line 27: Wouldn't it make more sense to make sure that `backendUrl` is required, so that we could have it as a fallback value for any of the missing values? Based on `type Params` it's seemingly required, but `NhostClientConstructorParams` makes it optional.

szilarddoro: This way we could make our constructor less error prone by simply just using the fallback value if any of the custom URLs are missing. What do you think about it?

plmercereau: the logic is that either `backendUrl` is present, or all of the `graphqlUrl` etc are present.
The typing is not ideal but it works.

In the case of custom endpoints, there is no server to fall back to.
If we enforce `backendUrl` while wants to define their own custom endpoints, they would still give a value that would not be used eventually to pass the typings error, leading to potential confusion.

packages/nhost-js/src/core/nhost-client.ts
plmercereau: :) 
done

nunopato (commented):
Could you provide some context as to why this is useful? T

szilarddoro:
> Could you provide some context as to why this is useful?

https://github.com/nhost/nhost/issues/199 - This issue mentioned that it would be great to have custom URLs for certain modules. I'd say that we should get more input on this topic to see the big picture.

nunopato:
Yes, I agree. I will leave a comment on the issue to know about the use cases the user is thinking of. A Nhost app is always one endpoint so I am not sure this is useful.

plmercereau:
closing as it needs a full refactoring now we introduced the new subdomain syntax

PR #320, opened 2022-03-31
Proposal of a new hook signature

plmercereau:
Here is an example what I gather from the community call.
A few comments:
- this would fix error / isError being out of sync
- this would introduce a breaking change.

To do:
- [ ] validate the design
- [ ] extend to other hooks
- [ ] update documentation
- [ ] update examples


plmercereau:
Not going that way - see https://github.com/nhost/nhost/issues/322

PR #335, opened 2022-04-04
React hook signature & add promises to hook action results

plmercereau:
## Post-merge to-do list
- [ ] publish
- [ ] once published, bump versions in example, and remove the deprecated syntax in the react-apollo example

## Recap
1. deprecate the following syntax
```js
const { signInEmailPasswordless } = useSignInEmailPasswordless('nuno@fcporto.pt')
signInEmailPasswordless()
```
Instead, use
```js
const { signInEmailPasswordless } = useSignInEmailPasswordless()
signInEmailPasswordless('nuno@fcporto.pt')
```
-> no breaking change has been introduced. For instance, `useSignUpEmailPassword('szilard@brussels.be','1234', options)` will appear as deprecated but will work, while `useSignUpEmailPassword(options)` will work too.

2. Return a promise in hook actions:
```js
const { signInEmailPasswordless } = useSignInEmailPasswordless()
const { isSuccess, isError, error } = await signInEmailPasswordless('johan@ikea.se')
```
3. Sync up the states (the re-rendering problem we talked about between isSuccess, isError, error etc: cannot reproduce
4. Documentation and examples of the above


examples/react-apollo/src/components/email-passwordless-form.tsx
```diff
@@ -6,9 +6,13 @@ import React, { useState, useEffect } from 'react'
...
-    redirectTo: '/profile'
-  })
+  const { signInEmailPasswordless, isError, isSuccess, error } = useSignInEmailPasswordless(
+    // TODO correct this once the new packages are released
+    undefined,
+    {
+      redirectTo: '/profile'
+    }
+  )
   const [showError, setShowError] = useState(true)
```
elitan, line 16: is this `showError` needed?

packages/react/src/hooks/user.ts
```diff
@@ -7,64 +7,171 @@ import {
...
+        type: 'REQUEST',
+        email: typeof valueEmail === 'string' ? valueEmail : stateEmail,
+        options: valueOptions
+      })
+      service.onTransition((state) => {
+        if (state.matches({ idle: 'error' })) {
+          const error = state.context.error
+          setIsError(true)
+          setError(error)
+          setIsLoading(false)
```
elitan, line 64: Is this fine now? It won't cause 3 re-renders?

examples/react-apollo/src/components/email-passwordless-form.tsx
plmercereau: yes, this is required to hide the error message when the user changes the email field

elitan: Ok, I would maybe not bother with that but I see your point.

examples/react-apollo/src/sign-up/email-password.tsx
```diff
@@ -15,8 +15,9 @@ export const EmailPassword: React.FC = () => {
   const navigate = useNavigate()
   const [confirmPassword, setConfirmPassword] = useState('')
   const { signUpEmailPassword, error, needsEmailVerification, isSuccess } = useSignUpEmailPassword(
-    email,
-    password,
+    // TODO change once the new packages have been release
+    undefined,
+    undefined,
     options
```
elitan, line 21: can we remove this now?

elitan (approved):

examples/react-apollo/src/sign-up/email-password.tsx
elitan: ignore

PR #345, opened 2022-04-06
fix: don't remove comments so they are present in the built .d.ts

plmercereau:
See https://github.com/microsoft/TypeScript/issues/14619


szilarddoro (approved):
Mega Approve 👍🏼

PR #358, opened 2022-04-07
Fix a typo in the React docs

Savinvadim1312:
Changed from `clientStorageGetter` to `clientStorageSetter` on the line referencing the setter function


plmercereau (approved):
Thanks for this!

PR #381, opened 2022-04-09
fix: apollo context headers in http requests

jovermier:
Headers such has `x-hasura-role` were not making it into the apollo requests.


plmercereau:
Hello @jovermier,
Thanks for your PR!
As far as I understand your problem, it has been likely solved by #387 and #388.
Plus the Apollo link were set already in [this line](https://github.com/nhost/nhost/pull/381/files#diff-8589888bc196b22ac29a0fb50310db8a41fb0895bd0ce96af403e1db404aa7c4L125).
I'm therefore closing this PR for now. If the latest release does not solve your problem, please open an issue with some insights on how to reproduce it.
Thanks!

Issue #378, opened 2022-04-09
Documentation should explain configuration necessary to allow users to upload files.


aaronblondeau:
I had to spend quite a long time figuring out how to allow users to upload files.

First I had to hack on nhost.storage to make it send an auth token.  It was not doing this by default when I was logged in via nhost.auth (nhost js sdk version 1.0.9):

```
nhost.storage.setAccessToken(nhost.auth.getAccessToken())
```
I think most users will expect it to automatically send the token when users are logged in.

Then I had to use the Hasura console to add permissions for the "user" role to create, select, update, and delete permissions on the files table.  This also was not documented anywhere obvious.

I suggest adding documentation about these steps where most users will wind up looking for help : https://docs.nhost.io/reference/sdk/storage#nhost-storage-upload

yureckey:
I think token problem is general - sdk is kinda not working as it should https://github.com/nhost/nhost/issues/373
But your other concerns are actually reflected in the docs:
https://docs.nhost.io/platform/storage

elitan:
This is a bug. If the user is signed in the auth header should be included when doing `nhost.storage.upload()`.

Similar issue here: https://github.com/nhost/nhost/issues/373 (as mentioned above)

PR #382, opened 2022-04-09
Keep authentication status and access token in sync

### LINKED ISSUE #373: cannot quite workout how to use graphql.request with signed-in user
I'm using nhost v2 with Vue3, and it's not quite clear how to use it properly.
After user logged-in I'm keeping nhost session and client in pinia store.
When (after login) there is nhost.graphql.request I see that it lacks authorization header, which obviously gives error (for tables where I've set permissions).
[{"extensions": {"path": "$","code": "validation-failed"},"message": "no mutations exist"}]
looking through methods I see there is setAccessToken, so I've set nhost.graphql.setAccessToken(nhost.auth.getSession().accessToken);
And it worked after that. But there is no token refreshing, so it is stopping working after 15 minutes (default expiration time), 
[{"extensions": {"path": "$","code": "invalid-jwt"},"message": "Could not verify JWT: JWTExpired"}]
it's not clear when and how to refresh token and why SDK doesn't handle it?
What am I missing? I'm reading official docs and I've tried to update to latest SDK
(npm update @nhost/nhost-js "up to date")
could it be error on my side? (could I erase something when saving client/session in Pinia store?)

plmercereau:
The authentication events where not set correctly, leading the main Nhost client not to update internal states of storage/graphql/functions sub-clients when using non-react clients.
The use of private fields (`#`) is also avoided as they conflict with the use of proxies in Vue, leading to errors in the upcoming Vue library.
Fixes #373 and #378

### DIFF
--- a/.changeset/old-poets-design.md
+++ b/.changeset/old-poets-design.md
@@ -1,0 +1,10 @@
+---
+'@nhost/core': patch
+'@nhost/hasura-auth-js': patch
+'@nhost/nhost-js': patch
+---
+
+Keep authentication status and access token in sync
+The authentication events where not set correctly, leading the main Nhost client not to update internal states of storage/graphql/functions sub-clients when using non-react clients.
+The use of private fields (`#`) is also avoided as they conflict with the use of proxies in Vue, leading to errors in the upcoming Vue library.
+Fixes #373 and #378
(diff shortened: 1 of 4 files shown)


szilarddoro (approved):
Access token is properly attached to every underlying client 👍🏼

PR #388, opened 2022-04-11
feat: graphql-ws and useAuthSubscription

### LINKED ISSUE #183: Use new `graphql-ws` package
Use `graphql-ws` instead of `subscriptions-transport-ws` for the websocket client.

plmercereau:
- Use graphql-ws instead of subscriptions-transport-ws
- Introduce useAuthSubscription

closes #183


packages/apollo/package.json
```diff
@@ -55,11 +55,11 @@
   "dependencies": {
     "@nhost/core": "workspace:^",
     "graphql": "16",
-    "subscriptions-transport-ws": "^0.11.0"
+    "graphql-ws": "^5.7.0"
```
szilarddoro, line 58: ❤️

packages/apollo/src/ws.ts
```diff
@@ -0,0 +1,52 @@
...
+  const started = () => _started
+
+  const client = createClient({
+    ...options,
+    on: {
+      ...options.on,
+      connected: () => {
+        _started = true
+      },
+      opened: (socket: any) => {
```
szilarddoro, line 24: I would love to see the typing here. Even though we can't directly assign `WebSocket` to the function parameter because of how the `graphql-ws` typing is created, we can create a local variable and assign the type to it.

I was thinking about this approach:
```js
opened: (originalSocket) => { // it has a type of "unknown"
  const socket = originalSocket as WebSocket
  
  // rest of the code
}
```

From `graphql-ws`'s `client.d.ts` file:
> The first argument is actually the `WebSocket`, but to avoid
> bundling DOM typings because the client can run in Node env too,
> you should assert the websocket type during implementation.

szilarddoro (commented):
Good stuff 👍🏼

szilarddoro (approved):

PR #391, opened 2022-04-12
Update example with the latest version of React Router, v6

gdangelo:


docs/content/docs/reference/react/protecting-routes.mdx
```diff
@@ -4,49 +4,67 @@ title: 'Protecting routes'
...
   }
 
   if (!isAuthenticated) {
-    return <Redirect to="/login" />
+    return <Navigate to="/login" state={{ from: location }} replace />;
   }
 
-  return children
-}
+  return children;
```
szilarddoro, line 31: Could you please change this to `<Outlet />`? React Router renders children of layouts via outlets.

docs/content/docs/reference/react/protecting-routes.mdx
```diff
@@ -4,49 +4,67 @@ title: 'Protecting routes'
...
-        </Route>
-      </Routes>
-    </BrowserRouter>
-  )
+    <NhostReactProvider nhost={nhost}>
+      <BrowserRouter>
+        <Routes>
+          <Route path="/" element={<Home />} />
+          <Route path="/login" element={<Login />} />
+          <Route path="/dashboard" element={<ProtectedRoute>}>
```
szilarddoro, line 61: Please close ProtectedRoute's JSX tag

docs/content/docs/reference/react/protecting-routes.mdx
```diff
@@ -4,49 +4,67 @@ title: 'Protecting routes'
...
   if (isLoading) {
-    return <div>Loading...</div>
+    return (
+      <div>Loading...</div>
+    );
   }
 
   if (!isAuthenticated) {
-    return <Redirect to="/login" />
+    return <Navigate to="/login" state={{ from: location }} replace />;
```
szilarddoro, line 28: For the sake of demonstration I think we can keep it simple and just remove the `state` prop here

szilarddoro (changes requested):
Your PR is greatly appreciated! 🙌🏼

Could you please make these minor changes before we merge it?

docs/content/docs/reference/react/protecting-routes.mdx
gdangelo: You're right. I've mixed things up since I first wrote the example with an extra `Layout` component that rendered the `Outlet`. But for simplifying the example, I removed it and only used the `ProtectedRoute`.

I'm pushing the changes.

docs/content/docs/reference/react/protecting-routes.mdx
gdangelo: Nice catch! Sorry for missing that.

docs/content/docs/reference/react/protecting-routes.mdx
gdangelo: I don't think it is adding much complexity. But you're right; let's remove it since it's the Nhost docs and not the React Router one 😉

szilarddoro: Precisely 😄  

Thanks!

szilarddoro (approved):
Thanks! 🎉 

I'll migrate your changes to our new docs as well, which will replace this documentation page soon. You can visit https://github.com/nhost/nhost/pull/392 to learn more.

PR #392, opened 2022-04-12
Migration to Docusaurus

szilarddoro:
This PR contains the code of our new documentation page based on Docusaurus.

## Reasons for the migration

We were thinking about the easiest way to add collapsible menu items, dark mode, search, or other improvements to our existing documentation. We thought there must be a better way to focus more on the content, so we decided to look for alternatives with a good reputation.

Docusaurus is a powerful alternative to our handmade documentation site. Instead of tinkering with the actual UI code, we can now focus on the quality and details of the content.


elitan (changes requested):
When I click **Reference** -> **React** the URL is `/category/react` but I would expect it to be `reference/react`.

CleanShot 2022-04-12 at 15 24 20

szilarddoro:
> When I click **Reference** -> **React** the URL is `/category/react` but I would expect it to be `/reference/react`.
> 
> CleanShot 2022-04-12 at 15 24 20

Should be fixed now

elitan (approved):

PR #435, opened 2022-04-21
Chore/include examples in monorepo

plmercereau:
Put the examples back inside the monorepo workspace.
Also ran prettier + lint fix


szilarddoro (approved):
Would be cool to have a prettier fix step in a pre-commit hook or something

PR #440, opened 2022-04-22
fix: improve reliability of the token refresher

plmercereau:
The token refresher had an unreliable behaviour, leading to too many refreshes, or refreshes that are missed, leading to an expired access token (JWT).

The internal refresher rules have been made more explicit in the code. Every second, this runs:

- If the client defined a `refreshIntervalTime` and the interval between when the last access token has been created and now is more than this value, then it triggers a refresh
- If the access token expires in less than five minutes, then it triggers a refresh

If a refresh fails, then it switches to a specific rule: it will make an attempt to refresh the token every five seconds

### DIFF
--- a/packages/core/src/machines/index.ts
+++ b/packages/core/src/machines/index.ts
@@ -4,6 +4,7 @@ import { assign, createMachine, send } from 'xstate'
 import {
   NHOST_JWT_EXPIRES_AT_KEY,
   NHOST_REFRESH_TOKEN_KEY,
+  REFRESH_TOKEN_RETRY_INTERVAL,
   TOKEN_REFRESH_MARGIN
 } from '../constants'
 import {
@@ -31,6 +32,10 @@ export * from './send-verification-email'
 export type AuthMachineOptions = {
   backendUrl: string
   clientUrl?: string
+  /**
+   * Interval in seconds before refreshing the JWT regardless of its expiration.
+   * When undefined, the option is ignored and the refresh will start 3 minutes before the access token (JWT) expiration
+   */
   refreshIntervalTime?: number
   clientStorageGetter?: StorageGetter
   clientStorageSetter?: StorageSetter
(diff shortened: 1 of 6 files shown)


packages/core/src/machines/index.ts
```diff
@@ -567,13 +583,26 @@ export const createAuthMachine = ({
...
         isAutoSignInDisabled: () => !autoSignIn,
-        refreshTimerShouldRefresh: (ctx) =>
-          ctx.refreshTimer.elapsed >
-          Math.max(
-            (Date.now() - ctx.accessToken.expiresAt.getTime()) / 1_000 - TOKEN_REFRESH_MARGIN,
-            refreshIntervalTime
-          ),
-
+        refreshTimerShouldRefresh: (ctx) => {
+          const expirestAt = ctx.accessToken.expiresAt
```
szilarddoro, line 587: We can leverage destructuring here.

```ts
const { expiresAt } = ctx.accessToken
```

packages/core/src/machines/index.ts
```diff
@@ -567,13 +583,26 @@ export const createAuthMachine = ({
...
+          if (!expirestAt) return false
```
szilarddoro, line 588: I would suggest to use curly braces here

```ts
if (!expiresAt) {
  return false
}

// ...
```

packages/hasura-auth-js/src/utils/helpers.ts
```diff
@@ -11,7 +11,13 @@ import { ClientStorage, ClientStorageType, Session } from './types'
 export const isBrowser = () => typeof window !== 'undefined'
 
 export const getSession = (context?: AuthContext): Session | null => {
-  if (!context || !context.accessToken.value || !context.refreshToken.value) return null
+  if (
+    !context ||
+    !context.accessToken.value ||
+    !context.refreshToken.value ||
+    !context.accessToken.expiresAt
+  )
```
szilarddoro, line 19: to improve readability, I'd suggest that we should use curly braces around the return statement

packages/hasura-auth-js/src/hasura-auth-client.ts
```diff
@@ -283,7 +283,7 @@ export class HasuraAuthClient {
       interpreter.send({ type: 'SIGNOUT', all: params?.all })
       interpreter.onTransition((state) => {
         if (state.matches({ authentication: { signedOut: 'success' } })) resolve({ error: null })
```
szilarddoro, line 285: to improve readability, I'd suggest that we should use curly braces around these statements

szilarddoro (commented):
LGTM 👍🏼 I have only a couple of formatting suggestions.

packages/core/src/machines/index.ts
plmercereau: ha! bad habbit :)

PR #459, opened 2022-04-25
feat: capture url errors & improved SSR / NextJs startup

### LINKED ISSUE #471: Unable to logout when signing in with Social Login because #refreshToken in URL
I'm using `@nhost/nextjs` and when using a social login to sign in, when we gets redirected to the site the refresh token stays on the URL. When I try to sign out, it does signs out but immediately signs the user back because the refreshToken is still in the URL.

plmercereau:
- Capture hasura-auth errors from the url
- get refresh token from url in Next, and remove the refresh token from the url when possible
- When `autoSignin` was active, the client was fetching the refresh token twice on startup. This behaviour has been corrected.
- Remove `refreshToken` from the url when `autoSignIn` is set
- Look for the refresh token both in the query parameters and in the hash

Closes #471


packages/core/src/utils.ts
```diff
@@ -27,3 +27,31 @@ export const rewriteRedirectTo = (
...
+  name = name.replace(/[\[\]]/g, '\\$&')
+  const regex = new RegExp('[?&#]' + name + '(=([^&#]*)|&|#|$)'),
+    results = regex.exec(url)
+  if (!results) return null
+  if (!results[2]) return ''
+  return decodeURIComponent(results[2].replace(/\+/g, ' '))
+}
+
+export function removeParameterFromWindow(name: string) {
+  const location = window?.location
```
szilarddoro, line 43: This will break in a Node environment. We should return if `window` is of the type `undefined`.
```ts
if (typeof window === 'undefined') {
  return
}

// ...
const { location } = window
```

packages/core/src/utils.ts
```diff
@@ -27,3 +27,31 @@ export const rewriteRedirectTo = (
           : options?.redirectTo
       }
     : options
+
+export function getParameterByName(name: string, url?: string) {
+  if (!url) url = window?.location?.href || ''
```
szilarddoro, line 32: This will break in a Node environment. We should return if `window` is of the type `undefined`.
```ts
if (typeof window === 'undefined') {
  return
}

// ...
const { location } = window
```

packages/nextjs/src/create-server-side-client.ts
```diff
@@ -0,0 +1,44 @@
...
+import { NhostClient } from '@nhost/nhost-js'
+
+/**
+ * Creates an Nhost client that runs on the server side.
+ * It will try to get the refesh token in cookies, or from the request URL
+ * If a refresh token is found, it uses it to get an up to date access token (JWT) and a user session
+ * This method resolves when the authentication status is known eventually
+ * @param backendUrl
+ * @param context
+ * @returns
```
szilarddoro, line 16: Could you please provide some description for this tag please?

It would be displayed in a weird way if left as is:
https://github.com/nhost/nhost/pull/459/files#diff-0f7398ca23f3fcc1358ae4887c30e6c5351df493f32b8f44c021d8333f8b7d97R15

packages/nextjs/src/create-server-side-client.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+export const createServerSideClient = async (
+  backendUrl: string,
+  context: GetServerSidePropsContext
+) => {
+  const cookies = Cookies(context.req, context.res)
+  const nhost = new NhostClient({
+    backendUrl,
+    clientStorageGetter: (key) => {
+      console.log(context.resolvedUrl, context.query, context.req.url)
```
szilarddoro, line 26: I think we can get rid of this `console.log`

packages/nextjs/src/create-server-side-client.ts
```diff
@@ -0,0 +1,44 @@
...
+ * @param backendUrl
+ * @param context
+ * @returns
+ */
+export const createServerSideClient = async (
+  backendUrl: string,
+  context: GetServerSidePropsContext
+) => {
+  const cookies = Cookies(context.req, context.res)
+  const nhost = new NhostClient({
```
szilarddoro, line 23: Would be great if people could provide some options to this internal `NhostClient`. Even if it's just a limited set of options, we should let them tweak with things like `refreshIntervalTime`, etc. We shouldn't let people override crucial options though, such as `clientStorageGetter`.

For example:
```ts
const ssrClient = createServerSideClient('https://<my-url>.nhost.run', ctx, {
  refreshIntervalTime: 5000
})
```

What do you think @plmercereau ? Is it something that we wish to use only internally?

packages/nextjs/src/get-session.ts
```diff
@@ -44,30 +43,17 @@ import { refresh } from './utils'
...
-          NHOST_JWT_EXPIRES_AT_KEY,
-          new Date(Date.now() + (session.accessTokenExpiresIn || 0) * 1_000).toISOString(),
-          {
-            httpOnly: false,
-            sameSite: true
-          }
-        )
+  const nhost = await createServerSideClient(backendUrl, context as any)
+  const { accessToken, refreshToken, user } = nhost.auth.client.interpreter!.state.context
+  // console.log(nhost.auth.getAuthenticationStatus())
```
szilarddoro, line 50: could you please remove this comment?

szilarddoro (changes requested):

packages/nextjs/src/create-server-side-client.ts
plmercereau: This method is meant to be called once when rendering on the server side, so we don't want activate the auto refresher - `refreshIntervalTime` being related to this. Plus we overload the refresh toker getter/setter, so there's no other option left for now.

packages/nextjs/src/get-session.ts
plmercereau: that's kindly asked :)
yes

szilarddoro (approved):

PR #461, opened 2022-04-25
feat: add accessors for decoded access token and Hasura claims

### LINKED ISSUE #255: Add `getClaim` to hasura-auth-js
- Users are missing an API to retrieve the value of a given custom claim. 
- Same as in here https://github.com/nhost/nhost-js-sdk/blob/9d3d34c29e4c078abb591fcdc96eee4728c2e5f6/src/Auth.ts#L275

plmercereau:
cf changesets

Closes #255

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -69,3 +69,21 @@ export type Provider =
   | 'strava'
   | 'gitlab'
   | 'bitbucket'
+
+// TODO share with hasura-auth
+export interface JWTHasuraClaims {
+  // ? does not work as expected: if the key does not start with `x-hasura-`, then it is typed as `any`
+  // [claim: `x-hasura-${string}`]: string | string[]
+  [claim: string]: string | string[] | null
+  'x-hasura-allowed-roles': string[]
+  'x-hasura-default-role': string
+  'x-hasura-user-id': string
+  'x-hasura-user-is-anonymous': string
+}
+
+// https://hasura.io/docs/1.0/graphql/core/auth/authentication/jwt.html#the-spec
+export interface JWTClaims {
+  sub?: string
+  iat?: number
+  'https://hasura.io/jwt/claims': JWTHasuraClaims
+}
--- a/packages/react/src/hooks/user.ts
+++ b/packages/react/src/hooks/user.ts
@@ -1,3 +1,4 @@
+import jwt_decode from 'jwt-decode'
 import { useMemo, useState } from 'react'
 
 import {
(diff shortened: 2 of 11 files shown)


nunopato (approved):
lgtm but I would like someone else to look at it also /cc @szilarddoro

packages/react/src/hooks/user.ts
```diff
@@ -680,3 +683,59 @@ export const useConfigMfa: ConfigMfaHook = () => {
...
+ * import { useDecodedAccessToken } from '@nhost/react'
+ * const Component = () => {
+ *    const decodedToken = useDecodedAccessToken()
+ *    return <div>Decoded access token: {JSON.stringify(decodedToken)}</div>
+ * }
+ * ```
+ */
+export const useDecodedAccessToken = () => {
+  const jwt = useAccessToken()
+  return useMemo(() => (jwt ? jwt_decode<JWTClaims>(jwt) : null), [jwt])
```
szilarddoro, line 701: `useMemo` is unnecessary here, because `jwt` is the only value that can change in this context.

packages/react/src/hooks/user.ts
```diff
@@ -680,3 +683,59 @@ export const useConfigMfa: ConfigMfaHook = () => {
...
+ * import { useHasuraClaims } from '@nhost/react'
+ * const Component = () => {
+ *    const hasuraClaims = useHasuraClaims()
+ *    return <div>JWT claims in the `https://hasura.io/jwt/claims` namespace: {JSON.stringify(hasuraClaims)}</div>
+ * }
+ * ```
+ */
+export const useHasuraClaims = () => {
+  const claims = useDecodedAccessToken()
+  return useMemo(() => claims?.['https://hasura.io/jwt/claims'] || null, [claims])
```
szilarddoro, line 718: `useMemo` is unnecessary here, because `claims` is the only value that can change in this context.

packages/react/src/hooks/user.ts
```diff
@@ -680,3 +683,59 @@ export const useConfigMfa: ConfigMfaHook = () => {
...
+ * import { useHasuraClaim } from '@nhost/react'
+ * const Component = () => {
+ *    const claim = useHasuraClaim('user-id')
+ *    return <div>User id extracted from the JWT access token: {claim}</div>
+ * }
+ * ```
+ */
+export const useHasuraClaim = (name: string) => {
+  const hasuraClaims = useHasuraClaims()
+  return useMemo(
```
szilarddoro, line 737: `useMemo` is unnecessary here, because we would recalculate the return value either way. Either if `name` or `hasuraClaims` changes.

packages/core/src/types.ts
```diff
@@ -113,3 +113,21 @@ export type Provider =
...
+  // ? does not work as expected: if the key does not start with `x-hasura-`, then it is typed as `any`
+  // [claim: `x-hasura-${string}`]: string | string[]
+  [claim: string]: string | string[] | null
+  'x-hasura-allowed-roles': string[]
+  'x-hasura-default-role': string
+  'x-hasura-user-id': string
+  'x-hasura-user-is-anonymous': string
+}
+
+// https://hasura.io/docs/1.0/graphql/core/auth/authentication/jwt.html#the-spec
```
szilarddoro, line 128: Could you please change this to a proper JSDoc? Our DocGen would pick up this comment and add this to the markdown output.

szilarddoro (changes requested):

packages/core/src/types.ts
szilarddoro: Apparently the DocGen didn't see this change because `.lintstagedrc.json` doesn't contain this package. Could you please add it to the regex?

plmercereau: done

szilarddoro (approved):

PR #477, opened 2022-04-27
DocGen improvements

szilarddoro:
This PR improves several features of our newly introduced DocGen CLI tool.

Some notable improvements and fixes:
- Static comment for all auto-generated files indicating that documentation files are not to be overwritten
- Updated the heading level of notes and examples when a function has multiple signatures
- Indicating deprecation in navigation sidebar if a class, a type or all of the signatures of a function are deprecated
- Multi-line comments will not break parameter tables


plmercereau (approved):
getting better and better!
I will change the signatures of the other hooks

szilarddoro:
> getting better and better! I will change the signatures of the other hooks

I'll merge this PR, then any changes to the signatures will be reflected in markdown files correctly.

PR #480, opened 2022-04-27
docs: enable algolia search

plmercereau:
app id / key / index does not seem to work


elitan:
Should work now @plmercereau 

image

PR #468, opened 2022-04-26
Feat/vue

plmercereau:
## `@nhost/vue`

--> [reference documentation](https://docs-git-feat-vue-nhost.vercel.app/reference/vue)

- Nhost Vue client, also defined as a Vue plugin
- Vue3 composables
- Quickstart ([doc](https://docs-git-feat-vue-nhost.vercel.app/platform/quickstarts/vue) + [source](https://github.com/nhost/nhost/tree/feat/vue/examples/vue-quickstart))
- [`vue-apollo` example](https://github.com/nhost/nhost/tree/feat/vue/examples/vue-apollo)
- Config: Vite, Eslint

## Other
- use the [same methods and typings](https://github.com/nhost/nhost/tree/feat/vue/packages/core/src/promises) to interact with xstate machines in both `@nhost/hasura-auth-js`, `@nhost/react` hooks and `@nhost/vue` composables
- simplify the auth machine in putting  fields validations in services rather than in actions and in a given state
- Add `devTools` option to the Nhost client
- Docusaurus: common `create-nhost-app` fragment used in several places in the documentation
- Minor improvements to the `react-apollo` example
- extend `@nhost/hasura-auth-js` Jest tests

PR #481, opened 2022-04-28
fix: Temporary removal of @link inline tags

szilarddoro:


szilarddoro:
@plmercereau please note that `__tsdoc__` was renamed to `.docgen`. However this won't directly affect your PR https://github.com/nhost/nhost/pull/468, in order to keep the naming consistent, you should also rename your folder and update `vue.typedoc.json` and `vue.docgen.json` accordingly.

plmercereau (approved):

PR #485, opened 2022-04-28
Add content for the Next.js quickstart guide

gdangelo:


elitan:
Vercel is unable to build these changes. Are you able to build the docs locally with your changes @gdangelo ?

plmercereau:
Also what comes to mind when using Nextjs is SSR. Maybe we could add a section about how to use `getServerSideProps`?

plmercereau:
I would also add a link to the [Nextjs reference documentation](https://docs.nhost.io/reference/nextjs), in particular the explanation on [how SSR/CSR works with auth](https://docs.nhost.io/reference/nextjs/introduction)

gdangelo:
> Also, what comes to mind when using Nextjs is SSR. Maybe we could add a section about how to use `getServerSideProps`?

> I would also add a link to the [Nextjs reference documentation](https://docs.nhost.io/reference/nextjs), in particular the explanation on [how SSR/CSR works with auth](https://docs.nhost.io/reference/nextjs/introduction)

I agree. But as it is a quickstart guide, maybe we could just add a note and link it to the Next.js reference documentation for further details on SSR. What do you think?

> Finally, there's a limitation in our Next implementation: the Apollo client cache is not hydrated, meaning the GraphQL operation will run on the client side. It's not trivial to implement as the Apollo and Next hydration patterns are quite different.

Does it mean I can't do something like this with Nhost to fetch data in SSR pages:

```js
export async function getServerSideProps() {
  const { data } = await client.query({
    query: gql`
      query Todos {
        todos {
          id
        }
      }
    `,
  });

  //...
}
```

But again, in this guide, I'm using Apollo client to fetch data on the client-side. So I'm not sure if it's worth mentioning your point here. I think this is also something to add to the reference guide if Nhost users want further details.

PR #503, opened 2022-05-03
Patch/react 18

plmercereau:
- Adapt to React 18
- Bump to Apollo client 3.6.2
- add a decent NextJs example
- fix authentication headers not being attached to GraphQL operations when authenticated
- Don't take previous errors into account when using SMS and deanonymisation


szilarddoro (commented):
I received an "Email not verified" error after signing in with a verified user when using the Next.js example. Refreshing the page solved the issue, because my access token was immediately displayed on the screen.

One more thing: It would be useful to display which menu is active in the left sidebar.

The changes look cool btw, didn't find any functional errors. ✅

plmercereau:
> I received an "Email not verified" error after signing in with a verified user when using the Next.js example. Refreshing the page solved the issue, because my access token was immediately displayed on the screen.
> 
> One more thing: It would be useful to display which menu is active in the left sidebar.
> 
> The changes look cool btw, didn't find any functional errors. ✅

Updated the example

szilarddoro (approved):

PR #507, opened 2022-05-04
Make code more consistent, implement the 'custom and 'cookie' storage types

plmercereau:


packages/react/src/provider.tsx
```diff
@@ -10,6 +10,7 @@ export const NhostReactContext = createContext<NhostClient>({} as NhostClient)
 export const NhostReactProvider: React.FC<{
```
szilarddoro, line 10: My suggestion would be to expose the interface of `NhostReactProvider` and use `PropsWithChildren` type to add children to the list of props.

```diff
- export const NhostReactProvider: React.FC<{
-   nhost: NhostClient,
-   initial?: NhostSession,
-   children?: React.ReactNode
- }> = /* ... */
+ export interface NhostReactProviderProps {
+   nhost: NhostClient
+   initial?: NhostSession
+ }
+
+ export function NhostReactProvider({
+   nhost,
+   initial,
+   ...props
+ }: PropsWithChildren<NhostReactProviderProps>) { /*...*/ }
```

plmercereau: thanks for the suggestion! done.

packages/core/src/storage.ts
```diff
@@ -28,22 +29,99 @@ export const defaultClientStorageSetter: StorageSetter = (key, value) => {
...
+      console.error(`clientStorage.${key} is not a function`)
+    }
+  })
+}
+
+export const localStorageGetter = (
+  clientStorageType: ClientStorageType,
+  clientStorage?: ClientStorage
+): StorageGetter => {
+  if (clientStorage) {
```
szilarddoro, line 47: I'd make this a bit more readable by restructuring the code:

```js
if (!clientStorage) {
  return defaultClientStorageGetter
}

if (clientStorageType === 'react-native') {
  checkStorageAccessors(...)
  return (key) => clientStorage.getItem?.(key)
}

if (clientStorageType === 'capacitor') {
  checkStorageAccessors(...)
  return (key) => clientStorage.get?.key({ key })
}

// etc...
```

packages/core/src/storage.ts
```diff
@@ -28,22 +29,99 @@ export const defaultClientStorageSetter: StorageSetter = (key, value) => {
...
-      Cookies.set(key, value)
-      // localStorage.setItem(key, value)
-    } else {
-      Cookies.remove(key)
-      // localStorage.removeItem(key)
+export const localStorageSetter = (
+  clientStorageType: ClientStorageType,
+  clientStorage?: ClientStorage
+): StorageSetter => {
+  if (clientStorage) {
```
szilarddoro, line 84: see my previous comment

packages/core/src/storage.ts
```diff
@@ -28,22 +29,99 @@ export const defaultClientStorageSetter: StorageSetter = (key, value) => {
...
+  })
+}
+
+export const localStorageGetter = (
+  clientStorageType: ClientStorageType,
+  clientStorage?: ClientStorage
+): StorageGetter => {
+  if (clientStorage) {
+    if (clientStorageType === 'react-native') {
+      checkStorageAccessors(clientStorage, ['getItem'])
```
szilarddoro, line 49: wouldn't it make sense to only execute this function if we are in a debug mode or something?

packages/core/src/types.ts
```diff
@@ -3,7 +3,27 @@
 import { InterpreterFrom } from 'xstate'
 
 import { AuthMachine } from './machines'
+import { StorageGetter, StorageSetter } from './storage'
 
+export interface AuthOptions {
+  /** Time interval until token refreshes, in seconds */
+  refreshIntervalTime?: number
+  clientStorage?: ClientStorage
```
szilarddoro, line 11: Could you please provide a short description for this attribute?

szilarddoro (changes requested):
I think we can improve the readability a bit.

packages/core/src/storage.ts
plmercereau: yes. legacy code. Will comment it for now on

PR #512, opened 2022-05-05
chore: prettier

plmercereau:
may wait until Nuno's review before merging


szilarddoro:
I was thinking about incorporating our Prettier configuration into DocGen. It'll always have a different output than what our formatter script would expect and therefore it'd always change the auto-generated markdown files.

I'm also struggling a lot with the import sorting. We should make our Prettier plugin's configuration ([@trivago/prettier-plugin-sort-imports](https://github.com/trivago/prettier-plugin-sort-imports)) consistent with our ESLint configuration.

szilarddoro:
I think we should revisit this issue at some point during a cooldown week as this PR has too many conflicts with the main branch. Do you mind if we close this PR?

plmercereau:
Yeah, better to close it and to start again from the main branch

PR #525, opened 2022-05-06
fix: correct cookie storage type

plmercereau:


packages/core/src/storage.ts
```diff
@@ -71,43 +58,46 @@ export const localStorageGetter = (
       }
     }
   }
-  if (clientStorageType === 'custom') {
-    if (clientStorage.getItem && clientStorage.removeItem) {
-      return clientStorage.getItem
+
+  if (clientStorage) {
```
szilarddoro, line 62: I think we should perform an early exit here if `clientStorage` is undefined. We can make the code more readable if we reduce the depth of blocks IMO.

packages/core/src/storage.ts
```diff
@@ -119,23 +109,44 @@ export const localStorageSetter = (
...
-  if (clientStorageType === 'custom') {
-    if (!clientStorage.removeItem) {
-      throw Error(
-        `clientStorageType is set to 'custom' but clientStorage is missing a removeItem property`
-      )
-    }
-    if (clientStorage.setItem) {
+  if (clientStorage) {
+    if (clientStorageType === 'react-native') {
+      // checkStorageAccessors(clientStorage, ['setItem', 'removeItem'])
```
szilarddoro, line 114: can we get rid of these comments?

packages/core/src/storage.ts
```diff
@@ -89,36 +77,37 @@ export const localStorageSetter = (
...
+      if (isBrowser) {
+        if (value) {
+          Cookies.set(key, value)
+        } else {
+          Cookies.remove(key)
+        }
+      }
+    }
+  }
+  if (!clientStorage) {
```
szilarddoro, line 94: duplicate if statement, we already check if clientStorage exists

packages/core/src/storage.ts
```diff
@@ -89,36 +77,37 @@ export const localStorageSetter = (
   clientStorageType: ClientStorageType,
   clientStorage?: ClientStorage
 ): StorageSetter => {
-  if (!clientStorage || clientStorageType === 'localStorage' || clientStorageType === 'web') {
+  if (clientStorageType === 'localStorage' || clientStorageType === 'web') {
     return defaultClientStorageSetter
   }
-
+  if (clientStorageType === 'cookie') {
```
szilarddoro, line 83: as per our discussion, `cookie` type can be checked at the beginning of this function

szilarddoro (approved):

PR #543, opened 2022-05-10
Fix/signup states

plmercereau:
- Correct `useSignUpEmailPassword` loading state
- Correct email verification states: email verification returns `false` when the client has started and it is still undetermined
- Minor improvements in the examples
- Improve readability in the authentication hooks


packages/react/src/hooks/registration.ts
```diff
@@ -111,21 +109,32 @@ export const useSignUpEmailPassword: SignUpEmailPasswordHook = (
   const stateOptions = c || (typeof a !== 'string' ? a : undefined)
 
   const service = useAuthInterpreter()
-  const isError =
-    !!service.status && service.state.matches({ authentication: { signedOut: 'failed' } })
+  const isError = useSelector(
+    service,
+    (state) => !state.matches({ authentication: { signedOut: 'failed' } })
```
elitan, line 114: Should this really be `!state.matches` and not `state.matches` without the `!` in the beginning?

elitan: Other than that it looks good! 👍

elitan (approved):

PR #557, opened 2022-05-16
Testing framework

szilarddoro:
The `@nhost/core` package contains the implementation of a quite complex authentication flow using an evergrowing state machine. It's essential to automated mechanisms here to protect us from the introduction of unintended side-effect during development.


plmercereau (approved):
👌

PR #566, opened 2022-05-18
accept query parameters from both relative `redirectTo` URLs and the client URL

plmercereau:
URLs where malformed When the client URL was defining query parameters and a relative `redirectTo` URL was passed on as an option.
It is now possible to define query parameters in both the base client URL and the relative `redirectTo`

### DIFF
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -18,30 +18,29 @@ export const encodeQueryParameters = (baseUrl: string, parameters?: Record<strin
 }
 
 export const rewriteRedirectTo = <T extends RedirectOption>(clientUrl: string, options?: T) => {
-  if (options?.redirectTo) {
-    const baseClientUrl = new URL(clientUrl)
-    const clientParams = Object.fromEntries(new URLSearchParams(baseClientUrl.search))
-    const url = new URL(
-      options.redirectTo.startsWith('/')
-        ? baseClientUrl.origin + options.redirectTo
-        : options.redirectTo
-    )
-    const additionalParams = new URLSearchParams(url.search)
-    let combinedParams = Object.fromEntries(additionalParams)
-
-    if (options.redirectTo.startsWith('/')) {
-      combinedParams = { ...clientParams, ...combinedParams }
-    }
-    let pathName = baseClientUrl.pathname
-    if (url.pathname.length > 1) {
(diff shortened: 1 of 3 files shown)


packages/core/src/utils.ts
```diff
@@ -17,15 +17,33 @@ export const encodeQueryParameters = (baseUrl: string, parameters?: Record<strin
...
-  options?.redirectTo
-    ? {
-        ...options,
-        redirectTo: options?.redirectTo?.startsWith('/')
-          ? clientUrl + options.redirectTo
-          : options?.redirectTo
-      }
-    : options
+export const rewriteRedirectTo = <T extends RedirectOption>(clientUrl: string, options?: T) => {
+  if (options?.redirectTo) {
```
szilarddoro, line 21: I don't know if it's just me, but reversing the logic of if statements here makes it easier on the eye to me:

```ts
if (!options?.redirectTo) {
  return options
}

// then we can proceed without the else branch...
const baseClientUrl = new URL(clientUrl)
// ...
```

packages/core/tests/utils.test.ts
```diff
@@ -0,0 +1,70 @@
+import { RedirectOption } from '../src/types'
+import { rewriteRedirectTo } from '../src/utils'
+
+test(`should not add redirectTo when none is given`, async () => {
```
szilarddoro, line 4: Awesome 👍🏼

packages/core/src/utils.ts
```diff
@@ -17,15 +17,33 @@ export const encodeQueryParameters = (baseUrl: string, parameters?: Record<strin
...
-export const rewriteRedirectTo = <T extends RedirectOption>(clientUrl: string, options?: T) =>
-  options?.redirectTo
-    ? {
-        ...options,
-        redirectTo: options?.redirectTo?.startsWith('/')
-          ? clientUrl + options.redirectTo
-          : options?.redirectTo
-      }
-    : options
+export const rewriteRedirectTo = <T extends RedirectOption>(clientUrl: string, options?: T) => {
```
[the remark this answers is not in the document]
plmercereau: why not!

szilarddoro (approved):

GavanWilhite:
Thanks for the quick turnaround folks!

PR #581, opened 2022-05-20
Docs: Database and Event Triggers

elitan:


plmercereau (approved):
Nice!
A suggestion for the future: to explain the two different triggers, what are the pros and cons and when it is recommended to use them (sql triggers: for simple processing, internal to the db, ran synchronously vs event triggers, for more complex tasks and interactions outside of the DB, ran asynchronously)

plmercereau:
we could also add a section with selected links to hasura and postgres resources such as some from
- https://hasura.io/docs/latest/graphql/core/guides/index/
- https://hasura.io/learn/
- https://hasura.io/blog/tutorials/

elitan:
Those are great ideas @plmercereau. I'm now only going over the docs on a basic level. I'll come back to more advanced topics in the future if no one else does it before.

PR #582, opened 2022-05-20
Anonymous users

plmercereau:
See changesets


packages/vue/src/useSignInAnonymous.ts
```diff
@@ -5,10 +5,29 @@ import { useAuthenticated } from './useAuthenticated'
...
 
-// TODO documentation when available in Nhost Cloud - see changelog
-// TODO deanonymize
-// TODO review nhost.auth.signIn()
-/** @internal not ready yet */
+/**
+ * Use the hook `useSignInAnonymous` to sign in a user anonymously.
+ *
+ * As a result, the user will have the `anonymous` role and subsequent set of permissions.
+ * The user can then register at a later stage using email+password sign-up, passwordless email (magic link), or passwordless SMS.
```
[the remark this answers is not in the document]
plmercereau: 👍

packages/core/src/machines/index.ts
```diff
@@ -385,23 +302,106 @@ export const createAuthMachine = ({
...
+                  }
+                ],
+                onError: [
+                  {
+                    cond: 'unverified',
+                    target: 'incomplete.needsEmailVerification'
+                  },
+                  {
+                    actions: 'saveRegistrationError',
+                    target: 'incomplete.failed'
```
szilarddoro, line 349: I don't think we should treat failed registration as incomplete. A user would restart the whole registration workflow if there was an error during the process, wouldn't they?

`registration.failed` or `registration.error` makes more sense to me here

packages/core/src/machines/index.ts
```diff
@@ -385,23 +302,106 @@ export const createAuthMachine = ({
...
+              invoke: {
+                src: 'passwordlessEmail',
+                id: 'passwordlessEmail',
+                onDone: {
+                  actions: 'clearContext',
+                  target: ['#nhost.authentication.signedOut', 'incomplete.needsEmailVerification']
+                },
+                onError: {
+                  actions: 'saveRegistrationError',
+                  target: 'incomplete.failed'
```
szilarddoro, line 365: Same as my other commend regarding "incomplete vs. failed"

packages/core/src/machines/index.ts
szilarddoro: If refactoring the logic here would require too many changes, then we can come back to this at a later time.

packages/core/src/machines/index.ts
```diff
@@ -385,23 +302,106 @@ export const createAuthMachine = ({
...
+              invoke: {
+                src: 'passwordlessSms',
+                id: 'passwordlessSms',
+                onDone: {
+                  actions: 'clearContext',
+                  target: ['#nhost.authentication.signedOut', 'incomplete.needsOtp']
+                },
+                onError: {
+                  actions: 'saveRegistrationError',
+                  target: 'incomplete.failed'
```
szilarddoro, line 380: Same as my other commend regarding "incomplete vs. failed"

packages/core/src/machines/index.ts
```diff
@@ -385,23 +302,106 @@ export const createAuthMachine = ({
...
+              invoke: {
+                src: 'passwordlessSmsOtp',
+                id: 'passwordlessSmsOtp',
+                onDone: {
+                  actions: ['saveSession', 'reportTokenChanged'],
+                  target: '#nhost.authentication.signedIn'
+                },
+                onError: {
+                  actions: 'saveRegistrationError',
+                  target: 'incomplete.failed'
```
szilarddoro, line 395: Same as my other commend regarding "incomplete vs. failed"

szilarddoro (changes requested):

packages/core/tests/anonymousSignIn.test.ts
```diff
@@ -0,0 +1,142 @@
...
+        "error": "OK",
+        "message": "Network Error",
+        "status": 200,
+      },
+    }
+  `)
+})
+
+test('should deanonymise a user with email and password', async () => {
+  server.use(correctAnonymousHandler, deamonymisationSuccessfulHandler)
```
szilarddoro, line 81: you can include these handlers in `handlers/server.ts` as default handlers for successful scenarios

packages/core/tests/anonymousSignIn.test.ts
```diff
@@ -0,0 +1,142 @@
...
+      "authentication": {
+        "error": "OK",
+        "message": "Network Error",
+        "status": 200,
+      },
+    }
+  `)
+})
+
+test(`should fail if server returns an error`, async () => {
```
szilarddoro, line 60: This test is the same as the first one. The handler should be changed here.

packages/core/tests/anonymousSignIn.test.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  const state: AuthState = await waitFor(authService, (state: AuthState) =>
+    state.matches({
+      authentication: 'signedOut',
+      registration: { incomplete: 'needsOtp' }
+    })
+  )
+
+  expect(state.context.user).toBeNull()
+})
```
szilarddoro, line 142: What happens if the deanonymization process is not successful? Could you please also cover that scenario with a couple of tests?

szilarddoro (changes requested):

docs/docs/reference/docgen/react/content/use-sign-in-anonymous.mdx
```diff
@@ -3,8 +3,26 @@
...
+Use the hook `useSignInAnonymous` to sign in a user anonymously.
+
+As a result, the user will have the `anonymous` role and subsequent set of permissions.
+The user can then be converted as a regular user at a later stage using email+password sign-up, passwordless email (magic link), or passwordless SMS.
+
+```tsx
+const { signInAnonymous, isLoading, isSuccess, isError, error } =
+  useSignInAnonymous()
+
+console.log({ isLoading, isSuccess, isError, error })
```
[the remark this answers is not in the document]
plmercereau: No, this is how Yohan did for the other hooks in his last documentation updates. I have no opinion expect to document things in a consistent way

packages/core/src/machines/index.ts
plmercereau: The logic is: a failed registration is incomplete.
The main reason why it is structured that way is to share transitions with other incomplete sub-states: `SIGNUP_EMAIL_PASSWORD`, `PASSWORDLESS_EMAIL` etc.
These transitions are on the `incomplete` state so we avoid running them when a registration step is already ongoing (for instance `passwordlessEmail`), or when registration is complete.

We could copy-paste every transition in a new `failed` state and refactor, but I'd say we keep it that way and we can change this on a separate PR

szilarddoro: To me an incomplete state would be a state where we already have some progress, but has not been finished for some reason (such as the email verification). I agree that we should not necessarily change it in this PR.

It'd also be great to see the opinion of @elitan / @nunopato regarding this.

docs/docs/reference/docgen/react/content/use-sign-in-anonymous.mdx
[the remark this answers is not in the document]
szilarddoro: Yes, I agree that we should be consistent. 👍🏼

packages/core/src/machines/index.ts
elitan: I don't know what "incomplete" means in this context.

I don' know what "share transitions with other incomplete sub-states" means.

Generally, I'm so lost in this xstate / machine code to the point that I can not contribute with anything meaningful. Sorry. 

I feel this is over-engineered. But that's a different topic.

szilarddoro: I'd suggest we should look into this issue later. I understand your concern about `xstate`, especially regarding the complexity it brings seemingly, although I feel that it helps a lot with transitions between different states of the SDK. We should definitely look for ways to make the code more compact.

szilarddoro (approved):

PR #601, opened 2022-05-24
Add `subdomain` and `region` as options to `NhostClient` 

nunopato:
This PR adds `subdomain` and `region` as new arguments to the `NhostClient` constructor. `backend_url` will eventually be deprecated.


elitan:
Might cause problems if people use `http://localhost:1337` as `backendUrl` and adds a region to the constructor.

packages/nhost-js/src/core/nhost-client.ts
```diff
@@ -39,8 +41,12 @@ export class NhostClient {
     start = true
   }: NhostClientConstructorParams) {
     if (!backendUrl) throw new Error('Please specify a `backendUrl`. Docs: [todo]!')
+
+    const subdomain = new URL(backendUrl).host.split('.')[0]
```
elitan, line 45: Might cause problems if people use http://localhost:1337 as backendUrl and adds a region to the constructor.

packages/nhost-js/src/core/nhost-client.ts
```diff
@@ -6,9 +6,10 @@ import { NhostGraphqlClient } from '../clients/graphql'
 
 export interface NhostClientConstructorParams extends Omit<NhostAuthConstructorParams, 'url'> {
   /**
-   * Nhost backend URL.
+   * Nhost backend URL and Region.
```
plmercereau, line 9: Each option requires its own annotation. This one will go to `backendUrl` while there will be no information about `region` in the generated and inline TS documentation

packages/nhost-js/src/core/nhost-client.ts
nunopato: this was removed because we decided to also have `subdomain` as an option.

packages/nhost-js/src/core/nhost-client.ts
```diff
@@ -6,9 +6,21 @@ import { NhostGraphqlClient } from '../clients/graphql'
...
+   * Nhost backend URL
+   * Will be deprecated in favor of `subdomain` and `region`
+   */
+  backendUrl?: string
+
+  /**
+   * App subdomain (e.g., ifieniwenfiwenwng)
+   * Use `localhost` in development
+   */
+  subdomain?: string
```
plmercereau, line 18: ideally the typings should be review so we require either  `{ backendUrl }` or `{ subdomain, region }`, but not both, for instance in the way we do it for `nhost.auth.signIn`

packages/nhost-js/src/core/nhost-client.ts
```diff
@@ -40,13 +53,27 @@ export class NhostClient {
     devTools,
     start = true
   }: NhostClientConstructorParams) {
-    if (!backendUrl) throw new Error('Please specify a `backendUrl`. Docs: [todo]!')
+    if (!backendUrl && !subdomain)
```
plmercereau, line 56: with the above suggestion, typings would not let this case to happen

packages/nhost-js/src/core/nhost-client.ts
```diff
@@ -40,13 +53,27 @@ export class NhostClient {
...
+
+    const urlFromEnv = (
+      backendUrl?: string,
+      subdomain?: string,
+      region?: string,
+      service?: string
+    ) => {
+      if (backendUrl) return `${backendUrl}/v1/${service}`
+
+      if (subdomain === 'localhost') return `http://${subdomain}`
```
plmercereau, line 70: I'm not sure I understood. Assuming the CLI works in the same way as of today, and the user wants to use the CDN in cloud but to develop locally, his config would be something like
```js
{
  subdomain: 'localhost'
  region: 'eu-central-1' 
}
```

but then, local hasura-auth would be `http://localhost` while it should be `http://localhost:1337/v1/auth`

plmercereau (changes requested):

packages/nhost-js/src/core/nhost-client.ts
```diff
@@ -38,9 +53,27 @@ export class NhostClient {
     devTools,
     start = true
   }: NhostClientConstructorParams) {
-    if (!backendUrl) throw new Error('Please specify a `backendUrl`. Docs: [todo]!')
+    if (!backendUrl && !subdomain)
+      throw new Error('Please specify either `backendUrl` or `subdomain`. Docs: [todo]!')
+
+    if (subdomain && subdomain !== 'localhost' && !region)
```
plmercereau, line 59: won't work with IPs e.g. `127.0.0.1`

packages/nhost-js/src/core/nhost-client.ts
nunopato: added a bunch of types :+1:

packages/nhost-js/src/utils/types.ts
```diff
@@ -0,0 +1,57 @@
+import type { AxiosResponse } from 'axios'
```
plmercereau, line 1: I would rather keep types in `src/types` as it is the most common place to put typings that I've seen in ts libraries so far

plmercereau (approved):

plmercereau:
Please add a changeset

packages/nhost-js/src/core/nhost-client.ts
```diff
@@ -27,12 +17,11 @@ export class NhostClient {
    * Nhost Client
    *
    * @example
-   * const nhost = new NhostClient({ url });
+   * const nhost = new NhostClient({ subdomain, region });
```
szilarddoro, line 20: Could you please wrap this in backticks and put `ts` to make it properly styled in the docs?

szilarddoro (dismissed):

guicurcio:
In the current PR, we're using the subdomain to indicate localhost:

```
{ subdomain: 'localhost' }
{ subdomain: 'localhost:1337' }
 ```

**Comment by nunopato** [comment]
> One idea is to use region: 'localhost', like this:

`subdomain` makes more sense conceptually speaking because `localhost` it is a domain. `Region` is a concept that only exists in the Nhost Cloud offering.

File: .changeset/pink-kids-dream.md [hunk 9]
```diff
@@ -0,0 +1,5 @@
+---
+'@nhost/nhost-js': patch
+---
+
+Introduce `subdomain` and `region` alongside the soon-to-be deprecated `backendUrl` argument.
```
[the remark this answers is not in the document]
nunopato: we will keep `backendUrl` for self-hosting only and will have a "cut off" date for folks to stop using it with Nhost cloud (`subdomain` and `region` should be used instead)

elitan (approved):

elitan:
Updated NhostClient parameters everywhere except on quickstart guides as they require the new UI for the Nhost Dashboard. Once the new Nhost Dashboard is released showing subdomain and region we'll go back and update the guides.

PR #607, opened 2022-05-26
refactor: simpler and more consistent file structure in @nhost/react

plmercereau:
Keep the same logic as in Vue: 1 hook = 1 file, file name = hook name


packages/react/src/useChangeEmail.ts
```diff
@@ -0,0 +1,76 @@
...
+
+import { useNhostClient } from './useNhostClient'
+
+interface ChangeEmailHandler {
+  (email: string, options?: ChangeEmailOptions): Promise<ChangeEmailHandlerResult>
+  /** @deprecated */
+  (email?: unknown, options?: ChangeEmailOptions): Promise<ChangeEmailHandlerResult>
+}
+
+interface ChangeEmailHookResult extends ChangeEmailState {
```
szilarddoro, line 20: Please export this interface. This change should be applied to all the other `Result` interfaces.

My reasoning here is simple: We should make enhancing the behaviour of our hooks as easy to the users of our SDK as possible. By exporting everything that can help this will make our SDK a lot more flexible.

packages/react/src/useProviderLink.ts
```diff
@@ -0,0 +1,46 @@
...
+    () =>
+      new Proxy({} as Record<Provider, string>, {
+        get(_, provider: string) {
+          return encodeQueryParameters(
+            `${nhost.auth.client.backendUrl}/signin/provider/${provider}`,
+            rewriteRedirectTo(nhost.auth.client.clientUrl, options as any)
+          )
+        }
+      }),
+    [nhost, options]
```
szilarddoro, line 44: Considering that these are the only two values in this block that can change, you can remove the `useMemo` call.

szilarddoro (changes requested):

szilarddoro (approved):

PR #643, opened 2022-05-31
Light mode for now to match landing page and dashboard

elitan:


elitan:
Still a bit 😅 

CleanShot 2022-05-31 at 17 07 20

PR #648, opened 2022-06-01
ESM and build process improvements

szilarddoro:
This is a follow-up PR that patches a couple of remaining imports having no extensions and therefore causing issues with projects using `@nhost/apollo` package.

The process of copying `.esm.js` files to `.esm.mjs` was moved to `vite.lib.config.js` instead of running things via `package.json` scripts. This way we can change the behaviour in one place and it also works if packages are built in development mode.


packages/apollo/src/index.ts
```diff
@@ -7,10 +7,10 @@ import {
   RequestHandler,
   split,
   WatchQueryFetchPolicy
-} from '@apollo/client/core'
-import { setContext } from '@apollo/client/link/context'
-import { GraphQLWsLink } from '@apollo/client/link/subscriptions'
-import { getMainDefinition } from '@apollo/client/utilities'
+} from '@apollo/client/core/index.js'
```
plmercereau, line 10: why do we have now to give the full path of the js file?
If this is mandatory, is there a way to raise an error when building the package? For the moment, both vs code / eslint nor `pnpm build` accept a value like `@apollo/client/core`

plmercereau (approved):

PR #650, opened 2022-06-01
docs: Add skip option to useQuery call to skip query if id is falsy

gdangelo:


plmercereau (approved):

elitan:
Should we do the same for other guides, like Next.js?

gdangelo:
> Should we do the same for other guides, like Next.js?

#651

PR #663, opened 2022-06-02
e2e tests: groundwork on the react-apollo example

plmercereau:
- Basic tests: email + password sign-up, including email verification, and failed attempt to sign-up when the email is already in use.
- The hierarchy and structure of the test files in `cypress/e2e` will likely change in the future
- Next, I want to focus on adjusting the GH action

### Design
- cypress and tests dependencies are part of the example package so the example can be copy-pasted outside of the monorepo
- the nhost backend that is used remains in the example directory because of the above, and also because examples don't necessarily use the same schema and configuration
- as a result, each example package would spawn its own nhost cli process. It would be orchestrated by a root command like `pnpm/turborepo --filter="@nhost-examples/*" ci` that would be called from a GH action on each PR
- I intentionally did not use the attribute [`data-cy` as a selector](https://docs.cypress.io/guides/references/best-practices#Selecting-Elements), so we don't sacrifice the readability of the example code

### Changes on global setup
- adapted base eslint rules to test files

### Other changes in the `react-apollo` example
- fixed all the linting problems
- bumped hasura, hasura-auth, and hasura-storage
- removed references outside of the example folder so it can be copy-pasted outside of the monorepo (tsconfig)
- fixed warnings on install and build


examples/react-apollo/cypress/support/e2e.ts
```diff
@@ -0,0 +1,33 @@
...
+      signUpEmailPassword(email: string, password: string): Chainable<Element>
+      confirmEmail(email: string): Chainable<Element>
+    }
+  }
+}
+
+Cypress.Commands.add('signUpEmailPassword', (email, password) => {
+  cy.visit('/sign-up')
+  cy.contains('Continue with email + password').click()
+  cy.get('[placeholder="First name"]').type(faker.name.firstName())
```
szilarddoro, line 17: I think we should incorporate https://github.com/testing-library/cypress-testing-library. Forgot to mention this library when we were discussing the approaches we will take.

Using this library would make our queries a lot more declarative:
```ts
cy.get('[placeholder="Last name"]')
```
would become
```ts
cy.findByPlaceholderText('Last name')
```

There are tons of useful [queries](https://testing-library.com/docs/queries/about) that we could use in the future.

examples/react-apollo/cypress/support/e2e.ts
```diff
@@ -0,0 +1,33 @@
+import { faker } from '@faker-js/faker'
+
+import 'cypress-mailhog'
```
szilarddoro, line 3: wow, this is awesome

examples/react-apollo/cypress/e2e/1-sign-up/successful-email-password.cy.ts
```diff
@@ -0,0 +1,15 @@
...
+  it('should redirect to /sign-in when not authenticated', () => {
+    cy.visit('/')
+    cy.location('pathname').should('equal', '/sign-in')
+  })
+
+  it('should sign-up with email and password', () => {
+    const email = faker.internet.email()
+    cy.signUpEmailPassword(email, faker.internet.password())
+    cy.contains('Verification email sent').should('be.visible')
+    cy.confirmEmail(email)
```
szilarddoro, line 13: We should have the assertion here instead of in the `confirmEmail` function.

examples/react-apollo/cypress/e2e/1-sign-up/existing-email.cy.ts
```diff
@@ -0,0 +1,11 @@
+import { faker } from '@faker-js/faker'
+
+context('Failed attempt to sign up with an email already present in the database', () => {
+  it('shoud raise an error when trying to sign up with an existing email', () => {
+    const email = faker.internet.email()
+    const password = faker.internet.password(10)
+    cy.signUpEmailPassword(email, password)
```
szilarddoro, line 7: I'd put some assertions here to confirm that registration was successful at first attempt.

szilarddoro (changes requested):
Good job 👍🏼

The next PR should be about running these tests in CI.

examples/react-apollo/cypress/support/e2e.ts
plmercereau: nice

szilarddoro (approved):
Let's wait for Nuno's review as well, but it looks good to me.

PR #667, opened 2022-06-03
fix: correct rewriting options when `clientUrl` is not available

plmercereau:
The client URL is set to `window.location.origin`, so it can rewrite redirection urls that are
passed on to authenticaion methods. However, `clientUrl` is set to `''` when running on the server
side. This fix then avoid raising an error when trying to rewrite `redirectTo` on non-browser
environment, and forces `useProviderLink` to be rendered on the client side.

### DIFF
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -17,14 +17,26 @@ export const encodeQueryParameters = (baseUrl: string, parameters?: Record<strin
   else return baseUrl
 }
 
-export const rewriteRedirectTo = <T extends RedirectOption>(clientUrl?: string, options?: T) => {
+/**
+ * Transform options that include a redirectTo property so the
+ * redirect url is absolute, given a base clientUrl.
+ * If no client url is given, any relative redirectUrl is removed while
+ * the other options are sent as-is.
+ * @param clientUrl base client url
+ * @param options
+ * @returns
+ */
+export const rewriteRedirectTo = <T extends RedirectOption>(
+  clientUrl?: string,
+  options?: T
+): (Omit<T, 'redirectTo'> & { redirectTo?: string }) | undefined => {
   if (!options?.redirectTo) {
     return options
   }
   const { redirectTo, ...otherOptions } = options
-  // If the clientUrl is not defined, we can't rewrite the redirectTo
+  // * If the clientUrl is not defined, we can't rewrite the redirectTo
   if (!clientUrl) {
-    // If redirectTo is a relative path, we therefore pull it out of the options
+    // * If redirectTo is a relative path, we therefore pull it out of the options
     if (redirectTo.startsWith('/')) {
       return otherOptions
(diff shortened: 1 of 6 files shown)


packages/react/src/useProviderLink.ts
```diff
@@ -29,13 +29,25 @@ import { NhostReactContext } from './provider'
  * ```
  */
 export const useProviderLink = (options?: ProviderOptions) => {
+  /**
+   * @internal When using Nextjs or any SSR framework, nhost.auth.client.clientUrl will be set to `undefined`
+   * as its value is set to window.location.origin.
+   * This is because the request context is not available when setting up the client `new NhostClient()` outside of
+   * the React/Nextjs context.
+   */
+  const [isSSR, setIsSSR] = useState(true)
```
szilarddoro, line 38: Shouldn't SSR related logic be part of `@nhost/nextjs` instead?

By having SSR related features in `@nhost/react` makes me think if we really need a dedicated separate package for Next.js or just have an `@nhost/react` package that takes care about these kind of things.

packages/core/src/utils.ts
```diff
@@ -17,29 +17,35 @@ export const encodeQueryParameters = (baseUrl: string, parameters?: Record<strin
   else return baseUrl
 }
 
-export const rewriteRedirectTo = <T extends RedirectOption>(clientUrl: string, options?: T) => {
+export const rewriteRedirectTo = <T extends RedirectOption>(clientUrl?: string, options?: T) => {
   if (!options?.redirectTo) {
     return options
   }
+  const { redirectTo, ...otherOptions } = options
```
szilarddoro, line 24: Could you please extend the `utils.test.ts` test suite with these changes?

packages/core/src/utils.ts
```diff
@@ -17,29 +17,35 @@ export const encodeQueryParameters = (baseUrl: string, parameters?: Record<strin
   else return baseUrl
 }
 
-export const rewriteRedirectTo = <T extends RedirectOption>(clientUrl: string, options?: T) => {
+export const rewriteRedirectTo = <T extends RedirectOption>(clientUrl?: string, options?: T) => {
```
szilarddoro, line 20: Could you please provide description for this function?

packages/react/src/useProviderLink.ts
plmercereau: > Shouldn't SSR related logic be part of `@nhost/nextjs` instead?

maybe, although other React-based frameworks support ssr.

> By having SSR related features in `@nhost/react` makes me think if we really need a dedicated separate package for Next.js or just have an `@nhost/react` package that takes care about these kind of things.

I think there's a separate project to refactor `@nhost/nextjs` with its possible specifics. For instance to be able to get the client url from either the request context on the server-side or from window.location on the client side

szilarddoro (approved):

PR #676, opened 2022-06-06
Docs abut emails with CLI

elitan:


plmercereau (approved):
Not behind my computer,.but I think mailhog port is attributed dynamically. Maybe it's worth setting a default in the CLI config, and/or explaining how to set a fixed port in the CLI config file

PR #690, opened 2022-06-07
docs: add `useSignOut` inline example in `@nhost/vue`

plmercereau:


packages/vue/src/useSignOut.ts
```diff
@@ -9,6 +9,18 @@ import { useAuthInterpreter } from './useAuthInterpreter'
 /**
  * Use the composable `useSignOut` to sign out the user.
  *
+ * @example
+ * ```jsx
+ * import { useSignOut } from '@nhost/vue'
+ *
+ * const { signOut, isSuccess } = useSignOut()
+ *
+ * const handleSignOut = (e) => {
```
szilarddoro, line 18: This function needs to be asynchronous to be able to use the `await` keyword.

packages/vue/src/useSignOut.ts
```diff
@@ -9,6 +9,18 @@ import { useAuthInterpreter } from './useAuthInterpreter'
...
+ *   e.preventDefault();
```
szilarddoro, line 19: Please remove semicolon to keep code format consistent

szilarddoro (changes requested):

szilarddoro (approved):

PR #691, opened 2022-06-08
test: passwordless email, sign-in with token, sign-out

plmercereau:


.github/workflows/tests.yaml
```diff
@@ -67,12 +67,18 @@ jobs:
...
+        run: |
+          PACKAGE_FILE_NAME=$(echo "${{ matrix.package.name }}" | sed 's/@//g; s/\//-/g')
+          echo "::set-output name=fileName::$PACKAGE_FILE_NAME"
       # * Run this step only if the previous step failed, and some Cypress screenshots/videos exist
       - name: Upload Cypress videos and screenshots
         if: ${{ failure() && hashFiles(format('{0}/cypress/screenshots/**', matrix.package.path), format('{0}/cypress/videos/**', matrix.package.path)) != ''}}
         uses: actions/upload-artifact@v3
         with:
-          name: cypress-${{ matrix.package.name }}
+          name: cypress-${{ steps.file-name.outputs.fileName }}
```
plmercereau, line 81: `${{ matrix.package.name }}` cannot be used in a file name as it contains `/`. The package name is therefore transformed in the `file-name` step: `/` are replaced by `-`, and `@` are removed

examples/react-apollo/cypress/e2e/3-sign-in/email-password.cy.ts
```diff
@@ -0,0 +1,24 @@
...
+  const email = faker.internet.email()
+  const password = faker.internet.password()
+
+  it('should sign-up', () => {
+    cy.signUpEmailPassword(email, password)
+    cy.contains('Verification email sent').should('be.visible')
+    cy.confirmEmail(email)
+  })
+
+  it('should sign-in with email and password', function () {
```
szilarddoro, line 13: change this to arrow function please

examples/react-apollo/cypress/e2e/3-sign-in/token.cy.ts
```diff
@@ -0,0 +1,17 @@
+context('Sign in with a refresh token', () => {
+  it('should sign-up', () => {
+    cy.quickSignUp()
+  })
+
+  it('should sign-in with a refresh token', function () {
```
szilarddoro, line 6: change this to arrow function please

examples/react-apollo/cypress/e2e/4-authenticated/sign-out.cy.ts
```diff
@@ -0,0 +1,13 @@
+context('Sign out', () => {
+  it('should sign-up', () => {
+    cy.quickSignUp()
+  })
+
+  it('should sign out', function () {
```
szilarddoro, line 6: change this to arrow function please

examples/react-apollo/cypress/e2e/4-authenticated/sign-out.cy.ts
```diff
@@ -0,0 +1,13 @@
+context('Sign out', () => {
+  it('should sign-up', () => {
+    cy.quickSignUp()
```
szilarddoro, line 3: we should have an assertion here or move this to `beforeEach` instead

examples/react-apollo/cypress/support/e2e.ts
```diff
@@ -22,10 +40,54 @@ Cypress.Commands.add('signUpEmailPassword', (email, password) => {
...
+Cypress.Commands.add('signInEmailPassword', (email, password) => {
+  cy.visit('/sign-in')
+  cy.contains('Continue with email + password').click()
+  cy.findByPlaceholderText('Email Address').type(email)
+  cy.findByPlaceholderText('Password').type(password)
+  cy.contains('Sign in').click()
+  cy.saveRefreshToken()
+})
+
+Cypress.Commands.add('signInToken', function (path = '/') {
```
szilarddoro, line 59: `visitPathWithRefreshToken` would be a more descriptive name here, could you also please change the second argument to an arrow function? Arrow functions are used elsewhere, we should be consistent.

examples/react-apollo/cypress/support/e2e.ts
```diff
@@ -22,10 +40,54 @@ Cypress.Commands.add('signUpEmailPassword', (email, password) => {
...
 Cypress.Commands.add('confirmEmail', (email) => {
   cy.mhGetMailsByRecipient(email)
     .should('have.length', 1)
     .then(([message]) => {
       cy.visit(message.Content.Headers['X-Link'][0])
+      cy.saveRefreshToken()
     })
 })
+
+Cypress.Commands.add('quickSignUp', (givenEmail) => {
```
szilarddoro, line 76: can we rename this to `signUpAndConfirmEmail`, so that it becomes a bit more descriptive?

examples/react-apollo/cypress/support/e2e.ts
plmercereau: Changed the name, but did not put an arrow function as it requires `this` to get the access token

examples/react-apollo/cypress/e2e/4-authenticated/sign-out.cy.ts
plmercereau: remnants of the following reply

examples/react-apollo/cypress/e2e/3-sign-in/email-password.cy.ts
```diff
@@ -0,0 +1,24 @@
...
+  const email = faker.internet.email()
+  const password = faker.internet.password()
+
+  it('should sign-up', () => {
+    cy.signUpEmailPassword(email, password)
+    cy.contains('Verification email sent').should('be.visible')
+    cy.confirmEmail(email)
+  })
+
+  it('should sign-in with email and password', () => {
```
szilarddoro, line 13: Tests shouldn't depend on each other. Skipping the first one would break this test.

examples/react-apollo/cypress/support/e2e.ts
```diff
@@ -22,10 +42,58 @@ Cypress.Commands.add('signUpEmailPassword', (email, password) => {
   cy.contains('Continue with email + password').click()
 })
 
+Cypress.Commands.add('signUpEmailPasswordless', (email) => {
+  cy.visit('/sign-up')
+  cy.contains('Continue with passwordless email').click()
```
szilarddoro, line 47: `cypress-testing-library` exposes a helper, which may be useful here: `findByRole`.

Instead of `cy.contains('...')` we could use

```js
cy.findByRole('button', { name: /continue with passwordless email/i })
```

We can make the code more declarative here and also utilize cypress-testing-library even better.

For more information, see: https://testing-library.com/docs/cypress-testing-library/intro/#examples

examples/react-apollo/cypress/e2e/3-sign-in/token.cy.ts
```diff
@@ -0,0 +1,17 @@
+context('Sign in with a refresh token', () => {
+  it('should sign-up', () => {
+    cy.signUpAndConfirmEmail()
+  })
+
+  it('should sign-in with a refresh token', () => {
```
szilarddoro, line 6: This test depends on the previous one.

szilarddoro (changes requested):

examples/react-apollo/cypress/e2e/3-sign-in/token.cy.ts
```diff
@@ -0,0 +1,22 @@
+context('Sign in with a refresh token', () => {
+  it('should sign-in with a refresh token', () => {
+    cy.signUpAndConfirmEmail()
+    cy.contains('Profile page')
+    cy.clearLocalStorage().then(() => {})
```
szilarddoro, line 5: Calling `then` seems unnecessary here.

examples/react-apollo/cypress/e2e/4-authenticated/token-refresh.cy.ts
```diff
@@ -0,0 +1,16 @@
+import faker from '@faker-js/faker'
+
+context('Token refresh', () => {
```
szilarddoro, line 3: Could you please add tests for the case where the current token expires? (e.g: it's still in local storage, but the user has not visited the page for 4 weeks)

szilarddoro:
Please add this ESLint plugin to the project: https://github.com/cypress-io/eslint-plugin-cypress - it helps avoid common mistakes and in general helps improve the quality of our tests

plmercereau:
> Please add this ESLint plugin to the project: https://github.com/cypress-io/eslint-plugin-cypress - it helps avoid common mistakes and in general helps improve the quality of our tests

I just did. I however feel we could improve our eslint config, ideally to merge all configs into one. I couldn't find a way to make it work for both Vue and React as some of the rules conflict with each other. If you have a good idea, that's worth a PR

szilarddoro:
> > Please add this ESLint plugin to the project: https://github.com/cypress-io/eslint-plugin-cypress - it helps avoid common mistakes and in general helps improve the quality of our tests
> 
> I just did. I however feel we could improve our eslint config, ideally to merge all configs into one. I couldn't find a way to make it work for both Vue and React as some of the rules conflict with each other. If you have a good idea, that's worth a PR

Unfortunately I don't think we can avoid having separate ESLint configurations for different types of projects. I'll take a look at them in a cooldown period to see what we can improve.

szilarddoro:
Are there any extra tests you'd like to add to this PR @plmercereau?

plmercereau:
> Are there any extra tests you'd like to add to this PR @plmercereau?

yes, for instance email/password change, MFA, anonymous sign-in, but I want to add them in separate PRs

szilarddoro (approved):

PR #703, opened 2022-06-10
Use internal `*` dependencies in examples

plmercereau:
- there is no need to use the syntax `workspace:*`: as this syntax is only required to be replaced by the current versions inside the workspace when publishing to npm - and examples are not published
- dependencies across the workspace should however be flagged as `*` so it resolves packages locally. It otherwise resolves on npm (see corrected lockfile)
- as a result, it is then possible to copy-paste an example outside of the workspace and run `yarn` or `npm i`


plmercereau:
cc @mastapegs

szilarddoro (approved):

PR #710, opened 2022-06-14
Info about allowed roles

elitan:


guicurcio (approved):

docs/docs/platform/authentication/users.mdx
```diff
@@ -60,11 +60,29 @@ The default role is used when no role is specified in the GraphQL request. By de
...
 ### Allowed Roles
 
+Allowed roles are roles the user is allowed to use when making a GraphQL request. Usually you would change the role from `user` (the default role) to some other role because you want Hasura to use a different role to resolve permissions for a particular GraphQL request.
+
 By default, users have two allowed roles:
 
 - `user`
 - `me`
 
+You can manage what allowed roles users should get when they sign up under **Users** -> **Roles & Permissions**.
```
plmercereau, line 70: Maybe it's worth mentioning it is also required to add the given role to the `auth.roles` table

elitan: I think we should add new roles automatically to `auth.roles` from our Nhost Console.

We should not delete roles from the `auth.roles` when we remove them from **Allowed Roles**.

Anyways, I'll add the information about this in the mean time.

PR #714, opened 2022-06-15
test: anonymous sign-in and deanonymisation

plmercereau:


examples/react-apollo/cypress/e2e/4-authenticated/apollo.cy.ts
```diff
@@ -1,18 +1,24 @@
 import faker from '@faker-js/faker'
 
 context('Apollo', () => {
```
szilarddoro, line 3: Could you please also add test cases for unhappy paths?

examples/react-apollo/cypress/e2e/2-sign-up/anonymous.cy.ts
```diff
@@ -0,0 +1,54 @@
...
+  })
+
+  it('should sign-up anonymously', () => {
+    cy.contains('You are authenticated')
+    cy.fetchUserData().its('isAnonymous').should('equal', true)
+    cy.contains('You signed in anonymously')
+  })
+
+  it('should deanonymise with email+password', () => {
+    cy.fetchUserData()
```
szilarddoro, line 15: Isn't there a better way to get user information on the UI? This looks like a huge implementation detail that is being used here. We should use something here that would mirror how a user would get this information from the backend. (e.g: clicking some button or something like that)

szilarddoro: We should also not test the actual user object, but the UI instead whether it changed correctly thanks to the deanonymization.

szilarddoro: Okay, it makes sense then. Can we convert this to async / await btw? That would make this code more readable IMO

examples/react-apollo/cypress/support/e2e.ts
```diff
@@ -100,3 +113,26 @@ Cypress.Commands.add('disconnectBackend', () => {
 Cypress.Commands.add('getNavBar', () => {
   cy.get('.mantine-Navbar-root')
```
szilarddoro, line 114: Sorry, I didn't notice this in the previous PR.

Could you change this to `cy.findByRole('banner')` / `cy.findByRole('navigation')`? Provided that Mantine uses the proper semantic element for rendering the navigation bar.

There is [Testing Playground](https://chrome.google.com/webstore/detail/testing-playground/hejbmebodbijjdhflfknehhcgaklhano), which is a great tool for getting the most accessible queries.

examples/react-apollo/cypress/e2e/2-sign-up/anonymous.cy.ts
plmercereau: What do you mean?
https://docs.cypress.io/faq/questions/using-cypress-faq#Can-I-use-the-new-ES7-async-await-syntax
https://github.com/cypress-io/eslint-plugin-cypress/blob/master/docs/rules/no-async-tests.md

examples/react-apollo/cypress/support/e2e.ts
plmercereau: there are multiple `<nav>` elements (and none using the `banner` role), so I don't see other information than class to get the right one

szilarddoro: Then it'll be `navigation` and if you add an `aria-label` to the component describing what it is (e.g: `Main navigation`), then you'll be able to query it by using the following function call:

```ts
cy.findByRole(`navigation`, { name: /main navigation/i })
```

Relying on classes, especially of 3rd party UI libs can be dangerous if they change their implementation at some point in the future. It's a better practice to use accessible queries.

examples/react-apollo/cypress/e2e/2-sign-up/anonymous.cy.ts
szilarddoro: oh, sure

examples/react-apollo/cypress/support/e2e.ts
```diff
@@ -1,8 +1,8 @@
 import { faker } from '@faker-js/faker'
+import { User } from '@nhost/core'
```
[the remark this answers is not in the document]
plmercereau: It is a dependency of @nhost/react. Unless our monorepo setup doesn't allow it I think it's best not to add unecessary deps in examples to keep them as simple as they can

szilarddoro (approved):
Looks good! 👍🏼 

Two minor things:
- Build commands (`pnpm build` and `pnpm build:docs`) don't need the extra `--` to pass their arguments to `pnpm build:all`
- Initially tests were failing locally because I forgot to clean up things locally (most notably the `.nhost` folder). Could you please create a cleanup phase running before `e2e`?

plmercereau:
> Looks good! 👍🏼
> 
> Two minor things:
> 
> * Build commands (`pnpm build` and `pnpm build:docs`) don't need the extra `--` to pass their arguments to `pnpm build:all`

Updated the build scripts.

> * Initially tests were failing locally because I forgot to clean up things locally (most notably the `.nhost` folder). Could you please create a cleanup phase running before `e2e`?

I'm not 100% certain it is related to a problem with .nhost though. it feels more like something related to dist files not correctly updated after switching branches. Will keep it as-is until we can reproduce and identify the source of the problem - to remove `.nhost` is a bit of a pain as it is sometimes useful to keep Nhost data while developing.

PR #735, opened 2022-06-21
Add "redirectTo" in Provider SignIn interface

nachoaldamav:
The provider SignIn interface doesn't have "redirectTo" param.

### DIFF
--- a/packages/hasura-auth-js/src/utils/types.ts
+++ b/packages/hasura-auth-js/src/utils/types.ts
@@ -78,6 +78,7 @@ export interface SignInPasswordlessSmsOtpParams {
 export interface SignInWithProviderOptions {
   provider: Provider
   options?: ProviderOptions
+  redirectTo?: string
 }
 
 export type SignInParams =


packages/hasura-auth-js/src/utils/types.ts
```diff
@@ -78,6 +78,7 @@ export interface SignInPasswordlessSmsOtpParams {
 export interface SignInWithProviderOptions {
   provider: Provider
   options?: ProviderOptions
+  redirectTo?: string
```
plmercereau, line 81: Thanks for this.
However `redirectTo` is already included in `ProviderOptions`

nachoaldamav: Typescript shows this error for me:
```You cannot assign an argument of type "{ provider: "google"; redirectTo: string; }" to the parameter of type "SignInParams". The object literal can only specify known properties and 'redirectTo' does not exist in the 'SignInParams' type.```

Comment (plmercereau) [reply packages/hasura-auth-js/src/utils/types.ts:81]: try
```ts
nhost.auth.signIn({
  provider: 'google',
  options: { redirectTo: `${host}${pathname}` }
})
```

nachoaldamav: When i click in the login button it redirects to the production domain, but I'm working in localhost.

plmercereau: I would
1. check the `backendUrl` value used when creating the Nhost client
2. log the value of `${host}${pathname}`  
3. check the browser network logs to see which server is called
4. check the value the Google Oauth callback url - must be `http://localhost:1337/v1/auth/signin/provider/google/callback`

In any case, this discussion is not part of a pull request, and I'm therefore closing it. If the above suggestions don't end up solving your problem, don't hesitate to open an issue or to start a discussion, either on GitHub or Discord.

Good luck!

PR #737, opened 2022-06-22
Added ROSS Index badge

elitan:


plmercereau (approved):
Nice. Do we know where we'll be on the Q2 list?

elitan:
> Nice. Do we know where we'll be on the Q2 list?

We'll see once it comes out.

PR #768, opened 2022-06-28
Storage fixes and improvements

### LINKED ISSUE #767: Error when uploading multiple files
The `@nhost/react` SDK throws the following error when trying to upload multiple files:

```
index.esm.js?e606:263 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'error')
    at eval (index.esm.js?e606:263:1)
```
which points to
```
message: ((_b = response == null ? void 0 : response.data.error) == null ? void 0 : _b.message) || message
```

I simplified the relevant code, but 
```tsx
const { upload, add, clear } = useMultipleFilesUpload();

async function handleFileUpload(event: ChangeEvent<HTMLInputElement>) {
  const {files} = event.target;
  const defaultBucket = { id: 'default' } // <- comes from the API normally, not relevant here

  // some validation ...

  try {
    setLoading(true)

    // Suggestion #1: Maybe this is something that we could improve as file input fields return FileLists instead of arrays containing a list of Files
    // ---
    // Suggestion #2: I'd say it's not always straightforward that users want to upload files this way.
    // I just want to upload it immediately after the `onChange` handler is fired by simply calling `upload` with the `FileList`.
    add(Array.from(files))

    // Suggestion #3:
    await upload({ bucketId: defaultBucket.id })

  } catch (error) {
    // showing some error to the users
  } finally {
    setLoading(false)
    event.target.value = null
    clear()
  }
}
```

I can even make my browser freeze from time to time after selecting 3-4 files. I also added a couple of suggestion as comments to the code. Perhaps we can improve the API by applying these changes.

plmercereau:
close #767 

missing:
- [x] changesets
- [x] adapt inline examples
- [x] pass the tests


packages/react/src/useFileUpload.ts
```diff
@@ -5,22 +5,24 @@ import {
...
   FileUploadMachine,
   FileUploadState,
+  StorageUploadParams,
   UploadFileHandlerResult,
   uploadFilePromise
 } from '@nhost/hasura-storage-js'
 import { useInterpret, useSelector } from '@xstate/react'
 
 import { useNhostClient } from './useNhostClient'
+
```
szilarddoro, line 15: I can't leave comments on unchanged lines, but could you please fix the formatting of the `useFileUpload` example in this file? It gets formatted properly for auto-generated docs, but not for IDEs.

plmercereau: I don't get the problem. Could you share a snapshot?

szilarddoro: Screenshot 2022-06-29 at 14 00 21

it should be displayed like this:

```ts
const {
  add,
  upload,
  cancel,
  // rest...
} = useFileUpload();
```

PR #783, opened 2022-07-02
Docs: Use subdomain and region instead of backend url

elitan:

PR #786, opened 2022-07-04
Use :1337 if no port is specified when using "localhost"

nunopato:
- when using "localhost" without specifying a port, default to :1337


plmercereau (changes requested):
Please add a changeset

plmercereau:
subdomain: 'localhost:1234' is somehow semantically incorrect. We can live with it, but I think we should introduce a `port` param at a later stage, that would default to `1337` when `subdomain` equals `'localhost'`

elitan:
Update potential docs that currently use `localhost:1337` to `localhost`. Either after merging or in this PR.

https://github.com/nhost/nhost/pull/783#pullrequestreview-1027664862

elitan:
> subdomain: 'localhost:1234' is somehow semantically incorrect. We can live with it, but I think we should introduce a `port` param at a later stage, that would default to `1337` when `subdomain` equals `'localhost'`

This should work because there are developers who are using `subdomain: 'localhost:1337'` with the current version and things should not break for them just because of this update.

packages/nhost-js/src/utils/helpers.ts
```diff
@@ -19,7 +19,9 @@ export function urlFromParams(
   }
 
   if (backendOrSubdomain.subdomain !== undefined && backendOrSubdomain.subdomain.match(LOCALHOST)) {
-    return `http://${backendOrSubdomain.subdomain}/v1/${service}`
+    return backendOrSubdomain === 'localhost'
+      ? `http://localhost:1337/v1/${service}`
+      : `http://${backendOrSubdomain.subdomain}/v1/${service}`
```
elitan, line 24: For developers who are currently using `localhost:1337` this update will break their code.

I think we should allow for `localhost:<port>` here. Something like this:


```js
return backendOrSubdomain === 'localhost'
      ? `http://localhost:1337/v1/${service}`
      : backendOrSubdomain.startsWith('localhost:') (or regex) 
        ? `http://${backendOrSubdomain}/v1/${service}` 
        : `http://${backendOrSubdomain.subdomain}/v1/${service}`
```

elitan (changes requested):

plmercereau:
`examples/react-apollo/cypress/videos/1-unauthenticated/authentication-guards.cy.ts.mp4` etc should not be pushed

elitan:
> `examples/react-apollo/cypress/videos/1-unauthenticated/authentication-guards.cy.ts.mp4` etc should not be pushed

Should be fixed here: https://github.com/nhost/nhost/pull/786/commits/9feea987e68a3be00c8bc860d1e16fff7193bc59

packages/nhost-js/src/utils/helpers.ts
```diff
@@ -14,17 +14,38 @@ export function urlFromParams(
...
-  if (backendOrSubdomain.subdomain !== undefined && backendOrSubdomain.subdomain.match(LOCALHOST)) {
-    return `http://${backendOrSubdomain.subdomain}/v1/${service}`
+  // to make TS happy
+  if (!subdomain) {
+    throw new Error('`subdomain` must be set if `backendUrl` is not set.')
+  }
+
+  // check if subdomain is localhost[:port]
+  const subdomainLocalhostFound = subdomain.match(LOCALHOST_REGEX)
+  if (subdomainLocalhostFound && subdomainLocalhostFound.length > 0) {
```
plmercereau, line 34: could be simplified by
```js
if (LOCALHOST_REGEX.test(subdomain)
```

plmercereau (approved):

elitan (approved):

PR #787, opened 2022-07-04
JSON schema for `nhost/config.yaml` files

plmercereau:
Rationale:
- the `nhost/config.yaml` file generated by the CLI is not correct: some options changed over time, some have been removed.
- users modify this file to configure Nhost services, but have no guidance whether their configuration is valid and behave as expected
- The JSON schema aims to be published to the [JSON Schema store](https://www.schemastore.org/json/) so syntax highlighting is automatically available in most IDEs

Implementation
- The JSON schema validates every `nhost/config.yaml` present in the monorepo through the `yaml` VS Code extension
- The JSON schema aims to follow the [JSON Schema Store recommendations](https://github.com/SchemaStore/schemastore/blob/master/CONTRIBUTING.md)
  - Use draft v4 syntax
  - Test the schema
  - Use `title`, `description` and `$comment` as much as possible

Limitation
  - The schema is not versioned as our `config.yaml` version property is Hasura config's version, not our config's version.

Next:
- [ ] add more `title` and `description` properties
- [ ] extend the schema, for instance hasura environment variables


.gitignore
```diff
@@ -12,7 +12,6 @@ logs/
 .idea/
 .npm/
 .vagrant/
-/.vscode/
```
dbarrosop, line 15: removing this may cause trouble as people will commit their settings, won't it?

packages/config-schema/tests/index.test.ts
```diff
@@ -0,0 +1,33 @@
...
+
+import draft6MetaSchema from 'ajv/dist/refs/json-schema-draft-06.json'
+
+const ajv = new Ajv()
+ajv.addMetaSchema(draft6MetaSchema)
+
+const schema = JSON.parse(fs.readFileSync('schema.json', 'utf8'))
+
+describe('test valid schemas', () => {
+  const basePath = path.join(__dirname, 'valid')
```
dbarrosop, line 14: I'd advise against "on-the-fly generated tests" as it might lead to unexpected issues if the generation fails silently (i.e. tests are moved and then we are not testing anything at all). It might be better if you added the expected tests over an array and iterated over them instead of generating them dynamically.

.gitignore
plmercereau: yes, I'll ignore the file back, and with your suggestion to run the tests on  `**/nhost/config.yaml` in the repo, we'll be sure everything is configured correctly

examples/codegen-react-apollo/nhost/config.yaml
```diff
@@ -15,100 +16,100 @@ auth:
...
   locale:
-    allowed: en
+    allowed_locales: en
     default: en
   password:
-    hibp_enabled: false
+    hibp:
+      enabled: false
     min_length: 3
   provider:
```
L3K0V, line 41: Missing providers: 

- AzureAD
- WorkOS

packages/config-schema/schema.json
```diff
@@ -0,0 +1,335 @@
...
+        },
+        "port": {
+          "type": "integer",
+          "default": 4000
+        },
+        "server_url": {
+          "type": "string",
+          "deprecated": true
+        },
+        "mfa": {
```
L3K0V, line 310: I think Webauthn is missing

packages/config-schema/schema.json
```diff
@@ -0,0 +1,335 @@
...
+            "twilio": {
+              "type": "object",
+              "additionalProperties": false,
+              "properties": {
+                "enabled": { "type": "boolean" },
+                "account_sid": { "type": "string" },
+                "auth_token": { "type": "string" },
+                "messaging_service_id": { "type": "string" }
+              }
+            },
```
L3K0V, line 216: I think Twilio here is a leftover

```suggestion
```

plmercereau: yes, thanks for that Asen. This PR needs some refactoring - we will probably move to `zod` to get typings and transformers, in addition to a generated JSON schema

PR #788, opened 2022-07-04
fix: return promise when using SMS OTP and an error occurs

plmercereau:


packages/core/src/promises/signInSmsPasswordlessOtp.ts
```diff
@@ -33,7 +33,7 @@ export const signInSmsPasswordlessOtpPromise = (
           user: state.context.user,
           accessToken: state.context.accessToken.value
         })
-      } else if (state.matches({ authentication: { signedOut: 'failed' } })) {
```
franzwilhelm, line 36: So my guess was correct here? :D

plmercereau: yup! Thanks for this!

szilarddoro (approved):
Awesome 🎉 

Thanks @franzwilhelm for your investigation!

PR #791, opened 2022-07-04
feat(client): add typed document node support

notrab:


packages/nhost-js/tests/graphql.test.ts
```diff
@@ -125,4 +128,55 @@ describe('main tests', () => {
     expect(error).toBeTruthy()
     expect(data).toBeNull()
   })
+
+  it('GraphQL client works with TypedDocumentNode', async () => {
```
elitan, line 132: If we expect to get an error for this response, should we have a different title for the test?

packages/nhost-js/src/clients/graphql.ts
```diff
@@ -53,9 +55,10 @@ export class NhostGraphqlClient {
    * @docs https://docs.nhost.io/reference/javascript/nhost-js/graphql/request
    */
   async request<T = any, V = any>(
-    document: string | DocumentNode,
-    variables?: V,
-    config?: AxiosRequestConfig
+    document: string | DocumentNode | TypedDocumentNode<T, V>,
+    ...[variables, config]: V extends never
+      ? [variables?: never, config?: AxiosRequestConfig]
+      : [variables: V, config?: AxiosRequestConfig]
```
elitan, line 61: I have a hard time understanding this code. Can we increase the code readability? Or make a comment explaining what it does?

notrab: Since this function isn't used named args, and variables was previously optional we need to type it so variables is required when it is passed.

I will add a comment.

elitan:
I'm closing this because we plan to use https://github.com/prisma-labs/graphql-request as our graphql client, which will support TypedDocumentNode out of the box.

PR #819, opened 2022-07-13
Docs/vue example improvements

plmercereau:


szilarddoro:
Clicking the verification email redirected me to the example, but I had to refresh the page for the sidebar to change and the page to render some content.

szilarddoro:
Could you please change the forms so that they can be submitted by pressing `Enter`? Maybe it's just the type of the button in the form that's wrong (it should be `submit`) or it's not included in the form's boundaries.

plmercereau:
> 

could you check if you're still facing the issue? If you do, could you share a video to understand what happens?

szilarddoro:
@plmercereau - It's working now, I was redirected to the home page. One problem that I noticed is that there is no progress indicator when sign up is in progress. It was a bit confusing because I pressed `Enter`, the network request started but I had no visual feedback of this. Could you also please change the button's state to disabled while network requests are pending?

plmercereau:
> @plmercereau - It's working now, I was redirected to the home page. One problem that I noticed is that there is no progress indicator when sign up is in progress. It was a bit confusing because I pressed `Enter`, the network request started but I had no visual feedback of this. Could you also please change the button's state to disabled while network requests are pending?

done - although it seems vuetify is not rendering the "loading" button correctly

szilarddoro (approved):
This is a huge improvement already, thanks.

PR #823, opened 2022-07-15
patch: Using same cookie package and made sure cookies are not becoming session cookies

elitan:
- Using same cookie package ([`js-cookie`](https://www.npmjs.com/package/js-cookie)) for both `@nhost/nextjs` and `@nhost/core` packages.
- Adding `expires` to avoid the cookie turning into a session cookie that some browsers aggressively destroy.


plmercereau (changes requested):
Could you please add changesets?

plmercereau (approved):

PR #893, opened 2022-08-23
Make @nhost/storage-js work on server

elitan:


elitan:
Not sure why `nhost`, the CLI, isn't available here: https://github.com/nhost/nhost/runs/7972759971?check_suite_focus=true#step:5:90 🤔

elitan:
https://github.com/nhost/nhost/runs/7990786445?check_suite_focus=true#step:5:133

I suspect this is because of the tsconfig.ts issue. the functions container is never healthy.

It works on hasura-auth-js because the tests start as soon as auth is healthy (while functions are not).

plmercereau (changes requested):
could you please
- add `functions/package.json` everywhere the nhost cli is used
- change `:1337/v1/auth/healthz` to `:9695` in other e2e scripts?

For the moment the other e2e tests are not failing because they are cached by turborepo

plmercereau (approved):

packages/hasura-storage-js/src/hasura-storage-client.ts
```diff
@@ -33,29 +35,58 @@ export class HasuraStorageClient {
...
+   * fd.append('file', fs.createReadStream('./tests/assets/sample.pdf'))
+   * 
+   * await storage.upload({
+   *   formData: fd
+   * })
+   * ```
+   * 
    * @docs https://docs.nhost.io/reference/javascript/storage/upload
    */
   async upload(params: StorageUploadParams): Promise<StorageUploadResponse> {
```
szilarddoro, line 75: I think you should create two [overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads) for `upload`:

```ts
/**
 * ... first documentation ...
 */
upload(params: StorageUploadFileParams): Promise<StorageUploadResponse>;

/**
 * ... second documentation ...
 */
upload(params: StorageUploadFormDataParams): Promise<StorageUploadResponse>;

async upload(params: StorageUploadFileParams | StorageUploadFormDataParams): Promise<StorageUploadResponse> {
  // you'll still need to use a type check similar to what you are doing in the function body
}
```

This would enhance the IDE experience (note the 1/2 label in the tooltip) and would also let us create separate documentation for these functions. 
Screenshot 2022-08-24 at 16 09 51

Actually I think we could further improve this API by changing the function signature to this:
```ts
/**
 * ... first documentation ...
 */
upload(file: File, params: StorageUploadParams): Promise<StorageUploadResponse>;

/**
 * ... second documentation ...
 */
upload(formData: FormData, params: StorageUploadParams): Promise<StorageUploadResponse>;

async upload(fileOrFormData: File | FormData, params: StorageUploadParams) {
  // ...
}
```

This however requires a bigger change from our users on their end, but it would make the parameter list more logical IMO.

packages/hasura-storage-js/tests/upload.test.ts
```diff
@@ -0,0 +1,110 @@
...
+      bucketId: 'default'
+    })
+
+    expect(error).toBeNull()
+    expect(fileMetadata?.bucketId).toBe('default')
+  })
+
+  it('should upload a file with specific bucket id (test-bucket)', async () => {
+    const fd = new FormData()
+    // fd.append('file', fs.createReadStream('./tests/assets/sample.pdf'));
```
szilarddoro, line 99: please remove this comment

packages/hasura-storage-js/src/hasura-storage-client.ts
```diff
@@ -33,29 +35,58 @@ export class HasuraStorageClient {
...
-    const file = new FormData()
-    file.append('file', params.file)
+    let body
```
szilarddoro, line 76: this body is always a type of `FormData` isn't it?

szilarddoro (changes requested):

packages/hasura-storage-js/src/hasura-storage-client.ts
elitan: Yes.

szilarddoro (approved):

PR #907, opened 2022-08-29
fix: improve typings

plmercereau:
Minor typings improvement and cleanup before implementing new features e.g. WebAuthn.
Service calls in @nhost/core were not strongly typed, ending into possible type errors in xstate events.
The Hasura-auth API response types are now consistent and used in both `@nhost/hasura-auth-js` and `@nhost/core`. They have been put in `@nhost/core` as it is a dependency of `@nhost/hasura-auth-js`.
Some unused types has also been removed.


packages/core/src/types.ts
```diff
@@ -207,3 +208,48 @@ export type ClientStorageType =
   | 'react-native'
   | 'web'
   | 'cookie'
+
+// Hasura-auth API response types
+interface NullableErrorResponse {
+  error: ErrorPayload | null
+}
+
+export type NhostSessionResponse =
```
szilarddoro, line 217: I'd change this type to this:

```ts
export interface NhostSessionResponse {
  /**
   * some documentation...
   */
  session: NhostSession | null
  /**
   * some documentation...
   */
  error: ErrorPayload | null
}
```

packages/core/src/types.ts
```diff
@@ -207,3 +208,48 @@ export type ClientStorageType =
...
+  | { session: null; error: ErrorPayload }
+  | { session: NhostSession | null; error: null }
+
+export type SignUpResponse = NhostSessionResponse
+
+export interface SignInResponse {
+  session: NhostSession | null
```
szilarddoro, line 224: Could you please document these fields?

packages/core/src/types.ts
```diff
@@ -207,3 +208,48 @@ export type ClientStorageType =
...
+  error: ErrorPayload | null
+}
+
+export type NhostSessionResponse =
+  | { session: null; error: ErrorPayload }
+  | { session: NhostSession | null; error: null }
+
+export type SignUpResponse = NhostSessionResponse
+
+export interface SignInResponse {
```
szilarddoro, line 223: It seems that this type extends `NhostSessionResponse`.

packages/core/src/types.ts
plmercereau: we would loose the inference "if there is an error then the session is null"

szilarddoro: Yes, that's true. Then https://github.com/nhost/nhost/pull/907#discussion_r958083789 is also irrelevant.

packages/core/src/types.ts
plmercereau: What do you mean? I don't understand

szilarddoro: This is irrelevant, since we are not changing `NhostSessionResponse. (see my comment here: https://github.com/nhost/nhost/pull/907#discussion_r958093496)

szilarddoro (approved):

Issue #882, opened 2022-08-16
Nhost with Nuxt3 predictableActionArguments 


haexhub:
Hi,

I'm using nhost with nuxt3 and getting following warning:

It is highly recommended to set `predictableActionArguments` to `true` when using `createMachine`. https://xstate.js.org/docs/guides/actions.html

I couldn't find anything about it in the docs.

plmercereau:
Hi, thanks for bringing this up.
Please disregard this warning. We will update our code to fix it soon.

mikhailcodes:
Commenting here as I am receiving the same warning using Nhost with Expo/React Native

elleryfamilia:
Getting the same warning on Next.js

dohomi:
Same warning for my environment, using Expo and NextJS in monorepo setup

ammarfaris:
> Commenting here as I am receiving the same warning using Nhost with Expo/React Native

Yup, experiencing the same

PR #908, opened 2022-08-29
Upgrade tooling and dependencies e.g. pnpm, vitest, xstate

plmercereau:
- upgrade to pnpm 7.9.1 (latest version accepted by Vercel)
- lock pnpm version in `package.json` so we (hopefully) stop re-generating the lock file on every PR 
- Bump xstate and remove the [`predictableActionArguments` warning](https://github.com/nhost/nhost/issues/882)
- Simplify typings in tests
- Use `vi.stubGlobal` to [use Vitest 0.24](https://github.com/nhost/nhost/issues/921)
- Bump other tooling dependencies e.g. vite, typedoc, eslint


elitan (approved):

packages/vue/src/useSignUpEmailPassword.ts
```diff
@@ -1,4 +1,4 @@
-import { computed, ToRefs, unref } from 'vue'
```
szilarddoro, line 1: What do you think about enabling this in VSCode settings on a repository level through `.vscode`? I found it very useful in almost all cases. The only problem you can face by enabling this is that commenting out code blocks and saving the file results in lost imports, but it's very easy to restore them. Generally it helps keeping imports clean.

```json
{
  "editor.codeActionsOnSave": {
    "source.organizeImports": true
  }
}
```

plmercereau: Yes. I also like to use the `recommendations` field.
And we'll add this when using JSON schemas:
```json
{
  "yaml.schemas": {
    "./packages/config-schema/schema.json": "nhost/config.yaml"
  }
}
```

plmercereau: anyway I added the `.vscode` directory

package.json
```diff
@@ -23,7 +23,7 @@
     "clean:all": "pnpm clean && rm -rf ./{{packages,examples}/*,docs}/{.nhost,node_modules} node_modules",
     "clean": "rm -rf ./{{packages,examples}/*,docs}/{dist,umd,.next,.turbo,coverage}",
     "ci:version": "changeset version && pnpm install --frozen-lockfile false",
-    "coverage": "pnpm run test --coverage",
+    "coverage": "pnpm run test -- --coverage",
```
nunopato, line 26: is this change on purpose? :eyes:

PR #910, opened 2022-08-30
Authentication with security keys and the WebAuthn API

plmercereau:
Introduce the SDK's part of security keys with WebAuthn

## Note
A changeset will be added once this PR passed its review

## `@nhost/hasura-auth-js`
```ts
nhost.auth.signIn({ email, securityKey: true })
nhost.auth.addSecurityKey(nickname?: string)
```
- `securityKey: true` makes the distinction with `nhost.auth.signIn({ email })` that authenticates with a magic link.
- No `nhost.auth.removeSecurityKey` or `nhost.auth.listSecurityKeys` have been added, as listing/deleting keys occur through the GraphQL api, and `@nhost/hasura-auth-js` cannot access to the parent `nhost.graphql` client.
To import `@nhost/nhost-js` in `@nhost/hasura-auth-js` would create a circular dependency.
Happy to discuss about the matter.

## `@nhost/react`
```ts
const { 
    signInSecurityKeyEmail,
    accessToken,
    error,
    isError,
    isLoading,
    isSuccess,
    needsEmailVerification,
    user
} = useSignInSecurityKeyEmail()
const { add, remove, list } = useSecurityKeys()
```
## `@nhost/vue`
Will come in another PR once this one is merged

## Tests
No e2e tests have been added because [cypress does not support the WebAuthn API](https://github.com/cypress-io/cypress/issues/6991).
Szilard you may have an opinion about testing the Auth state.

## React-Apollo example
Example has been updated to show case the addition, listing and deletion of security keys, and the sign-in.

## Documentation
- Inline documentation is up to date
- Configuration/guide will come in another PR

## Internals
- Some bundlers e.g. NextJs or Webpack don't support pure ESM out of the box, in this case, `@simplewebauthn/browser`. Until a better workaround is found (or NextJs gets up to speed), we bundle this package to avoid users being forced to add `experimental.esmExternals: 'loose'` to their NextJs config - and to avoid the risk of other unidentified frameworks being incompatible with this release. 
- `CodifiedError` extends `Error`, with an `error` that contains the codified error. It helps catching rejected promises in a consistent way


examples/react-apollo/src/components/SignInPasswordlessForm.tsx
```diff
@@ -0,0 +1,87 @@
...
+        title: 'Error',
+        message: result.error?.message || null
+      })
+    } else {
+      setEmailVerificationToggle(true)
+    }
+  }
+  const signInDevice = async () => {
+    const result = await signInSecurityKeyEmail(email)
+    if (result.needsEmailVerification) {
```
szilarddoro, line 33: how about the following:
```ts
if (result.needsEmailVerification) {
  return
}

if (result.isError) {
  showNotification(...)

  return
}

navigate('/', { replace: true })
```

packages/react/src/useSecurityKeys.ts
```diff
@@ -0,0 +1,151 @@
...
+
+export type SecurityKey = { id: string; nickname?: string }
+
+export interface SecurityKeysHookResult extends ActionErrorState, ActionSuccessState {
+  /** Add a security key to the current user with the WebAuthn API */
+  add: AddSecurityKeyHandler
+  /**  List the security keys of the current user */
+  list: SecurityKey[]
+  /** Remove the given security key from the list of allowed keys for the current user */
+  remove: (id: string) => Promise<RemoveSecurityKeyHandlerResult>
```
szilarddoro, line 36: `add` function has `AddSecurityKeyHandler`. I'd suggest to either remove that and inline the `add` function's type in this interface or create one for this too.

packages/react/src/useSecurityKeys.ts
```diff
@@ -0,0 +1,151 @@
...
+}
+interface RemoveSecurityKeyHandlerResult extends ActionErrorState, ActionSuccessState {}
+interface RemoveSecurityKeyHandler {
+  (
+    /** Unique identifier of the security to remove */
+    id: string
+  ): Promise<RemoveSecurityKeyHandlerResult>
+}
+
+export type SecurityKey = { id: string; nickname?: string }
```
szilarddoro, line 28: We should use interfaces wherever possible and wherever it makes sense to be consistent.

packages/react/src/useSecurityKeys.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  const [isLoading, setIsLoading] = useState(false)
+
+  useEffect(() => {
+    const loadList = async () => {
+      setIsLoading(true)
+      // * Initial list
+      const userId = nhost?.auth.getUser()?.id
+      if (userId) {
+        const query = `
```
szilarddoro, line 85: I have a feeling that we should use `graphql-tag` for defining GraphQL queries / mutations.

packages/core/src/machines/index.ts
```diff
@@ -225,6 +233,29 @@ export const createAuthMachine = ({
                       }
                     }
                   }
+                },
+                securityKeyEmail: {
```
szilarddoro, line 237: Could you please create test cases for the security key related stuff?

packages/react/src/useSecurityKeys.ts
```diff
@@ -0,0 +1,151 @@
...
+ * Use the hook `useSecurityKeys` to list, add and remove the WebAuthn security keys of the user.
+ *
+ * When WebAuthn is enabled, the `add` function will work as expected.
+ *
+ * You have to make sure the current user has the right Hasura permissions on the `auth.user_authenticators` table to list and/or remove their keys:
+ *
+ * - List keys: permissions to select the `id` and `nickname` columns
+ * - Remove keys: permissions to delete rows
+ *
+ * It is recommended add the custom chck `{ user_id: { _eq: "X-Hasura-User-Id" } }` to make sure the user can only list and remove their own keys.
```
szilarddoro, line 53:
```suggestion
 * It is recommended to add the custom check `{ user_id: { _eq: "X-Hasura-User-Id" } }` to make sure the user can only list and remove their own keys.
```

szilarddoro (changes requested):

packages/react/src/useSecurityKeys.ts
plmercereau: for development purpose, yes. but I didn't here for two reasons
- graphql is not a dependency
- with a Graphql tag, it would 1. parse the string 2. stringify the document node to end up with the same result -> unnecessary computation

elitan: Yea, I think it's fine to skip `grpahql-tag`.

What we could start doing is add `#graphql` to the query string as many IDEs, then highlights the code as GraphQL:

CleanShot 2022-09-06 at 21 02 24@2x

packages/react/src/useSecurityKeys.ts
```diff
@@ -0,0 +1,154 @@
...
+ * Use the hook `useSecurityKeys` to list, add and remove the WebAuthn security keys of the user.
+ *
+ * When WebAuthn is enabled, the `add` function will work as expected.
+ *
+ * You have to make sure the current user has correct Hasura permissions on the `auth.user_authenticators` table to list and/or remove their keys:
+ *
+ * - List keys: permissions to select the `id` and `nickname` columns
+ * - Remove keys: permissions to delete rows
+ *
+ * It is recommended to add the custom check `{ user_id: { _eq: "X-Hasura-User-Id" } }` to make sure the user can only list and remove their own keys.
```
elitan, line 56: Is this the first time we use the GraphQL API and rely on permissions to be correct to get data?

As an example, we don't do that for `nhost.auth.getUser()`.

If users have incorrect permissions set for the `auth.user_authenticators` table, then `list` will have incorrect values? Not sure I think that's a good idea.

I think we should either
- Not have `list`, and instead recommend users to use GraphQL to list security keys
- Have a specific endpoint in Hasura Auth for users to get their keys

What do you think?

packages/core/src/machines/index.ts
```diff
@@ -683,6 +714,22 @@ export const createAuthMachine = ({
...
+            { email }
+          )
+          let credential: AuthenticationCredentialJSON
+          try {
+            credential = await startAuthentication(options)
+          } catch (e) {
+            throw new CodifiedError(e as Error)
+          }
+          return postRequest<SignInResponse>('/signin/webauthn/verify', { email, credential })
+        },
```
elitan, line 732: What's the flow if users must verify their email? Is it the following:

- Fill in email in sign up form
- Click submit
- Get prompted to add their security key (e.g. with face id)
- Email is sent to the user
- User clicks on the link in the email to verify their email
- User is signed in?

packages/hasura-auth-js/src/hasura-auth-client.ts
```diff
@@ -347,6 +358,17 @@ export class HasuraAuthClient {
...
+   * Use `nhost.auth.addSecurityKey to add a security key to the user, using the WebAuthn API.
+   * @param nickname optional human-readable nickname for the security key
+   *
+   * @docs https://docs.nhost.io/reference/javascript/auth/add-security-key
+   */
+  async addSecurityKey(nickname?: string): Promise<{ error: ErrorPayload | null }> {
+    const { error } = await addSecurityKeyPromise(this._client, nickname)
+    return { error }
+  }
+
```
elitan, line 371: Is there no `removeSecurityKey` method on `nhost.auth`?

packages/react/src/useSecurityKeys.ts
```diff
@@ -0,0 +1,154 @@
...
+  const isSuccess = !error
+  const isError = !!error
+
+  const [isLoading, setIsLoading] = useState(false)
+
+  useEffect(() => {
+    const loadList = async () => {
+      setIsLoading(true)
+      // * Initial list
+      const userId = nhost?.auth.getUser()?.id
```
elitan, line 86: What happens if the `nhost` instance is not available here? Maybe we should check `nhost` and throw an error if it's not available?

elitan (changes requested):

packages/react/src/useSecurityKeys.ts
plmercereau: - `nhost.auth.getUser()` only decodes the JWT, which is a bit tricky as it's only updated every 15 minutes. For instance, if the user changed their display name through a mutation, `getUser()` will send the wrong value. This is not something we can or should extend elsewhere
- no permission -> no list
- no list: I think it will make DX overcomplicated to tell "you can add a key, but if you want to list them or to remove one, then you will have to write your own query/mutation, etc etc". It feels really wrong to offer a simple solution to add a key, but not to list/remove them, which should be a really basic thing do
- specific endpoint: "no, god, please, no" (Michael Scott). We have a GraphQL endpoint that does the job, with the two authz layers that are configurable through metadata. I believe we should avoid re-creating a side REST API. GraphQL & Hasura is a perk, not a constraint.

packages/core/src/machines/index.ts
plmercereau: - fill in email
- click submit
- email is not verified: does not go further
- up to the frontend to decide what to do

packages/hasura-auth-js/src/hasura-auth-client.ts
plmercereau: as explained in the opening of this PR, there is no list/remove in `nhost.auth` as `nhost.auth` has no access to the GraphQL api - `nhost.auth` is in `@nhost/hasura-auth-js`, where as `nhost.graphql` is in `@nhost/nhost-js`. `nhost-js` depends on `hasura-auth-js` and not the other way around. To do so would create a circular dependency

packages/react/src/useSecurityKeys.ts
plmercereau: if `nhost` is not available, it won't do anything. But then, when it becomes available, the action will execute again as we are in `useEffect`

packages/react/src/useSecurityKeys.ts
elitan: I'm just worried that we're introducing a new pattern of managing data. `list` and `remove` is just an abstractios of GraphQL. They are simply graphql queries and mutations under the hood.

We don't have this GraphQL abstraction pattern anywhere else in our SDK.

I feel we should leave GraphQL operations to the developer and their GraphQL client of choice.

packages/core/src/machines/index.ts
elitan: Wouldn't it be better if the user could add their security keys but not sign in until the email is verified?

Just like email + password. The user signs up using email and password and then needs to verify their email to sign in.

packages/hasura-auth-js/src/hasura-auth-client.ts
elitan: Does that mean we only have support for `add` in `@nhost/nhost-js`?

But we have support for `add`, `list`, and `remove` in `@nhost/react`?

packages/react/src/useSecurityKeys.ts
```diff
@@ -0,0 +1,154 @@
...
+        })
+        if (e) {
+          // non-standard error
+          setError({ error: 'graphq-error', message: JSON.stringify(e), status: OTHER_ERROR_CODE })
+        } else if (data) {
+          setList(data.authUserAuthenticators)
+        }
+        setIsLoading(false)
+      }
+    }
```
elitan, line 110: How about an early exit here, instead of wrapping everything in an if-statement?

```js

// handle exception if `userId` is not set
if (!userId) {
  // do something
}

// everything is good, continue.

File: packages/react/src/useSecurityKeys.ts [hunk 12]
```diff
@@ -0,0 +1,154 @@
...
+        isSuccess: true
+      }
+    } else {
+      return {
+        // non-standard error
+        error: { error: 'graphq-error', message: JSON.stringify(e), status: OTHER_ERROR_CODE },
+        isError: true,
+        isSuccess: false
+      }
+    }
```
elitan, line 150: I would recommend an early exit here as well. First, check errors and other edge cases. Then continue with the execution if all is good.

```js
if (e) {
  // handle error
  return
}

if (!data) {
  // handle if no data is available
  return
}

// All is good, continue with execution
```

packages/react/src/useSecurityKeys.ts
```diff
@@ -0,0 +1,154 @@
...
+          { userId: string }
+        >(query, {
+          userId
+        })
+        if (e) {
+          // non-standard error
+          setError({ error: 'graphq-error', message: JSON.stringify(e), status: OTHER_ERROR_CODE })
+        } else if (data) {
+          setList(data.authUserAuthenticators)
+        }
```
elitan, line 107: What happens if there is no error and no `data`?

Maybe we should do the same checks here:

```js

if (e) {
 // return handle error
}

if (!data) {
  // return handle if no data available
}

// continue with execution, all is good.
```

elitan (changes requested):

packages/core/src/machines/index.ts
plmercereau: Not sure I understood, but the two ways I did, no, it won't be better:
- if by "add security key" you mean try to authenticate with the key, if the email is not verified, but connected still the user would not have been able to add their key to their account - no key is available.
- if you mean "add security key", we don't add it on sign up, and we shouldn't security-wisde: while the email is not verified yet, a malicious person could then add a security key while waiting for the real user to validate their email. Once done, the malicious person would become able to authenticate

packages/react/src/useSecurityKeys.ts
plmercereau: I understand. We decided to set three APIs and only one GraphQL. From the developer perspective, I would expect not to be forced to juggle across APIs to use simple CRUD feature.

packages/hasura-auth-js/src/hasura-auth-client.ts
plmercereau: yes

packages/react/src/useSecurityKeys.ts
plmercereau: I don't think it's possible but I can add this case if you want

packages/react/src/useSecurityKeys.ts
plmercereau: sorry, remnants of the old times early breaks were still a bad pattern

packages/core/src/machines/index.ts
elitan: Sorry, let me rephrase it.

I'm thinking we should treat a security key like a password.

A user signs up with email+password, or email+security key. 

If the user must verify their email before signing in, they do so.

Once the email verification step is completed, the user can sign in with email+password or email+security key.

plmercereau: sign in:
the frontend can decide to either send an alert like "you must verify your email first" or send a new verification email with `nhost.auth.sendVerificationEmail`.
It follows the same patterns as a user trying to authenticate with an unverified email and a password

sign up:
- create an user with email+password, passwordless email, oauth - whatever method that gets an email
- authenticate
- add a sign-in key

We talked about removing the webauthn sign-up endpoint two weeks ago and it was basically the main refactoring of auth v0.11, as the logic was a duplicate to sign-up with email+password and wouldn't allow other registration methods

elitan: We agreed offline to keep things as it is. 

- Users can add a security key once they are signed in.
- When users have added a security key, they can sign in using their email and security key.

Later, we might come back to allow users to sign up with an email + security key just like they would using an email + password.

packages/react/src/useSecurityKeys.ts
elitan: We agreed offline to remove `list` and `remove` as helpers in the React SDK because:

- To list and remove security keys can be done with the GraphQL API already.
- We can't have `list` and `remove` in the `@nhost/nhost-js` for technical reasons, which made the SDKs inconsistent.

packages/hasura-auth-js/src/hasura-auth-client.ts
elitan: See this comment: https://github.com/nhost/nhost/pull/910#discussion_r968458359

packages/react/src/useAddSecurityKey.ts
```diff
@@ -0,0 +1,65 @@
...
+interface AddSecuritKeyHook {
+  (): AddSecuritKeyHookResult
+}
+
+/**
+ * Use the hook `useAddSecurityKey` to add a WebAuthn security key.
+ *
+ * @example
+ * ```tsx
+ * const { add, isLoading, isSuccess, isError, error } = useSecurityKeys()
```
elitan, line 34: "useSecurityKeys" => "useAddSecurityKey"

examples/react-apollo/src/profile/security-keys.tsx
```diff
@@ -1,12 +1,64 @@
...
+    e.preventDefault()
+    const { key, error } = await add(nickname)
+    if (error) {
+      console.log(error)
+    } else {
+      setNickname('')
+    }
+    if (key) {
+      setList([...list, key])
+    }
```
elitan, line 52: Isn't `key` always set if there is no error?

If so, this can be replaced with something like:


```js
  if (error) {
    return console.log(error)
  } 

  setNickname('')
  setList([...list, key])
```

elitan (changes requested):

elitan (approved):

elitan (approved):
🚀

examples/react-apollo/src/profile/security-keys.tsx
szilarddoro: The reason was not just to support the docgen, but because it's generally a better practice to rely on interfaces as long as types are not needed. If type aliases make more sense, then go for type aliases.

This is what the official TS documentation also suggests: https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces

PR #916, opened 2022-09-03
feat: Stripe GraphQL

elitan:
package readme: https://github.com/nhost/nhost/tree/feat/graphql-stripe/packages/stripe-graphql-js


packages/stripe-graphql/README.md
```diff
@@ -0,0 +1,141 @@
...
+  // get allowed stripe customer ids for this customer
+  const allowedStripeCustomerIds = user?.workspaceMembers
+    .filter((wm) => {
+      return typeof wm.workspace.stripeCustomerId === 'string';
+    })
+    .map((wm) => {
+      return wm.workspace.stripeCustomerId as string;
+    });
+
+    return { stripe, allowedStripeCustomerIds }
```
plmercereau, line 130: - `stripe` did not/should not change in this method. And even though it would have changed, its reference wouldn't have. Hence the idea to set `stripe` as a parameter on server creation
- From what I understand, the purpose of this method is to answer to the question: is the current requested authorised to get this information? I would simplify its signature to `boolean`.

packages/stripe-graphql/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "README.md"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "scripts": {
+    "dev": "NODE_ENV=development ts-node-dev dev/server.ts",
+    "build": "tsc",
+    "test:watch": "vitest",
+    "test:coverage": "vitest run --coverage",
```
plmercereau, line 33: I think we definitely need to start writing e2e asap, from a basic test application in the same way we did for `@nhost/nhost-js`, `@nhost/hasura-auth-js` and `@nhost/hasura-storage-js`

plmercereau:
and we'll need to add a changeset
good stuff ;)

packages/stripe-graphql/src/index.ts
```diff
@@ -0,0 +1,142 @@
+import * as dotenv from 'dotenv'
+import { GraphQLScalarType } from 'graphql'
+import Stripe from 'stripe'
+
+import { CORSOptions, createServer, GraphQLYogaError, YogaInitialContext } from '@graphql-yoga/node'
+
+dotenv.config()
```
plmercereau, line 7: not sure we should add `dotenv` to the package - we could let the user decide to use it
or then, to make it global to the `functions` environment?

packages/stripe-graphql/README.md
```diff
@@ -0,0 +1,141 @@
...
+```
+
+## Quick Start
+
+### Serverless Function Setup
+
+Inside a Nhost project, create a file in `functions/graphql/stripe.ts`:
+
+```js
+import Stripe from 'stripe'
```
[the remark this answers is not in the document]
elitan: Yea the question is if `@nhost/stripe-graphql` should manage `stripe` or if users should be required to pass their `stripe` instance?

packages/stripe-graphql/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+```graphql
+query {
+  users {
+    # User in your database
+    id
+    displayName
+    userData {
+      stripeCustomerId # Customer's Stripe Customer Id
+      stripePaymentMethods {
```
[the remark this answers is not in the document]
plmercereau: We could also get some inspiration from the [OneGraph API](https://www.onegraph.com/graphiql) that interfaces Stripe with GraphQL, amongst many others

plmercereau:
I would also add to the readme an explanation about permissions on remote schemas: Hasura handles field permissions, whereas we are here handling "record-based" permissions.

packages/stripe-graphql/README.md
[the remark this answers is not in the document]
elitan: I've updated the example a bit.

Where there is a stripe customer id, there can be a relationship to the `StripeCustomerId` type. The relationship is named as the user wish, but I would recommend `stripeCustomer`.

From the `stripeCustomer` it's possible to get more info from that Stripe customer.

packages/stripe-graphql/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+Connect data in your database with data from Stripe, via GraphQL.
+
+```graphql
+query {
+  users {
+    # User in your database
+    id
+    displayName
+    userData {
```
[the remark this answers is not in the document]
elitan: Since we can't create columns on the `auth.users` table, we can't create a relationship to/from `users` since the remote schema requires a stripe customer id column.

It's only possible to create this stripe customer id column in another table (e.g. `userData`) and create the relationship at that table.

elitan:
Should we use snake_case everywhere in the GraphQL API or transform everything to camelCase?

### Option 1:

Use snake_case everywhere. 

**Pros:** 

- Follows Stripe APIs naming convention.

**Cons:**

- Breaks GraphQL naming convention.
- Consumers of the GraphQL API is usually JavaScript, which use camelCase.

### Option 2: 

Transform everything to camelCase.

**Pros:**

- Follows GrpahQL naming convention.
- Consumers of the GraphQL API is usually JavaScript, which use camelCase.

**Const**:

- Breaks Stripe API's naming convention.


## Note

It's always possible to use aliases in GraphQL so users can get the data in the form they want no matter what we decide.

Example:

If we decide on using camelCase but the user wants the data returned in snake_case naming convention.

```graphql
stripe_customer: stripeCustomer {
  id
  name
  address {
    city
    postal_code: postalCode
  }
}

File: packages/stripe-graphql/README.md [hunk 7]
```diff
@@ -0,0 +1,153 @@
...
+
+<p align="center">
+  <img alt="npm" src="https://img.shields.io/npm/v/@nhost/stripe-graphql">
+  <img alt="npm" src="https://img.shields.io/npm/dm/@nhost/stripe-graphql">
+  <a href="LICENSE">
+    <img src="https://img.shields.io/badge/license-MIT-yellow.svg" alt="license: MIT" />
+  </a>
+</p>
+
+Stripe GraphQL API for Hasura Remote schemas.
```
nunopato, line 12: I think `with` or `using` instead of `for` reads better.

packages/stripe-graphql/README.md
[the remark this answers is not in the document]
nunopato: I think the package should manage `stripe` so users only have to import `@nhost/stripe-graphql`.

nunopato:
I would call the package `@nhost/stripe-js` because we might want to have implementations in other languages and that way we follow the same pattern as the other packages.

plmercereau:
> Should we use snake_case everywhere in the GraphQL API or transform everything to camelCase?
> 
> ### Option 1:
> Use snake_case everywhere.
> 
> **Pros:**
> 
> * Follows Stripe APIs naming convention.
> 
> **Cons:**
> 
> * Breaks GraphQL naming convention.
> * Consumers of the GraphQL API is usually JavaScript, which use camelCase.
> 
> ### Option 2:
> Transform everything to camelCase.
> 
> **Pros:**
> 
> * Follows GrpahQL naming convention.
> * Consumers of the GraphQL API is usually JavaScript, which use camelCase.
> 
> **Const**:
> 
> * Breaks Stripe API's naming convention.
> 
> ## Note
> It's always possible to use aliases in GraphQL so users can get the data in the form they want no matter what we decide.
> 
> Example:
> 
> If we decide on using camelCase but the user wants the data returned in snake_case naming convention.
> 
> ```graphql
> stripe_customer: stripeCustomer {
>   id
>   name
>   address {
>     city
>     postal_code: postalCode
>   }
> }
> ```

Option 2. Users will use this package to get a GraphQL api, not to play with the Stripe API, and we should make our best to respect these conventions. Moreover this convention will spread into JS code, codegen etc.
OneGraph does it and their API is very clean

plmercereau:
> I would call the package `@nhost/stripe-js` because we might want to have implementations in other languages and that way we follow the same pattern as the other packages.

npm packages cannot be anything else than javascript

elitan:
> > I would call the package `@nhost/stripe-js` because we might want to have implementations in other languages and that way we follow the same pattern as the other packages.
> 
> npm packages cannot be anything else than javascript

Yea I think `@nhost/stripe-graphql` is fine as it's the name of the npm package (JS only).

However, we have one package called `@nhost/nhost-js` which I actually think we should consider renaming to `@nhost/client`.

packages/stripe-graphql/README.md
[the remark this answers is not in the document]
elitan: I think that's the best option too, so we control what Stripe npm version and Stripe API version are being used.

packages/stripe-graphql/src/schema/customer.ts
```diff
@@ -0,0 +1,66 @@
...
+        }),
+        endingBefore: t.arg.string({
+          required: false
+        }),
+        limit: t.arg.int({
+          required: false
+        })
+      },
+      nullable: false,
+      resolve: async (customer, { type, startingAfter, endingBefore, limit }, { stripe }) => {
```
hayes, line 38: This could be implemented as a relay connection instead (using the relay plugin):

```ts
t.connection({
  type: 'StripePaymentMethod',
  args: {
    type: t.arg({
      type: PaymentMethodTypes,
      required: true,
      defaultValue: 'card',
    }),
  },
  resolve: (customer, { first, last, before, after, type }) => {
    const paymentMethods = await stripe.customers.listPaymentMethods(customer.id, {
      type,
      starting_after: after || undefined,
      ending_before: before || undefined,
      limit: before ? last : first,
    });

    return {
      pageInfo: {
        hasNextPage: paymentMethods.has_more,
        hasPreviousPage: !!after,
        startCursor: paymentMethods.data[0]?.id,
        endCursor: paymentMethods.data[paymentMethods.data.length - 1]?.id,
      },
      edges: paymentMethods.data.map((paymentMethod) => ({
        cursor: paymentMethod.id,
        node: paymentMethod,
      })),
    };
  },
});
```

packages/stripe-graphql/src/schema/payment-methods.ts
```diff
@@ -0,0 +1,64 @@
...
+//     required: false
+//   }),
+//   limit: t.int({
+//     required: false
+//   })
+// }))
+
+export const PaymentMethodTypes = builder.enumType('PaymentMethodTypes', {
+  values: [
+    'acss_debit',
```
hayes, line 35: It seems like the unofficial standard for graphql enums is to use all upper case values for enums.

This could be implemented as:

```ts
values: {
  ACSS_DEBIT: {  value:  'acss_debit' },
  AFFIRM: { value: 'affirm' }
  ...
} as const
```

I'm not too familiar with stripe, or if these values might be sourced from some other API/source/endpoint and need to match up (in that case, this conversion would not be ideal)

packages/stripe-graphql/src/schema/stripe.ts
```diff
@@ -0,0 +1,39 @@
...
+        return customer
+      }
+    })
+  })
+})
+
+builder.queryFields((t) => ({
+  stripe: t.field({
+    type: 'Stripe',
+    resolve: () => ({} as any)
```
hayes, line 37: looks like you are defining the backing type for the 'Stripe' object in the builder as having a customer property, removing that customer property should allow you to avoid the any here.  The type defined in the builder is used for the expected return type of a resolve, and for the parent argument in resolvers for that type, but does not constrain what fields you can defined (you can still define your customer field on `Stripe` even if the backing typescript type does not have it.

packages/stripe-graphql/src/schema/subscriptions.ts
```diff
@@ -0,0 +1,25 @@
+import { builder } from '../builder'
+
+builder.objectType('StripeSubscriptions', {
```
hayes, line 3: These list-wrapping object types would more commonly be represented as relay connection objects.

hayes (commented):
This looks awesome, cool to see Pothos in here.  Let me know if you run into any issues, or have questions about pothos.

packages/stripe-graphql/src/schema/customer.ts
elitan: Would this mean we would "break" the Stripe's API convention? 

I'm trying to keep the API similar to what Stripe's API returns, except for transforming snake_case to camelCase.

hayes: yes this would be aligning with graphql/relay conventions over stripe conventions.  I don't know if that is a good tradeoff here, and sticking with stripes conventions may be more valuable.  The relay connection pattern can be useful because many clients have support for automatically paginating lists of this shape, which can make it easier to build reusable components/libraries that will work with various data sources.  I don't have strong opinions here, just figured I'd mention it just in case. 

https://relay.dev/graphql/connections.htm

packages/stripe-graphql/src/schema/payment-method.ts
```diff
@@ -0,0 +1,10 @@
+import { builder } from '../builder'
+
+builder.objectType('StripePaymentMethod', {
+  fields: (t) => ({
+    id: t.exposeString('id'),
+    object: t.exposeString('object'),
+    created: t.exposeInt('created'),
+    livemode: t.exposeBoolean('livemode')
```
hayes, line 8: likely would want to expose type as well (`type: t.expose({ type: PaymentMethodTypes })`)

packages/stripe-graphql/src/schema/payment-methods.ts
```diff
@@ -0,0 +1,64 @@
...
+//   }),
+//   endingBefore: t.string({
+//     required: false
+//   }),
+//   limit: t.int({
+//     required: false
+//   })
+// }))
+
+export const PaymentMethodTypes = builder.enumType('PaymentMethodTypes', {
```
hayes, line 33: I would prefix this name with `Stripe` (`PaymentMethodTypes`) to keep things namespaced

packages/stripe-graphql/src/schema/customer.ts
elitan: Thanks. For now, I think we'll stick to Stripe's convention, so we don't add too much abstraction on top of Stripe's API. Might come back to this at a later stage.

packages/stripe-graphql/src/schema/subscriptions.ts
elitan: Skipping for now: https://github.com/nhost/nhost/pull/916#discussion_r964431726

packages/stripe-graphql/src/schema/stripe.ts
elitan: Could you guide me on how to do this with Pothos?

I want the `customer` (and later other fields such as `customers`, `paymentMethods`, `plans` etc.) to be scoped under `stripe` like this:

```graphql
query {
  stripe {
    customer(id: "cus_xxx") {
      id
      name
    }
    plans {
      id
      name
    }
  }
}
```

packages/stripe-graphql/src/schema/payment-method.ts
elitan: Hmm, I can't seem to get enums to work here. Do I need to add the `StripePaymentMethodTypes` to the SchemaBuilder somehow?

CleanShot 2022-09-07 at 09 19 12

elitan: I managed to get it to work like this:

CleanShot 2022-09-07 at 09 53 42

packages/stripe-graphql/src/schema/stripe.ts
hayes: I think you have it basically right, you just need to delete line 12 of builder.ts

Those type mappings are basically telling Pothos what data represents the corresponding graphql object.   In this case you are saying a stripe object is represented by an object with a customer property, and here you are returning an object without that property.   You don't actually need (or want) that customer property as part of the parent objects in your resolvers on the Stripe object, so you can just remove that from the type definition in builder.ts

packages/stripe-graphql/src/schema/payment-method.ts
hayes: You probably shouldn't add the enum to the builder like that.  That's shouldn't work either... Weird.   The issue here is that the type property is not compatible with that enum for some reason.  

Instead what I would do is use t.field with a resolver.  In the resolver you can either cast the type as the correct value (or any), or do some conversation or check to ensure the value is what you want.  Using t.field with a resolver will also give you a better error message, when there is a type compatibility issue t.expose has pretty useless errors, that specific type error is one of the worst parts of Pothos.   It might be something as simple as type being nullable

packages/stripe-graphql/README.md
```diff
@@ -0,0 +1,141 @@
...
+const stripe = new Stripe(process.env.STRIPE_SECRET_KEY!, {
+  apiVersion: '2022-08-01'
+})
+
+const server = createStripeGraphQLServer({
+  context: () => {
+    return { stripe }
+  }
+})
+
```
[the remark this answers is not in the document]
elitan: Stripe is no longer required to be returned via the context.

packages/stripe-graphql/README.md
elitan: - Stripe is no longer required here.
- Not really, `allowedStripeCustomerIds` is an array of allowed stripe customer ids with which users should be allowed to interact.

plmercereau (changes requested):

plmercereau (approved):

PR #922, opened 2022-09-06
Add workos provider

plmercereau:


packages/core/src/types.ts
```diff
@@ -87,7 +87,14 @@ export interface DeanonymizeOptions extends RegistrationOptions {
   email?: string
   password?: string
 }
-export interface ProviderOptions extends RegistrationOptions, RedirectOption {}
+
+export interface CommonProviderOptions extends RegistrationOptions, RedirectOption {}
+export interface WorkOsOptions {
+  connection?: string
+  organization?: string
+  provider?: string
```
elitan, line 95: Is at least one of these properties required to be set?

If yes, is there a way to make typescript understand that?

packages/hasura-auth-js/src/utils/types.ts
```diff
@@ -51,10 +52,16 @@ export interface SignInPasswordlessSmsOtpParams {
...
 }
-export interface SignInWithProviderOptions {
-  provider: Provider
-  options?: ProviderOptions
-}
+
+export type SignInWithProviderOptions = { provider: Provider; options?: CommonProviderOptions } & (
+  | {
+      provider: 'workos'
+      options?: WorkOsOptions
```
elitan, line 59: is `options` really optional? Seems like at least one option is required: https://github.com/nhost/hasura-auth/pull/228/files#diff-2d9f3e8dec1e6d72e36daf98fedb945a296f319781dd4c64c5d06711c19a5e69R17-R18

elitan (changes requested):

packages/core/src/types.ts
plmercereau: no, if none of these is set, `hasura-auth` will fall back to a default value. If no default value in `hasura-auth`, it will fail.
But with these options, it is possible to let the frontend dev decide which connection/org/provider to use

packages/hasura-auth-js/src/utils/types.ts
plmercereau: one is required IF there is no default in `hasura-auth`

szilarddoro: I can't wrap my head around this part of the code.

```ts
// First scenario:
const value1: SignInProviderOptions = {
  provider: 'anything-but-workos',
  options: {
    /* non-WorkOS related configuration, aka CommonProviderOptions */
  }
}

// Second scenario:
const value2: SignInProviderOptions = {
  provider: 'workos',
  options: {
    /* WorkOS related configuration, aka WorkOsOptions /*
  }
}

// Third scenario:
const value3: SignInProviderOptions = {
  provider: 'anything-but-workos',
  // we are not providing options here
}
```

Please correct me if I'm wrong but then the type alias should be:
```ts
export type SignInProviderOptions = 
  | { provider: Exclude<Provider, 'workos'>, options?: CommonProviderOptions }
  | { provider: 'workos', options?: WorkOsOptions }
```

Maybe you want to extend `CommonProviderOptions` when defining `WorkOsOptions` if that's the reason why the first `{provider: Provider, options?: CommonProviderOptions` is there.

plmercereau: the `workos` provider has specific options __in addition to__ common options.
In the current types:
- any provider has common options
- workos has some more options
- all the other providers don't have additional options
In doing so, the presence of common options for every provider is made explicit. 

But I guess this works too:
```ts
export type SignInWithProviderOptions =
  | {
      provider: 'workos'
      options?: CommonProviderOptions & WorkOsOptions
    }
  | {
      provider: Exclude<Provider, 'workos'>
      options?: CommonProviderOptions
    }
```

plmercereau: @szilarddoro I commited your workaround about the documentation but it makes Vercel fail. I suggest I revert it for now and we track this in a separate issue

szilarddoro: Yes, the formatting of `Promise<void>` seems to be failing.

elitan: Do we need to do any adjustments given the conversation above?

szilarddoro: You mean this?

```ts
export type SignInWithProviderOptions =
  | { provider: 'workos', options?: CommonProviderOptions & WorkOsOptions }
  | { provider: Exclude<Provider, 'workos'>, options?: CommonProviderOptions }
```

This is exactly what I suggested, so I think it's okay.

plmercereau: allright so I'll change my code again so anhone can understand it.

elitan (approved):

PR #942, opened 2022-09-20
WebAuthn: signup

plmercereau:
Introduce:
```js
const { session, error } = await nhost.auth.signUp({ email, securityKey: true })

// React hook:
const {
    signUpSecurityKeyEmail,
    accessToken,
    error,
    isError,
    isLoading,
    isSuccess,
    needsEmailVerification,
    user
  } = useSignUpSecurityKeyEmail()

await signUpSecurityKeyEmail(email, options)
```

- [x] inline documentation
- [x] tests
- [x] changesets


packages/core/src/machines/index.ts
```diff
@@ -772,7 +804,31 @@ export const createAuthMachine = ({
...
           }
         },
-
+        signUpSecurityKey: async (_, { email, options }) => {
+          if (!isValidEmail(email)) {
+            return Promise.reject<SignUpResponse>({ error: INVALID_EMAIL_ERROR })
+          }
+          // TODO anonymous users
+          const nickname = options?.nickname
+          if (nickname) delete options.nickname
```
szilarddoro, line 813: Could you please elaborate on why deleting `options.nickname` is necessary here?

packages/core/tests/securityKeySignUp.test.ts
```diff
@@ -0,0 +1,190 @@
...
+const authService = interpret(authMachine)
+
+beforeAll(() => server.listen({ onUnhandledRequest: 'error' }))
+afterAll(() => server.close())
+
+beforeEach(() => {
+  authService.start()
+})
+
+afterEach(() => {
```
szilarddoro, line 60: Please clean up after each test by restoring the `mockAuthenticator` to its original implementation.

packages/react/src/useSignUpSecurityKeyEmail.ts
```diff
@@ -0,0 +1,88 @@
...
+    service,
+    (state) => state.context.user,
+    (a, b) => a?.id === b?.id
+  )
+  const accessToken = useSelector(service, (state) => state.context.accessToken.value)
+
+  return {
+    accessToken,
+    error,
+    isError,
```
szilarddoro, line 81: I was thinking about this for a while now. What do you think about introducing a `status` value besides `isLoading` / `isSuccess` and `isError` at some point in the future? It could hold values any of these values `loading | success | `error`. (See [TanStack Query's code](https://github.com/TanStack/query/blob/900d2c51306a91372ae38d0d6846ae61a165c326/packages/query-core/src/query.ts#L488) where something similar is implemented).

It would make it more convenient to use these hooks for React users in some cases.

For example instead of this:

```ts
function Component() {
  const { isLoading, isSuccess, isError, signUp } = useSignUp()
  
  useEffect(() => {
    if (isLoading) { ... }
    if (isError) { ... }
    if (isSuccess) { ... }
  }, [isLoading, isSuccess, isError, /* Many other things in this dep. array */])

  render <SomeComponent />
}
```

Users could use something like this:
```tsx
function Component() {
  const { status, signUp } = useSignUp()
  
  useEffect(() => {
    if (status === 'loading') { ... }
    if (status === 'error') { ... }
    if (status === 'success') { ... }
  }, [status, /* Many other things in this dep. array */])

  render <SomeComponent />
}
```
Maintaining huge dependency arrays is a mental overhead which we can help by eliminating the need to use three separate values. [This blog](https://kentcdodds.com/blog/stop-using-isloading-booleans) post also talks about this topic. I find it very interesting.

plmercereau: I like the idea and also feel our way of doing things is not ideal

packages/core/src/machines/index.ts
plmercereau: the `'/signup/webauthn` accepts any property in `SignUpOption`, meaning `nickname` is not allowed. I set a lax validation in hasura-auth, meaning unknown properties will be ignored rather than raise an error, but it's better not to send it.
I added an explanation in the code

szilarddoro: Thanks for the explanation. This however raises another question: Why do we even need this to be part of `SignUpSecurityKeyOptions`? I just checked the code and don't see this being used anywhere. It's just simply removed here and that's it. Maybe I'm missing something here.

plmercereau: The option was available in the react hook but not correctly typed in `nhost.auth.signUp`.
I just pushed a fix.
You can now write:
```js
await nhost.auth.signUp({ email: 'bob@example.com', securityKey: true, options: { nickname: 'my laptop' } })
```

packages/react/src/useSignUpSecurityKeyEmail.ts
```diff
@@ -0,0 +1,88 @@
...
+interface SignUpSecurityKeyHook {
+  (options?: SignUpSecurityKeyOptions): SignUpSecurityKeyHookResult
+}
+
+/**
+ * Use the hook `useSignUpSecurityKey` to sign up a user with security key and an email using the WebAuthn API.
+ *
+ * @example
+ * ```tsx
+ * const { signUpSecurityKey, needsEmailVerification, isLoading, isSuccess, isError, error } = useSignUpSecurityKey()
```
elitan, line 30: comment hook name is `useSignUpSecurityKey` but the actual name is `useSignUpSecurityKeyEmail`.

I think it should be `useSignUpEmailSecurityKey`.

elitan: According to the comment docs here it's possible to get `isLoading, isSuccess`, etc from the hook.

It is possible to get these state variables both from the hook and also later from `signUpSecurityKey`?

elitan (changes requested):
There are some name inconsistencies.

I think we should use the following:

```js
const { signUpEmailSecurityKey, needsEmailVerification, isLoading, isSuccess, isError, error } = useSignUpEmailSecurityKey()
```

packages/react/src/useSignUpSecurityKeyEmail.ts
plmercereau: > According to the comment docs here it's possible to get `isLoading, isSuccess`, etc from the hook.
> 
> It is possible to get these state variables both from the hook and also later from `signUpSecurityKey`?

yes, except `isLoading` as `signUpSecurityKey` is a promise that is resolved when loading is done

examples/react-apollo/src/sign-in/security-key.tsx
```diff
@@ -1,35 +1,36 @@
...
   const [email, setEmail] = useState('')
   const navigate = useNavigate()
   const [emailVerificationToggle, setEmailVerificationToggle] = useState(false)
 
-  const signIn = async () => {
-    const result = await signInSecurityKeyEmail(email)
-    if (result.isError) {
+  const signIn = async (e: FormEvent<HTMLFormElement>) => {
+    e.preventDefault()
+    const { isError, isSuccess, needsEmailVerification, error } = await signInSecurityKeyEmail(
```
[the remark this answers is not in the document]
plmercereau: both works, the example only illustrates one way on doing things

examples/react-apollo/src/sign-in/security-key.tsx
```diff
@@ -1,35 +1,36 @@
-import { useState } from 'react'
+import { FormEvent, useState } from 'react'
 import { useNavigate } from 'react-router-dom'
 
-import { Button, Modal, TextInput } from '@mantine/core'
+import { Button, Modal, SimpleGrid, TextInput } from '@mantine/core'
 import { showNotification } from '@mantine/notifications'
 import { useSignInSecurityKeyEmail } from '@nhost/react'
```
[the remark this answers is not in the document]
plmercereau: as you wish - breaking change from previous release

packages/react/src/useSignUpSecurityKeyEmail.ts
elitan: @szilarddoro Feel free to create a GitHub issue for this. 👍

elitan (approved):
🚀

PR #961, opened 2022-09-26
added the w_member_social scope to linkedin

Matangub:
w_member_social is the scope in the oauth2 flow that allows creation, deletion or update of posts.

I could make the husky script to work so I would love if you guys can take this pr, remove the husky and deploy this.

### DIFF
--- a/config/.husky/pre-commit
+++ b/config/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-npx lint-staged --config config/.lintstagedrc.js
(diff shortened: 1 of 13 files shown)


elitan:
If you need this for your app it's controlled using the `AUTH_PROVIDER_LINKEDIN_SCOPE` environment variable in [Hasura Auth](https://github.com/nhost/hasura-auth/blob/main/src/config/providers.ts#L99-L102). If you're self-hosting, that's the variable you need to update.

If you need a different scope on an app running at Nhost we can update the scope manually for you (it's not yet possible to do in the Nhost Dashboard). Feel free to reach out to support@nhost.io for that.

PR #963, opened 2022-09-27
doc (database): UI updates and direct access

elitan:
https://docs-git-docs-database-nhost.vercel.app/platform/database


docs/docs/platform/database/index.mdx
```diff
@@ -121,6 +109,34 @@ DROP TABLE "public"."customers";
   </TabItem>
 </Tabs>
 
+## Postgres Access
+
+It's possible to access your Postgres database directly with your favorite Postgres client.
+
+:::info
+
+For projects older than September 2022, it's not possible to connect directly to the database via a connection string. We're working on a one-click fix this this and expect the fix to be available in October 2022.
```
szilarddoro, line 118:
```suggestion
For projects older than September 2022, it's not possible to connect directly to the database via a connection string. We're working on a one-click fix that can be expected to be available in October 2022.
```

docs/docs/platform/database/index.mdx
```diff
@@ -148,12 +164,14 @@ nhost/seeds/default/002-products.sql
 
 ## Backups
 
-Databases on [Pro and Enterprise plans](https://nhost.io/pricing) are backed up automatically.
+Databases on the [Pro and Enterprise plans](https://nhost.io/pricing) are backed up automatically every day.
```
szilarddoro, line 167: Maybe worth mentioning the exact time of the backup here (GMT).

szilarddoro (changes requested):

szilarddoro:
I can't leave comment on a line that has not been edited, but could you please change the SQL of the table `docs/docs/platform/database/index.mdx` at line 65?

This would be similar to the SQL we prepare for the Database UI:
```sql
CREATE TABLE "public"."customers" (
  "id" bigint PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
  "name" text NOT NULL
);
```

This would make it more compatible with our Database UI and I just checked and Serial is a Postgres specific implementation, however identity is an SQL standard.

docs/docs/platform/database/index.mdx
elitan: We'll skip it for now as it's not 100% deterministic, whereas "every day" is.

dbarrosop: I wouldn't mention more than "daily backups". If you say it is at 02.00 UTC people will start having opinions about it while if you leave it out they will mostly not care much about it.

docs/docs/platform/database/index.mdx
```diff
@@ -9,52 +9,40 @@ import TabItem from '@theme/TabItem'
...
-7. Click **Add Table**.
+### Create Table
 
-When a table is created the table is instantly available through the [GraphQL API](/platform/graphql).
+1. Go to **Database** in the left menu.
+2. Click **New table**.
+3. Enter a **name** for the table.
+4. Add **columns**.
+5. Select a **Primary Key** (usually the `id` column).
+6. (Optional) Select an **identity** column (for integer columns only).
```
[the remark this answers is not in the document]
elitan: Updated this. What do you think?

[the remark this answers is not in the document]
szilarddoro: It's good, thanks 👍🏼

szilarddoro (approved):

Issue #980, opened 2022-10-01
feature: have createStripeGraphQLServer accept async isAllowed function


ejkkan:
Problem:
Right now we can't  pass an async function according to the TS spec.
This way we can't really fetch more info about a user. 

The example from the stripe [package release page](https://nhost.io/blog/open-source-stripe-graphql-api) could be considered a use-case for it

Solution:
Just adding Promise support for the isAllowed type

```ts
export declare type CreateServerProps = {
    cors?: CORSOptions;
    isAllowed?: (stripeCustomerId: string, context: Context) => boolean | Promise<boolean>;
    graphiql?: boolean;
};
```

ejkkan:
If you're interested. I could make an PR if I could get permissions

elitan:
> If you're interested. I could make an PR if I could get permissions

Yes please

elitan:
Fixed in https://github.com/nhost/nhost/pull/985

PR #985, opened 2022-10-03
feat/added async support for isAllowed prop for createStripeGraphQLServer

ejkkan:
fixes issue: https://github.com/nhost/nhost/issues/980

### DIFF
--- a/packages/stripe-graphql-js/src/schema/stripe.ts
+++ b/packages/stripe-graphql-js/src/schema/stripe.ts
@@ -17,7 +17,7 @@ builder.objectType('Stripe', {
       resolve: async (_parent, { id }, context) => {
         const { isAllowed } = context
 
-        if (!isAllowed(id, context)) {
+        if (!await isAllowed(id, context)) {
           throw new GraphQLYogaError('Not allowed')
         }
 
@@ -64,7 +64,7 @@ builder.objectType('Stripe', {
           starting_after: startingAfter || undefined
         })
 
-        const customerData = customers.data.filter((customer) => isAllowed(customer.id, context))
+        const customerData = customers.data.filter(async (customer) => await isAllowed(customer.id, context))
 
         customers.data = customerData
 
@@ -95,7 +95,7 @@ builder.objectType('StripeMutations', {
       resolve: async (_, { customer, configuration, locale, returnUrl }, context) => {
         const { isAllowed } = context
 
-        if (!isAllowed(customer, context)) {
+        if (!await isAllowed(customer, context)) {
           throw new GraphQLYogaError('Not allowed')
         }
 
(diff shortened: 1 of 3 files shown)


plmercereau (changes requested):
Thanks for this!
You also need to add an `await` when calling this function [here](https://github.com/nhost/nhost/blob/66ef8f91a3c3430ec877a2d7cbaf2f64df90338b/packages/stripe-graphql-js/src/schema/stripe.ts#L20), [here](https://github.com/nhost/nhost/blob/66ef8f91a3c3430ec877a2d7cbaf2f64df90338b/packages/stripe-graphql-js/src/schema/stripe.ts#L67), and [there](https://github.com/nhost/nhost/blob/66ef8f91a3c3430ec877a2d7cbaf2f64df90338b/packages/stripe-graphql-js/src/schema/stripe.ts#L98)

ejkkan:
> Thanks for this! You also need to add an `await` when calling this function [here](https://github.com/nhost/nhost/blob/66ef8f91a3c3430ec877a2d7cbaf2f64df90338b/packages/stripe-graphql-js/src/schema/stripe.ts#L20), [here](https://github.com/nhost/nhost/blob/66ef8f91a3c3430ec877a2d7cbaf2f64df90338b/packages/stripe-graphql-js/src/schema/stripe.ts#L67), and [there](https://github.com/nhost/nhost/blob/66ef8f91a3c3430ec877a2d7cbaf2f64df90338b/packages/stripe-graphql-js/src/schema/stripe.ts#L98)

Thanks. Please have another look when you have the time @plmercereau

packages/stripe-graphql-js/src/schema/stripe.ts
```diff
@@ -64,7 +64,7 @@ builder.objectType('Stripe', {
           starting_after: startingAfter || undefined
         })
 
-        const customerData = customers.data.filter((customer) => isAllowed(customer.id, context))
+        const customerData = customers.data.filter(async (customer) => await isAllowed(customer.id, context))
```
plmercereau, line 67: Unfortunately this won't work as `filter` doesn't support async functions :(
We could solve this with a reduced and `Promise.all`, but for the sake of readability, we could do something like this:

```suggestion
        const customerData: Stripe.Customer[] = []
        for (const customer of customers.data) {
          if (await isAllowed(customer.id, context)) {
            customerData.push(customer)
          }
        }
```

plmercereau (changes requested):

packages/stripe-graphql-js/src/schema/stripe.ts
ejkkan: My bad. Thanks for the suggestion.
Lets go for it :)

plmercereau:
Looks good to me! I added a changeset so a new version is released.
@elitan if you don't mind having a look before I merge the PR

plmercereau (approved):

elitan (approved):
LGTM

Issue #930, opened 2022-09-12
Nextjs: Nhost client not properly updating user session


msp-ampcome:
I am using nhost with Nextjs. I am trying to use createServerSideClient on a SSR page. but it is not passing the accesstoken when i am making queries with nhost.graphql.request(). I tried to get the session with nhost.auth.getSession(). it is returning null. getNhostSession() is also returning null.

I followed all the steps in the documentation.
i can retrieve the user data once the page is hydrated with useUserData
but i couldn't retrieve any auth related data on getServerSideProps

altschuler:
Same problem as https://github.com/nhost/nhost/issues/886

msp-ampcome:
@altschuler is there any workaround for this? or do we have to wait till they fix it?

altschuler:
I don't think there's much to do about it currently, but someone from nhost would know better :)

ejkkan:
Also seeing this

PR #986, opened 2022-10-03
fix: read nextjs context cookie in clientStorage

### LINKED ISSUE #886: getSession return null when use inside next.js getServerSideProps
Im try to get user session inside getServerSideProps by using - nhost.auth.getSession() but its return null after refreshing the page or route change.
This issue occurs after updating to @nhost/nextjs@1.6.2 or upper.

altschuler:
The nextjs package creates its own NhostClient with a custom `clientStorage` that reads cookies to check if there's an active user session. However, it doesn't read cookies from the nextjs context, meaning that it'll never find the `nhostRefreshToken` cookie when used inside `getServerSideProps`. 

Fixes #886 and #930.

### DIFF
--- a/.changeset/eleven-turkeys-clean.md
+++ b/.changeset/eleven-turkeys-clean.md
@@ -1,0 +1,5 @@
+---
+'@nhost/nextjs': patch
+---
+
+read nextjs context cookies in `createServerSideClient`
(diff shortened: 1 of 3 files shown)


plmercereau (approved):

plmercereau:
Thanks for this fix!

PR #1012, opened 2022-10-08
Recover documentation consistency

d4g0:
This file has diverged from the respective doc [page](https://docs.nhost.io/platform/quickstarts/vue#configure-nhost-with-vue). 
Maybe the docs should be updated instead to match the current file, in any case, i think will be nice to fix this and avoid confusion & friction to new comers trying the vue integration out.


elitan (approved):
Thanks for this! 🙏

Issue #752, opened 2022-06-26
`useAuthenticationStatus` on unstable network


elitan:
From [Discord](https://canary.discord.com/channels/552499021260914688/615892613081333794/988443302783250453).

If `/v1/auth/token` is not accessible, `isLoading` and `isAuthenticated` (from `useAuthenticationStatus`) are both `false`.

This can cause a problem when protecting routes:

```js
import { useAuthenticationStatus } from '@nhost/react'
import { Navigate, useLocation } from 'react-router-dom'

const ProtectedRoute = ({ children }) => {
  const { isAuthenticated, isLoading } = useAuthenticationStatus()
  const location = useLocation()

  if (isLoading) {
    return (
      <div>Loading</div>
    )
  }

  if (!isAuthenticated) {
    return <Navigate to="/sign-in" state={{ from: location }} replace />
  }

  return children
}

export default ProtectedRoute
```

If you're on a mobile phone, and the connection is not stable, it could be that our js SDK tries to sign in the user by making a request to `/v1/auth/token`. The endpoint is not reachable because of the bad connection from the mobile phone. So the SDK sets both `isLoading` and `isAuthenticated` to `false`, which will cause the user to be redirected to `/sign-in`.

Here's what we could do:

1. `isLoading` should only return `false` once the SDK is 100% sure about the `isAuthenticated` state. That means that `isLoading` should only be set to `false` if the SDK gets an actual response from `/v1/auth/token`. Either a `200 OK` or `401 Unauthorize`.
2. If the SDK can't reach `/v1/auth/token` because the network connection is unstable, the SDK should retry the request every 5 seconds until it successfully gets a response.
3. (not sure about this one): We can also introduce a new variable to `useAuthenticationStatus` indicating what's the reason for `isLoading` to be `false`. The reason will always be because the SDK is waiting for a network request, but it could be useful to inform if the network is stable and the SDK is making the first request, or if the network is unstable and the SDK is making the 10th request.

plmercereau:
We could set a retry logic when the auth service cannot be reached, but with the above suggestion, the loading status will remain indefinitely at `true` when the user is offline.
When offline or with a poor connectivity, it is possible to check if a refresh token is present in the local storage:
- If not present, we know the user is not authenticated -> `isLoading` to `false`,  `isAuthenticated` to `false` (current behaviour)
- if present, we know a user _should_ authenticate eventually, unless something goes wrong (for instance invalidated refresh token)

plmercereau:
[...]

PR #1022, opened 2022-10-11
feat: Retry the token import when starting offline

plmercereau:
On startup, retries to import the token until network is up and the server does not respond a 5xx error: the 5 first times every second, then every 5 second.

In the meantime the authentication status is loading.

`nhost.auth.getAuthenticationStatus()` and the `useAuthenticationStatus` react hoos/vue composable get a new `connectionAttempts` property.


packages/core/src/machines/index.ts
```diff
@@ -137,7 +142,20 @@ export const createAuthMachine = ({
                   actions: ['saveSession', 'reportTokenChanged'],
                   target: 'signedIn'
                 },
-                onError: { actions: ['saveAuthenticationError'], target: 'signedOut' }
+                onError: [
```
szilarddoro, line 145: Could you please write tests for this?

szilarddoro (changes requested):

packages/core/src/machines/index.ts
plmercereau: Yes, as soon as we will have validated the design

elitan:
Sounds good to me.

Just a few questions: 

1. What are the values of `isLoading` and `isAuthenticated` after the 20 tries?
2. What happens if Auth returns e.g. 500 (instead of a network error)?

plmercereau:
> 1. What are the values of `isLoading` and `isAuthenticated` after the 20 tries?

after the attempts, `isLoading: false` and `isAuthenticated: true`

> 2. What happens if Auth returns e.g. 500 (instead of a network error)?

It will not retry and will return `isAuthenticated: false`

elitan:
Understood.

What do you think about trying to authenticate indefinitely? This way, we would never get into an `isLoading: false, isAuthenticated: false` state when the user might have a valid refresh token.

Setting `isAuthenticated: false` when the server returns `500` also feels wrong to me. Again, it could be that the user has a valid refresh token. 

Going back to what I wrote [originally](https://github.com/nhost/nhost/issues/752#issue-1284818107), I think this is how it should work:

1. `isLoading` should only return `false` once the SDK is 100% sure about the `isAuthenticated` state. That means that `isLoading` should only be set to `false` if the SDK gets an actual response from `/v1/auth/token`. Either a `200 OK` or `401 Unauthorize`.
2. If the SDK can't reach `/v1/auth/token` because the network connection is unstable, the SDK should retry the request every 5 seconds until it successfully gets a response.
3. (not sure about this one): We can also introduce a new variable to `useAuthenticationStatus` indicating the reason for `isLoading` to be `false`. The reason will always be because the SDK is waiting for a network request, but it could be helpful to inform whether the network is stable and the SDK is making the first request or if the network is unstable and the SDK is making the 10th request.

plmercereau:
then we could add `connectionAttempts` to let the developer decide what to do after x attempts.
I would keep the first five attempts's delay to 1 second and the following to 5 seconds - 5 seconds is very long to wait in a unreliable internet connection that only failed to get one call

plmercereau:
but still, `nhost.auth.signIn` will never resolve when offline

packages/core/src/machines/index.ts
```diff
@@ -869,11 +890,22 @@ export const createAuthMachine = ({
               })
               return { session, error: null }
             } catch (exception) {
-              error = (exception as { error: ValidationErrorPayload }).error
+              error = (exception as { error: ErrorPayload }).error
```
szilarddoro, line 893: You can return the rejected promise when this error happens

szilarddoro (approved):
Looks good! 👍

PR #1026, opened 2022-10-13
Create TIE_ANIMATION.svg

aditimahabole:
This is a demo of how to do an svg animation


elitan:
What's this exactly? I'm closing for now because it does not look relevant, but I'm happy to re-open if I was wrong.

PR #1038, opened 2022-10-15
Fixing React Apollo CRM

elitan:


examples/react-apollo-crm/nhost/config.yaml
```diff
@@ -1,13 +1,5 @@
 metadata_directory: metadata
 services:
-  hasura:
-    image: hasura/graphql-engine:v2.10.1
-    environment:
-      hasura_graphql_enable_remote_schema_permissions: false
-  auth:
-    image: nhost/hasura-auth:0.13.2
-  storage:
-    image: nhost/hasura-storage:0.2.4
```
plmercereau, line 10: images are pinned by [@nhost/sync-versions](https://github.com/nhost/nhost/tree/main/packages/sync-versions) to be sure we run any e2e test in the monorepo with the versions determined in [nhost-cloud.yaml](https://github.com/nhost/nhost/blob/main/nhost-cloud.yaml), and not the default versions in the cli that is used in our CI pipelines. Didn't husky run the sync with a pre-commit hook>

examples/react-apollo-crm/nhost/metadata/databases/default/tables/tables.yaml
```diff
@@ -2,9 +2,9 @@
 - "!include auth_providers.yaml"
 - "!include auth_refresh_tokens.yaml"
 - "!include auth_roles.yaml"
-- "!include auth_user_authenticators.yaml"
```
plmercereau, line 5: the `auth_user_authenticators.yaml` file should also be removed

examples/react-apollo-crm/pnpm-lock.yaml
```diff
@@ -0,0 +1,4938 @@
+lockfileVersion: 5.4
```
plmercereau, line 1: why was the lockfile generated? it is supposed to use the lock file in the monorepo root. The other examples don't have any lockfile to avoid having to maintain it every time we release new versions or change the underlying dependencies. I would remove this

plmercereau (changes requested):

plmercereau (approved):

PR #1044, opened 2022-10-18
chore: 🤖 remove @nhost/react-auth

plmercereau:
The package is deprecated since more than 7 months.
We should remove it instead of continuing to bump its version


szilarddoro (approved):
🧹

PR #1051, opened 2022-10-19
feat: custom endpoints for each service

nunopato:
```ts
import { createClient } = from '@nhost/nhost-js'

// Nhost Cloud
const cloudClient = createClient({
  subdomain: 'somfewpnfnwenwfewfwe',
  region: 'eu-central-1',
})

// Nhost Local (CLI)
const localClient = createClient({
  subdomain: 'localhost',
})

// Self Hosting
const hostedClient = createClient({
  authUrl: 'auth.mydomain.io',
  storageUrl: 'storage.mydomain.io',
  graphqlUrl: 'graphql.mydomain.io',
  funtionsUrl: 'functions.mydomain.io',
})
```

This PR does 2 things:

1. Adds custom endpoint options to the **NhostClient** constructor: `authUrl`, `storageUrl`, `functionsUrl`, and `graphqlUrl` can now be passed as parameters to define the endpoints for each client. These are useful when self hosting. 
2. Fixes the support for using **NhostClient** in the context of the CLI and node (e.g. functions).

Expanding on 2.

Because the CLI uses `traefik` to forward requests to all services, the clients need to be initialized with the correct endpoints.    
When using `NhostClient` with node and the nhost CLI,  the endpoints for the different services should point to `traefik` (e.g. `http://traefik:1337/v1/auth`) instead of `localhost`. This case is specific to creating and using a client from within the CLI in a node environment. 

To support that case, the following environment variables are being introduced to allow the CLI to pass in the endpoints for the services it manages:
 
    - NHOST_AUTH_URL
    - NHOST_STORAGE_URL
    - NHOST_GRAPHQL_URL
    - NHOST_FUNCTIONS_URL 

When set, their value is used. 

### useful?

We could extend the use of those environment variables and add an additional “helper” function for users that want to use environment variables. 

```tsx
// `createClientFromEnv` reads the urls from the environment
import { createClientFromEnv } from '@nhost/nhost-js'

// NHOST_[AUTH|STORAGE|GRAPHQL|FUNCTIONS]_URL
const envClient = createClientFromEnv()

// VS 

// having to pass explicitly
const envClient = new NhostClient({
	authUrl = process.env.NHOST_AUTH_URL,
	storageUrl = process.env.NHOST_STORAGE_URL,
	graphqlUrl = process.env.NHOST_GRAPHQL_URL,
	functionsUrl = process.env.NHOST_FUNCTIONS_URL,
})
```


packages/nhost-js/src/utils/helpers.ts
```diff
@@ -49,3 +49,67 @@ export function urlFromParams(
...
+  return typeof process !== 'undefined' && process.env
+}
+
+/**
+ *
+ * @returns process.env.AUTH_URL if set, null otherwise
+ */
+export function getAuthUrlFromEnv(): string | null {
+  if (!isBrowser() && environmentIsAvailable()) {
+    return process.env.AUTH_URL ? process.env.AUTH_URL : null
```
elitan, line 75: I think this can be replaced with:

```js
    return process.env.AUTH_URL ?? null
```

packages/nhost-js/src/clients/auth.ts
```diff
@@ -0,0 +1,44 @@
...
+ * @param clientStorageSetter
+ * @param clientStorage
+ * @param clientStorageType
+ * @param autoRefreshToken
+ * @param autoSignIn
+ * @param start
+ * @param urlParams
+ * @returns
+ */
+export function getAuthClient(
```
plmercereau, line 19: this function and the following are not getters as they create a client every time they are called. They should be called `createAuthClient`

nunopato: :+1:

guicurcio:
The same client (and bundle) needs to work for any of the possible environments. Users should not have to change the code when working locally, nor if they want to move to Nhost cloud from self-hosting (or vice versa.) It's not clear to me how this would be accomplished with this PR. Can you clarify @nunopato?

plmercereau:
@nunopato here are my [suggestions](https://github.com/nhost/nhost/pull/1085)

config/vite.lib.config.js
```diff
@@ -31,7 +31,7 @@ export default defineConfig({
     globals: true,
     environment: 'jsdom',
     reporters: 'verbose',
-    include: [`${PWD}/src/**/*.{spec,test}.{ts,tsx}`, `${PWD}/tests/**/*.{spec,test}.{ts,tsx}`],
+    include: [`${PWD}/src/**/*.{spec,test}.{ts,tsx}`, `${PWD}/tests/**/**/*.{spec,test}.{ts,tsx}`],
```
plmercereau, line 34: is it really required?
If not, please remove

i18n/translations.md
```diff
@@ -0,0 +1,3 @@
+# Translations
+
+- [English](https://github.com/nhost/nhost)
```
plmercereau, line 3: maybe put this part in a distinct PR?

nunopato:
@guicurcio Users don't need to change the code between local vs cloud. But they have to if self-hosting because `subdomain` wouldn't exist.

packages/nhost-js/src/clients/nhost.ts
```diff
@@ -1,10 +1,41 @@
...
+    clientStorageType,
+    autoRefreshToken,
+    autoSignIn,
+    adminSecret,
+    devTools,
+    start,
+    ...urlParams
+  })
+}
+
```
plmercereau, line 38: This function is not used anymore except in tests, and they could run against `new NhostClient(params)`

If you however want to keep this, you can simplify:
```suggestion
export const createNhostClient = (params: NhostClientConstructorParams) => new NhostClient(params)
```

plmercereau (approved):

PR #1082, opened 2022-11-01
Example: Seed data and storage

elitan:


examples/seed-data-storage/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "@nhost-examples/seed-data-storage",
```
plmercereau, line 2: We need to tell the package is private so our CD won't try to publish it
```suggestion
  "name": "@nhost-examples/seed-data-storage",
  "private": true,
```

plmercereau (changes requested):

plmercereau (approved):

PR #1083, opened 2022-11-01
fix broken Next.js example

szilarddoro:


elitan (approved):

plmercereau (commented):
Do we need a `examples/nextjs/pnpm-lock.yaml` file?

examples/nextjs/next.config.js
```diff
@@ -1,5 +1,9 @@
 /** @type {import('next').NextConfig} */
-const nextConfig = {}
+const nextConfig = {
+  experimental: {
+    esmExternals: 'loose'
+  }
+}
```
plmercereau, line 6: I pulled your branch, removed this part, and it still works 🤷‍♂️

szilarddoro: That's weird because it tries to use `core/dist/index.cjs.js` which doesn't import but requires `@simplewebauthn/browser` which is a pure ESM package. It doesn't work for me without `esmExternals` set to `loose`. Did you restart the process after changing the settings?

szilarddoro:
> Do we need a `examples/nextjs/pnpm-lock.yaml` file?

Nope, removed it. Thanks.

examples/nextjs/next.config.js
plmercereau: I cleaned everything from `node_modules` to `.next`, reinstalled and built, no error without `esmExternals`.
That is weird indeed

szilarddoro: Okay, I found the issue. It was the difference between the dev and prod Vite configuration. When I was running `pnpm dev` in the root folder, it treated `@simplewebauthn/browser` as an external dependency, because the process was using the generic Vite dev configuration, not the one specified in `packages/core`.

PR #1085, opened 2022-11-02
refactor: custom urls

plmercereau:
Adjustments to @nunopato's [PR](https://github.com/nhost/nhost/pull/1051):
- split between 
  - unit tests `pnpm run test` that can run standalone
  - e2e test `pnpm run e2e` that run against a local Nhost project from the CLI
- simplify `createAuthClient -> createHasuraAuthClient -> new HasuraAuthClient` to `createAuthClient -> new HasuraAuthClient`
- allow and test `http[s]://` in subdomain
- use regex groups to increase readability


packages/nhost-js/tests/nhost.test.ts
```diff
@@ -1,5 +1,5 @@
 import { vi } from 'vitest'
```
nunopato, line 1: @plmercereau is there a reason why you are moving this outside of `clients`?

nunopato (approved):
left a small comment, but lgtm. I can always change that back in my PR.

packages/nhost-js/tests/nhost.test.ts
plmercereau: yes,
everything in `tests` is related to unit tests. They don't run tests against a real backend such as the ones put in the `e2e` folder. In our unit tests, we are checking the instantiation of the backend url(s) of Nhost client, but not its effective functioning.
The `nhost.graphql` client is tested against a Nhost CLI backend in the `e2e` part, and parts (auth & storage) are tested in their respective packages.

In our CI, all the unit tests are run together using turborepo with `pnpm run test`, whereas `e2e` each require the CLI to start, with a possible custom configuration defined in `package/x/nhost`. They cannot be run in parallel in the same machine (In GitHub actions, they are run in parallel on separate machines).

PR #1086, opened 2022-11-02
chore: upgrade msw

plmercereau:
MSW recommends to listen/close server in the `beforeAll`/`afterAll` hooks, but it is not working when upgrading to the latest version.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -66,6 +66,6 @@
   "devDependencies": {
     "@simplewebauthn/typescript-types": "^6.0.0",
     "@types/js-cookie": "^3.0.2",
-    "msw": "^0.39.2"
+    "msw": "^0.47.4"
   }
 }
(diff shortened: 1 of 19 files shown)


szilarddoro:
What do you mean it's not working?

plmercereau:
from the `main` branch:
```sh
cd packages/core
pnpm update msw@latest
pnpm run test
```

szilarddoro:
This only occurs in test files where multiple `describe` blocks are defined, so my guess is that vitest tries to run these in parallel and the server can't be opened for some reason.

```ts
describe('...', () => {
  beforeAll(() => server.listen({ onUnhandledRequest: 'error' }))
  afterAll(() => server.close())

  // ...
})
```

Moving the following code from the describe block in `refreshToken.test.ts` and `enableMfa.test.ts` seems to resolve the issue:

```ts
beforeAll(() => server.listen({ onUnhandledRequest: 'error' }))
afterAll(() => server.close())

// first
describe('...', () => { ... })

// second
describe('...', () => { ... })

// ...

// nth
describe('...', () => { ... })
```

szilarddoro (approved):

PR #1108, opened 2022-11-09
Refactor/merge core

plmercereau:
problems:
- `@nhost/hasura-storage-js` depends on `@nhost/hasura-auth-js` for its types
- not new: when A depends on B depends on C, A needs to define both B and C as a dependency. Annoying.


szilarddoro:
Can't we just simply move the types that are required by `hasura-storage-js` to that package?

plmercereau:
> Can't we just simply move the types that are required by `hasura-storage-js` to that package?

it requires the auth client interface, I see three options
1. leave it as is. Not ideal as storage depends on auth
2. copy the types in both packages
3. move the shared types into something like @nhost/types. It is something we should consider in the future but I don't know if it is a good idea to do it now

szilarddoro:
> > Can't we just simply move the types that are required by `hasura-storage-js` to that package?
> 
> it requires the auth client interface, I see three options
> 
> 1. leave it as is. Not ideal as storage depends on auth
> 2. copy the types in both packages
> 3. move the shared types into something like @nhost/types. It is something we should consider in the future but I don't know if it is a good idea to do it now

I'm fine with having a separate `@nhost/types` as a link between these packages, but that should be prepared in a separate PR IMO. A trade-off, for now, could be to duplicate types in both packages, but I just checked and `hasura-storage-js` is relying on the `HasuraAuthClient` class which boils down to several other types, so I'm not sure.

plmercereau:
> > > Can't we just simply move the types that are required by `hasura-storage-js` to that package?
> > 
> > 
> > it requires the auth client interface, I see three options
> > 
> > 1. leave it as is. Not ideal as storage depends on auth
> > 2. copy the types in both packages
> > 3. move the shared types into something like @nhost/types. It is something we should consider in the future but I don't know if it is a good idea to do it now
> 
> I'm fine with having a separate `@nhost/types` as a link between these packages, but that should be prepared in a separate PR IMO. A trade-off, for now, could be to duplicate types in both packages, but I just checked and `hasura-storage-js` is relying on the `HasuraAuthClient` class which boils down to several other types, so I'm not sure.

yes the cleanest way would be to have this `@nhost/types`. But if we do so, we should solve this problem first, I think:

> - not new: when A depends on B depends on C, A needs to define both B and C as a dependency. Annoying.

szilarddoro:
I don't think that'd be a problem if we have `@nhost/types`, `@nhost/hasura-auth-js` and `@nhost/hasura-storage-js`, as we would end up with this structure, wouldn't we?

Screenshot 2022-11-10 at 9 41 26

plmercereau:
> I don't think that'd be a problem if we have `@nhost/types`, `@nhost/hasura-auth-js`, and `@nhost/hasura-storage-js`, as we would end up with this structure, wouldn't we?
> 
> Screenshot 2022-11-10 at 9 41 26
> 
> So `A` depends on `C` and `B` depends on `C` in this scenario.

that's not what I meant, see for instance https://github.com/nhost/nhost/blob/refactor/merge-core/packages/react/package.json#L64-L66

In the above example, we added `@nhost/nhost-js` as a dependency, that has auth and storage as dependencies.
But if we want to do `import { blah } from '@nhost/hasura-auth-js'` then we need to explicitly add `@nhost/hasura-auth-js` as a dependency too.

plmercereau:
> I don't think that's a problem for us. I mean, we are not including for example React's dependencies in our `package.json` just because it depends on something. Package managers orchestrate this whole thing in the background, don't they?

I'm not saying we should add indirect dependencies (we shouldn't), but inside the monorepo we have to for a reason I don't understand. I'll check if it is still the case

plmercereau:
Regardless, it feels a bit weird to define the Auth client type outside of the Auth package

nunopato:
@plmercereau given that `nhost-js` depends on both `hasura-auth-js` and `hasura-storage-js`, can't we simply re-export all types from both and then use only `nhost-js`? I remember we discussed this some time ago but I am not sure what is the technical reason to not do this.

```js
import { FromAuth, FromStorage } from "@nhost/nhost-js" 
```

plmercereau:
> @plmercereau given that `nhost-js` depends on both `hasura-auth-js` and `hasura-storage-js`, can't we simply re-export all types from both and then use only `nhost-js`? I remember we discussed this some time ago but I am not sure what is the technical reason to not do this.
> 
> ```js
> import { FromAuth, FromStorage } from "@nhost/nhost-js" 
> ```

the problem is `hasura-storage-js` depends on the types of `hasura-auth-js`, and we can't use `nhost-js` as a dependency as it would create a circular dependency, so I'm afraid it won't solve the current problem

packages/hasura-auth-js/auth.typedoc.json
```diff
@@ -1,8 +1,12 @@
 {
   "$schema": "https://typedoc.org/schema.json",
-  "entryPoints": ["src/index.ts"],
+  "entryPoints": [
+    "src/hasura-auth-client.ts"
```
szilarddoro, line 4: I don't think we should change this. `src/index.ts` is the file exporting the relevant types. If we change this to `src/hasura-auth-client.ts`, then TypeDoc will complain about missing type exports.

e.g:
```
./src/types/client.ts:91:0 - warning SendVerificationEmailParams is referenced by HasuraAuthClient.sendVerificationEmail.sendVerificationEmail.__namedParameters but not included in the documentation.
./src/types/client.ts:96:0 - warning ChangeEmailParams is referenced by HasuraAuthClient.changeEmail.changeEmail.__namedParameters but not included in the documentation.
./src/types/client.ts:101:0 - warning DeanonymizeParams is referenced by HasuraAuthClient.deanonymize.deanonymize.params but not included in the documentation.
./src/types/client.ts:119:0 - warning OnTokenChangedFunction is referenced by HasuraAuthClient.onTokenChanged.onTokenChanged.fn but not included in the documentation.
./src/types/client.ts:117:0 - warning AuthChangedFunction is referenced by HasuraAuthClient.onAuthStateChanged.onAuthStateChanged.fn but not included in the documentation.
```

Did you experience any problems with `src/index.ts`?

plmercereau: Yes I did, you can reproduce with
```sh
pnpm run docgen
pnpm run build:docs
```

It seems the generated markdown files are not always correct.
For instance `customSet` in `src/types/local-storage.ts`

szilarddoro: Yes, the `|` character drives the markdown crazy. I'll prepare a fix for that problem soon. I don't want to lose this many types from our docs.

plmercereau:
@szilarddoro I don't understand why the dashboard now fails to build

plmercereau:
Closing as this is not in our current projects cycle

PR #1110, opened 2022-11-10
Correct client url

elitan:
Vite is using `http://127.0.0.1:5173`.


plmercereau (changes requested):
The CI is relying on that.
I would rather adapt the dev script:
```js
    "dev": "vite --port 3000",
```

plmercereau (approved):

PR #1114, opened 2022-11-11
feat (dashboard): Dashboard Deployment

szilarddoro:


plmercereau:
Internal dependencies in examples should be `"*"` so each example can be copy-pasted outside of the monorepo.
It would have been a problem if we would have published example packages, but we aren't.

szilarddoro:
> Internal dependencies in examples should be `"*"` so each example can be copy-pasted outside of the monorepo.
> It would have been a problem if we would have published example packages, but we aren't.

I'm just testing things here.

szilarddoro:
How likely is this "people copying an example outside of the monorepo" scenario btw? I always found that a strange reasoning TBH.

plmercereau:
Users that would want to use an example as a boilerplate.
Also our examples won't work in codesandbox/stackblitz

elitan (approved):

PR #1127, opened 2022-11-14
Correct handling of network error

plmercereau:
rationale: a contributor is preparing a Nuxt 3 module. They need either Axios or cross-fetch. We talked about replacing Axios with cross-fetch, but using Axios v1 is already an improvement.
- bump to Axios v1 (except in the dashboard). Adapt types
- it uncovered an issue with the network code that was not processed correctly: it was supposed to be `0` whereas it was `200`
- it then uncovered a bug: the auto-signin was not retrying the existing token in the url when there was no network
- remove deprecated `faker` methods

### DIFF
--- a/.changeset/lovely-cheetahs-suffer.md
+++ b/.changeset/lovely-cheetahs-suffer.md
@@ -1,0 +1,5 @@
+---
+'@nhost/core': patch
+---
+
+Retry auto-signin when starting offline
--- a/.changeset/nice-hornets-listen.md
+++ b/.changeset/nice-hornets-listen.md
@@ -1,0 +1,5 @@
+---
+'@nhost/core': patch
+---
+
+Return the correct error code for a network error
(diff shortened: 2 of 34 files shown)


elitan (approved):

szilarddoro (approved):
I think we should consider migrating to cross-fetch (or something similar) soon. Fetch API is so convenient to use, and also, we would use the same approach everywhere. There is only one Axios call in the dashboard, the rest is using cross-fetch.

elitan:
Yes, we should move away from Axios eventually.

plmercereau:
> I think we should consider migrating to cross-fetch (or something similar) soon. Fetch API is so convenient to use, and also, we would use the same approach everywhere. There is only one Axios call in the dashboard, the rest is using cross-fetch.

Yes, we should use cross-fetch. There will be some pitfalls (for instance I don't know if there is a similar pattern as axios intercepters in cross-fetch), but it will make our lives better in the end

PR #1145, opened 2022-11-18
fix(dashboard): avoid using `BACKEND_URL` locally

szilarddoro:


dashboard/src/utils/helpers.ts
```diff
@@ -56,6 +56,10 @@ export function getCurrentEnvironment(): Environment {
 }
 
 export function generateRemoteAppUrl(subdomain: string): string {
+  if (process.env.NEXT_PUBLIC_NHOST_PLATFORM !== 'true') {
+    return 'http://localhost:1337';
```
szilarddoro, line 60: @plmercereau here I was thinking, that maybe we should expose the base URL through the `NhostClient`. I can access the URL of every service, but not the "app URL" through the client. What do you think?

plmercereau (approved):

PR #1147, opened 2022-11-19
ci: fine-tune the dashboard release

plmercereau:
- Always run the dashboard tests before publishing the docker image
- Use Docker cache so we can reuse layers already built in previous runs or jobs (for instance the `pnpm install` layer :) )
- Create a GitHub release with the `changesets` changelog
- Improve the readability of the `git tag` attempt. Determine if the dashboard should be tagged+published+released in the `version` job, but actually `git tag` in the `publish` job so it can be untagged if the publishing failed.

@szilarddoro not tested yet on a fork, will do at the beginning the the week
I also prepared [a similar logic for hasura-auth](https://github.com/nhost/hasura-auth/pull/285)


.github/workflows/dashboard.yaml
```diff
@@ -1,11 +1,7 @@
 name: 'Dashboard'
 
 on:
-  push:
```
szilarddoro, line 4: I think we can keep this workflow when pushing to `main`.

.github/workflows/changesets.yaml
```diff
@@ -38,24 +40,36 @@ jobs:
...
+          GIT_TAG="${{ env.DASHBOARD_PACKAGE}}@$DASHBOARD_VERSION"
+          if [ -z "$(git tag -l | grep $GIT_TAG)" ]; then
+            echo "dashboardVersion=$DASHBOARD_VERSION" >> $GITHUB_OUTPUT
+          fi
 
-  publish:
-    runs-on: ubuntu-latest
+  test:
     needs: version
+    name: Run dashboard tests
```
szilarddoro, line 55:
```suggestion
    name: Run Dashboard tests
```

.github/workflows/changesets.yaml
```diff
@@ -38,24 +40,36 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
-      - name: Tag Dashboard
+      - name: Check dashboard tag
```
szilarddoro, line 43:
```suggestion
      - name: Check Dashboard tag
```

.github/workflows/dashboard.yaml
plmercereau: we can, but the workflow would run three times for the same commit:
- in the PR (check before merge)
- in the changeset workflow (check before release)
- in the dashboard workflow

szilarddoro (approved):

szilarddoro:
Let's test on a fork before merging it, but the code looks good.

plmercereau:
I corrected some other issues:
-  the GH release step was not correctly set
- I referenced an env var `$GIT_TAG` instead of an ouput value
- the Turbo token/team were not propagated - and the team was sometimes hard coded
- the custom cache in the `dashboard` workflow was not necessary as we already set a cache in the shared action

I will run the stuff in a fork

PR #1150, opened 2022-11-20
added test with large file

elitan:


plmercereau:
That would be nice to avoid having to download 25MB more because of this file when cloning the repo. Maybe a generated buffer when running the tests instead?

PR #1158, opened 2022-11-21
fix: 🐛 make `useUserRoles` reactive

plmercereau:


packages/react/src/useUserRoles.ts
```diff
@@ -16,7 +16,6 @@ export const useUserRoles = () => {
   const service = useAuthInterpreter()
   return useSelector(
     service,
-    (state) => state.context.user?.roles || [],
-    (a, b) => a.every((i) => b.includes(i) && b.every((i) => a.includes(i)))
+    (state) => (state.matches('authentication.signedIn') && state.context.user?.roles) || []
```
szilarddoro, line 19: This makes the code a bit more readable for me.

```suggestion
    (state) => {
      if (!state.matches('authentication.signedIn') {
        return []
      }

      return state.context.user?.roles) || []
    }
```

Shouldn't we apply the same change to the Vue composable?

PR #1164, opened 2022-11-22
Vue docs fix: «client» → «nhost»

alexander-mart:
https://docs.nhost.io/quickstarts/vue#update-user-data


alexander-mart:
This is right? :)

elitan (approved):
Thank you for this!

PR #1172, opened 2022-11-22
chore: nextjs 13

### LINKED ISSUE #1170: nhost/nhost-js should allow next 13 as peer dependency
CleanShot 2022-11-22 at 09 31 54

plmercereau:
Closes #1170

@elitan any possible breaking change introduced by Next 13?

### DIFF
--- a/.changeset/hip-coats-pump.md
+++ b/.changeset/hip-coats-pump.md
@@ -1,0 +1,5 @@
+---
+'@nhost/nextjs': patch
+---
+
+Allow NextJs 13 as peer dependency
--- a/packages/nextjs/package.json
+++ b/packages/nextjs/package.json
@@ -71,7 +71,7 @@
   },
   "peerDependencies": {
     "@nhost/react": "workspace:*",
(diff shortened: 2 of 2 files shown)


szilarddoro:
These are the breaking changes:
- https://nextjs.org/blog/next-13#breaking-changes

The minimum React version and the minimum Node version bumps may be the most important.

plmercereau:
> These are the breaking changes:
> 
> * https://nextjs.org/blog/next-13#breaking-changes
> 
> The minimum React version and the minimum Node version bumps may be the most important.

but using the right version of Node and React with Next 13 is not the Nhost SDK problem right?
I meant breaking changes specific to the interaction between Next 13 and the SDK

plmercereau:
> One note though: Maybe we want to set the required peer dependency version to `^12.0.10 || ^13.0.0` so that any Next.js 13.x version would be good for the SDK as a peer dependency.

yes, done

szilarddoro (approved):

PR #1181, opened 2022-11-22
feat(dashboard): make backend port configurable

szilarddoro:
I created an environment variable (`NEXT_PUBLIC_NHOST_LOCAL_BACKEND_PORT`) to let the CLI pass a custom port when a user wants to change it.

I've made some changes to other environment variables to make them more consistent:

`NEXT_PUBLIC_NHOST_MIGRATIONS_URL` -> `NEXT_PUBLIC_NHOST_MIGRATIONS_PORT`
`NEXT_PUBLIC_NHOST_HASURA_URL` -> `NEXT_PUBLIC_NHOST_HASURA_PORT`

`...NHOST_MIGRATIONS_PORT` and `...NHOST_HASURA_PORT` only make sense in a local environment. I think we want to use `...NHOST_LOCAL...` here too.

### DIFF
--- a/.changeset/nice-feet-work.md
+++ b/.changeset/nice-feet-work.md
@@ -1,0 +1,5 @@
+---
+'@nhost/dashboard': minor
+---
+
+feat(dashboard): make backend port configurable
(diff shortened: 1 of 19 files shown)


elephant3 (commented):
LGTM from my side. Built and tested Dockerfile locally with CLI. But a proper review from front-end folks is required

nunopato (approved):

PR #1194, opened 2022-11-23
feat(dashboard): add Vercel deployment

szilarddoro:
This PR adds the ability to trigger production deployments when a new tag is created for the Dashboard.

### DIFF
--- a/.github/workflows/changesets.yaml
+++ b/.github/workflows/changesets.yaml
@@ -18,6 +18,7 @@ env:
 
 jobs:
   version:
+    name: Version
     runs-on: ubuntu-latest
     outputs:
       hasChangesets: ${{ steps.changesets.outputs.hasChangesets }}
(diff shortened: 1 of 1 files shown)


.github/workflows/changesets.yaml
```diff
@@ -117,6 +126,11 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
```
szilarddoro, line 126: How does this part of the GH workflow work @plmercereau? I don't see any release created for the dashboard.

This seems to be a good alternative, although I haven't tested it yet: https://github.com/softprops/action-gh-release

plmercereau: this step reads the changelog generated by changesets and releases a version from the given tag.
This follows the same logic as the internals of the changesets GH action

plmercereau: You don't see any gh release yet because this step has never run since set - the two last releases failed because of the Next env vars issue

plmercereau: From the [fork](https://github.com/plmercereau/nhost/actions/runs/3514555552/jobs/5888772224)

szilarddoro: Ah, OK. Sorry for the false alarm. ⏰

plmercereau (approved):

PR #1215, opened 2022-11-25
feat(dashboard): add Environment Variables page

szilarddoro:


eugene-reuka (commented):
Looks great. 

Two comments: 
1) Replace "Enter Value" placeholder for Name field with "EXAMPLE_NAME"
2) allow entering lower case and make them upper case automatically

.changeset/tender-monkeys-beg.md
```diff
@@ -0,0 +1,5 @@
+---
+'@nhost/nhost-js': patch
+---
+
+chore(nhost-js): expose `url` in Functions and GraphQL client
```
szilarddoro, line 5: @plmercereau I know `url` was already exposed in the GraphQL client through `client.graphql.getUrl`, but it's `client.(auth|storage).url` in other packages. Not sure which way to pick though. Maybe it's better if we use a getter function everywhere.

elitan (approved):

PR #1217, opened 2022-11-25
fix(changesets): add missing `pnpm` command, pre-build project

szilarddoro:


szilarddoro:
What do you about this change in the publish step @plmercereau? I was thinking, that maybe it's better if we create a separate "Publish to Vercel" step that simply depends on the Docker publish.

plmercereau:
why would we want to add this step? it is not needed in the next steps, and as it is already executed in a dependent job the turborepo remote cache should already be available when building the images.

Yes, why not creating a separate job for Vercel

plmercereau:
> why would we want to add this step? it is not needed in the next steps, and as it is already executed in a dependent job the turborepo remote cache should already be available when building the images.
> 
> Yes, why not creating a separate job for Vercel

Oh, i missed the fact the build+deploy to vercel was happening inside the GH action - thought we were still triggering the hook.
Nice

.github/workflows/changesets.yaml
```diff
@@ -129,3 +122,26 @@ jobs:
...
+        with:
+          TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
+          TURBO_TEAM: ${{ secrets.TURBO_TEAM }}
+      - name: Setup Vercel CLI
+        run: pnpm add -g vercel
+      - name: Trigger a Vercel deployment
+        env:
+          VERCEL_ORG_ID: ${{ secrets.DASHBOARD_VERCEL_TEAM_ID }}
+          VERCEL_PROJECT_ID: ${{ secrets.DASHBOARD_VERCEL_PROJECT_ID }}
+        run: vercel pull --environment=production && vercel build --prod && vercel deploy --prebuilt --prod
```
plmercereau, line 147: Maybe multiple lines instead of multiple `&&` to improve readability
GH actions have `set -e` anyway

plmercereau (approved):

.github/workflows/changesets.yaml
szilarddoro: > > why would we want to add this step? it is not needed in the next steps, and as it is already executed in a dependent job the turborepo remote cache should already be available when building the images.
> > Yes, why not creating a separate job for Vercel
> 
> Oh, i missed the fact the build+deploy to vercel was happening inside the GH action - thought we were still triggering the hook. Nice

Nope. Vercel runs the build step ignore script even if we trigger the deployment through the webhook. We ignored `main` branch on purpose for our production deployment, but that meant that even webhooks would be ignored, so we need to use the CLI.

szilarddoro:
> Oh, i missed the fact the build+deploy to vercel was happening inside the GH action - thought we were still triggering the hook. Nice

Nope. Vercel runs the build step ignore script even if we trigger the deployment through the webhook. We ignored `main` branch on purpose for our production deployment, but that meant that even webhooks would be ignored, so we need to use the CLI.

PR #1227, opened 2022-11-27
fix(changesets): add Vercel deployment token to CI

szilarddoro:


.github/workflows/changesets.yaml
```diff
@@ -144,7 +144,8 @@ jobs:
...
           VERCEL_ORG_ID: ${{ secrets.DASHBOARD_VERCEL_TEAM_ID }}
           VERCEL_PROJECT_ID: ${{ secrets.DASHBOARD_VERCEL_PROJECT_ID }}
+          VERCEL_DEPLOY_TOKEN: ${{ secrets.DASHBOARD_VERCEL_DEPLOY_TOKEN }}
         run: |
-          vercel pull --environment=production
-          vercel build --prod
-          vercel deploy --prebuilt --prod
+          vercel pull --environment=production --token=$VERCEL_DEPLOY_TOKEN
+          vercel build --prod --token=$VERCEL_DEPLOY_TOKEN
+          vercel deploy --prebuilt --prod --token=$VERCEL_DEPLOY_TOKEN
```
plmercereau, line 151: the `--token` comes to replace the environment variable, right?
In that case, you can write things like
```
vercel pull --environment=production --token=${{ secrets.DASHBOARD_VERCEL_DEPLOY_TOKEN }}
```
and get rid of `VERCEL_DEPLOY_TOKEN: ${{ secrets.DASHBOARD_VERCEL_DEPLOY_TOKEN }}`
If you keep it, we tend to believe `VERCEL_DEPLOY_TOKEN` is some kind of system env var used by Vercel

plmercereau (approved):

.github/workflows/changesets.yaml
szilarddoro: The definition right above these 3 commands, but if you find it more readable, I can change it to your suggestion.

PR #1233, opened 2022-11-28
chore: remove deprecated methods

plmercereau:


szilarddoro (commented):
I know that's going to be a breaking change, but instead of just hiding the from the documentation, we should remove them entirely. I'm afraid that we'd just push this thing in front of us indefinitely if we don't create some kind of action plan for this.

plmercereau:
> I know that's going to be a breaking change, but instead of just hiding the from the documentation, we should remove them entirely. I'm afraid that we'd just push this thing in front of us indefinitely if we don't create some kind of action plan for this.

I'm fine with the suggestion. @elitan ?

elitan (approved):

plmercereau:
I then removed everything that was deprecated (all of which was deprecated since more than 6 months).
I think this is a major bump too, wdyt?

szilarddoro (approved):
yes-i-like-a42bb6

PR #1235, opened 2022-11-28
Feat: use the `graphql-request` client at `nhost.graphql`

plmercereau:
- Use the `graphql-request` client at `nhost.graphql
- Add new getters `nhost.graphql.httpUrl` and `nhost.graphql.wsUrl`, and deprecate `nhost.graphql.getUrl()`


notes.md
```diff
@@ -0,0 +1,8 @@
+# Note - remove this file
+
+- adjust existing documentation
+  - see how docgen behaves
+- change in the `nhost.auth.request` signature:
+  - other use of generic types
+  - the result is not the same
+- think a bit more about subscriptions and the ws client
```
elitan, line 8: remove

.changeset/clean-pandas-act.md
```diff
@@ -0,0 +1,58 @@
+---
+'@nhost/nhost-js': minor
```
elitan, line 2: Let's respect semver. I think this should be a major update. What do you think @nunopato?

.changeset/clean-pandas-act.md
```diff
@@ -0,0 +1,58 @@
...
+
+After:
+
+```js
+try {
+  const data = await nhost.graphql.request('{ myQuery { id } }')
+} catch (error) {
+  console.log(error)
+}
+```
```
elitan, line 30: I'm not sure we should move away from the previous `{ data, error}` approach because we're using it in other functions like `auth` and `storage`.

Is it possible to still use `{ data, error }` even with this package update?

.changeset/real-cats-fly.md
```diff
@@ -0,0 +1,7 @@
+---
+'@nhost/nhost-js': minor
+---
+
+Add new getters `nhost.graphql.httpUrl` and `nhost.graphql.wsUrl`
```
elitan, line 5:
```suggestion
Add new getters `nhost.graphql.getHttpUrl` and `nhost.graphql.getWsUrl`
```

elitan (changes requested):

.changeset/clean-pandas-act.md
plmercereau: We can, but the trade-off is it won't be compatible with graphql-codegen or another tool meant to work with graphql-request.
Graphql-request also returns errors (plural) whereas our current signature is singular

.changeset/clean-pandas-act.md
plmercereau: yes

packages/nhost-js/src/clients/graphql.ts
```diff
@@ -151,15 +103,26 @@ export class NhostGraphqlClient {
...
   setAccessToken(accessToken: string | undefined) {
-    if (!accessToken) {
-      this.accessToken = null
-      return
-    }
+    this.accessToken = accessToken || null
+    this.resetHeaders()
+  }
+
+  setEndpoint(value: string): GraphQLClient {
```
nunopato, line 110: if this is a setter, should it return a `GraphQlClient`?

packages/nhost-js/src/clients/graphql.ts
```diff
@@ -151,15 +103,26 @@ export class NhostGraphqlClient {
...
+    this._url = value
+    super.setEndpoint(value)
+    return this
+  }
+
+  setHeaders(value: Record<string, string>): GraphQLClient {
```
nunopato, line 116: same here

nunopato (commented):
would it make sense to avoid the breaking change now and have it introduced at a later stage with other major changes?

.changeset/clean-pandas-act.md
nunopato: I agree. The question is whether we can avoid introducing it now.

packages/nhost-js/src/clients/graphql.ts
plmercereau: no, that is not required. Setters in js usually return `void`

szilarddoro (commented):
I agree with the comments on setter return values (they shouldn't have one). Other than that, I think the PR is fine.

plmercereau:
> I agree with the comments on setter return values (they shouldn't have one). Other than that, I think the PR is fine.

Just to clarify: I returned `this` on the setters inherited from the graphql-request client. I wouldn't have otherwise

packages/nhost-js/src/clients/graphql.ts
nunopato: exactly, that is what I mean. We are returning `this` which is the `GraphQlClient`.

elitan:
> would it make sense to avoid the breaking change now and have it introduced at a later stage with other major changes?

Yes.

We have this PR and also https://github.com/nhost/nhost/pull/1246 that would can do in a single major update.

Do we have anything else in the pipeline for a major update?

nunopato:
> Just to clarify: I returned this on the setters inherited from the graphql-request client. I wouldn't have otherwise

not sure I understand, sorry. Do you mean that `graphql-request` is exposing the same setters and are returning a value?

plmercereau:
> > Just to clarify: I returned this on the setters inherited from the graphql-request client. I wouldn't have otherwise
> 
> not sure I understand, sorry. Do you mean that `graphql-request` is exposing the same setters and are returning a value?

Yes

szilarddoro (approved):

.changeset/clean-pandas-act.md
elitan: I want to come back to this, why can't the `{ data, error }` pattern be compatible with graphql-codegen? 

What other tools do we expect people to use graphql-request with?

The fact that the previous signature was singular doesn't concern me much. The `data` object here will be different anyways.

nunopato: | Graphql-request also returns errors (plural) whereas our current signature is singular

I don't think we should "tie" our package's API to that of the underlying GraphQL package's API. We want to expose the API we think is "right" while being able to change from `graphql-request` to `urql` in the future, if we want to.

plmercereau: > I want to come back to this, why can't the `{ data, error }` pattern be compatible with graphql-codegen?
> 
> What other tools do we expect people to use graphql-request with where the `{ data, error }` pattern would be incompatible?
> 
> The fact that the previous signature was singular doesn't concern me much. The `data` object here will be different anyways.

it will be imcompatible with the `graphql-request` codegen:
```graphql
query myUsers {
  users {
    id
  }
}
```

```ts
import { NhostClient } from '@nhost/nhost-js'
import { getSdk } from 'generated.ts'

const nhost = new NhostClient()
const { myUsers } = getSdk(client)

const result = await myUsers() // <---- we don't have control over this one
```

elitan: Let's use the new pattern for codegen described here: https://the-guild.dev/blog/typed-document-node

Similar to this syntax:

```js
import { request } from 'graphql-request'
import { graphql } from '../gql/gql'

const GET POSTS = graphql(`
  query GetPosts {
    posts {
      id
      title
    }
  }
`)

request('https://some-endpoint/graphql', GET_POSTS)
```

graphql-request ([and others](https://github.com/dotansimha/graphql-typed-document-node#built-in-support)) have support for this using TypedDocumentNode. 

- https://github.com/prisma-labs/graphql-request/blob/c17f82138e5c0e7960fd1fb2ba9ef439323d57dc/src/index.ts#L290

The equivalent syntax for our SDK would be:


```js
import { nhost } from '@/utils/nhost'
import { graphql } from '../gql/gql'

const GET POSTS = graphql(`
  query GetPosts {
    posts {
      id
      title
    }
  }
`)

const { data, error } = nhost.graphql.request(GET_POSTS);
```

What do you think?

plmercereau: why not

plmercereau: then the `errors` vs `error`: we would only pick the first one?

elitan: Hmm, I think we should use `errors` like this: `{ data, errors }`. This way, we don't magically hide anything from developers.

`errors` (plural) is what the GraphQL server responds with, so it makes sense to use `errors` instead of `error`.

plmercereau:
After a final discussion with @elitan we decided to keep this PR aside as using `graphql-request` while keeping our API signature consistent would not create value, given that:
- it is easier to adapt our current client to use generated TypeDocumentNode code
- the `graphql-request` api could have worked while using its `rawRequest`, but we couldn't find a way to make it work

I will therefore enhance our current client in the [PR that implements `cross-fetch`](https://github.com/nhost/nhost/pull/1351)

PR #1246, opened 2022-11-29
refactor: decommission `@nhost/core`

plmercereau:
Move the `@nhost/core` to `@nhost/hasura-auth-js` and `@nhost/hasura-storage-js`


config/tsconfig.base.json
```diff
@@ -28,22 +28,47 @@
     "resolveJsonModule": true,
     "esModuleInterop": true,
     "sourceMap": true,
-    "types": ["node"],
+    "types": [
```
szilarddoro, line 31: Have you changed something in your formatter configuration? It's not the same as the output my code editor has.

plmercereau: no, I haven't, but talking about that, the `source.organizeImports` vscode option conflicts with the lint sort rule
I think we should disable this lint rule once and for all...

szilarddoro: I support this idea 😄

szilarddoro (approved):

szilarddoro:
I think we should wait for someone else's approval on this PR.

nunopato (commented):
Is this a major version bump? core's functionality is being moved elsewhere but the API isn't changing, is it?

PR #1269, opened 2022-12-01
ci: improvements on the renovate changeset workflow

plmercereau:
- When pushing the changeset, the PR checks were disappearing. We need to use another PAT so it re-triggers the CI workflow. It does then make more sense to separate the CI and renovate workflows
- An empty changeset was pushed when the renovate version bump had no impact on any npm package (e.g. non-node renovations). It now checks if any npm package changed before creating the file
- the changeset was including dependent packages. It is confusing as we can't easily determine which package is responsible for the bump in reading the changeset, and it is unnecessary as the changeset action already determines the bumps of the dependents

### DIFF
--- a/.changeset/cosmetology-bisexuality-sponsor.md
+++ b/.changeset/cosmetology-bisexuality-sponsor.md
@@ -1,0 +1,5 @@
+---
+"@nhost/sync-versions": patch
+---
+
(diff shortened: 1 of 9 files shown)


plmercereau:
@szilarddoro I'm going to merge this one - all the oncoming renovate PRs don't have mandatory checks activated because of this

PR #1342, opened 2022-12-06
chore: improve internal dependencies

### LINKED ISSUE #1335: `@nhost/nextjs` should have `@nhost/react` as dependency
If `@nhost/nextjs` should have `@nhost/react` as dependency we don't need to install both `@nhost/nextjs` and. `@nhsot/react` when using `@nhost/nextjs`.

This will also help with autocomplete in IDEs because everything that's exported from `@nhost/react` is also exported from `@nhost/nextjs` so it's unambiguous for the user what package to use (both works I think).

It would also be easier to install `@nhost/nextjs` because users wouldn't have to also install `@nhost/react`.

plmercereau:
Vanilla
```
yarn add @nhost/nhost-js graphql
```

Nextjs
```
yarn add next react react-dom
yarn add @nhost/nextjs graphql
```

React
```
yarn add react react-dom
yarn add @nhost/react graphql
```

Vue
```
yarn add vue
yarn add @nhost/vue graphql
```

React Apollo
```
yarn add react react-dom
yarn add @nhost/react graphql
yarn add @nhost/react-apollo @apollo/client
```

Closes #1335


szilarddoro (commented):
Love it ❤️ 

We should also update the documentation (e.g: https://docs.nhost.io/quickstarts/nextjs#configure-nhost-with-nextjs)

packages/nextjs/package.json
```diff
@@ -63,21 +63,20 @@
...
     "xstate": "^4.33.5"
   },
   "peerDependencies": {
-    "@nhost/react": "workspace:*",
     "next": "^12.0.10 || ^13.0.0",
     "react": "^17.0.0 || ^18.0.0",
     "react-dom": "^17.0.0 || ^18.0.0"
   },
   "devDependencies": {
+    "@nhost/nhost-js": "workspace:*",
```
szilarddoro, line 77: Why do we need these as dev dependencies btw? These are all re-exported by @nhost/nhost-js, aren't they?

szilarddoro: Yes, sorry. I was looking at another package.json while writing the comment.

packages/react/package.json
```diff
@@ -61,15 +61,15 @@
...
-    "@nhost/hasura-storage-js": "workspace:*",
     "@nhost/nhost-js": "workspace:*",
     "@xstate/react": "^3.0.1",
     "immer": "^9.0.15",
     "jwt-decode": "^3.1.2",
     "xstate": "^4.33.5"
   },
   "devDependencies": {
+    "@nhost/hasura-auth-js": "workspace:*",
+    "@nhost/hasura-storage-js": "workspace:*",
```
elitan, line 72: Similar question as above, why do we need these packages as dev dependencies? 

We already have `@nhost/nhost-js` as a dependency which includes these dev dependencies.

packages/vue/package.json
```diff
@@ -73,6 +72,7 @@
   },
   "devDependencies": {
     "@nhost/docgen": "workspace:*",
+    "@nhost/hasura-storage-js": "workspace:*",
     "@nhost/hasura-auth-js": "workspace:*",
```
elitan, line 76: Are these packages needed here? `@nhost/nhost-js` is already a dependency.

elitan (approved):

szilarddoro (approved):

nunopato (approved):
:raised_hands:

szilarddoro:
> The dashboard fails to build. It seems this is somehow related to `form-data`, that fails to make a browser vs server check and then calls `window`. I could not find a way to solve this. What I tried:
> 
> * install `form-data` in every package
> * add an alias in the Vite config to [form-data's main lib](https://github.com/form-data/form-data/blob/master/package.json#L10)
> 
> I'm running out of ideas, maybe @szilarddoro could have a look?

https://github.com/form-data/form-data/issues/537

This issue is a little bit concerning.

plmercereau:
> > The dashboard fails to build. It seems this is somehow related to `form-data`, that fails to make a browser vs server check and then calls `window`. I could not find a way to solve this. What I tried:
> > 
> > * install `form-data` in every package
> > * add an alias in the Vite config to [form-data's main lib](https://github.com/form-data/form-data/blob/master/package.json#L10)
> > 
> > I'm running out of ideas, maybe @szilarddoro could have a look?
> 
> [form-data/form-data#537](https://github.com/form-data/form-data/issues/537)
> 
> This issue is a little bit concerning.

Yes it is, and we might need to find a replacement soon. I don't think this is the origin of the problem as before this PR it was working fine. This is somehow related to these changes.
I however tried to replace `form-data` by `formdata-node` and ran into the same issue (after I set `experimental.esmExternals` to `loose` in the Dashboard Nextjs config

PR #1343, opened 2022-12-06
Update vue.mdx

kematzy:
Add `VITE_NHOST_URL` environment variable in example to avoid confusion

### DIFF
--- a/docs/docs/quickstarts/vue.mdx
+++ b/docs/docs/quickstarts/vue.mdx
@@ -177,6 +177,7 @@ Finally, store the environment variables for `subdomain` and `region` in `.env`:
 ```yaml title=".env.development"
 VITE_NHOST_SUBDOMAIN=[subdomain]
 VITE_NHOST_REGION=[region]
+VITE_NHOST_URL='https://[subdomain].nhost.run'
 ```
 
 You find your Nhost project's `subdomain` and `region` in the [project overview](https://app.nhost.io):


elitan:
We should not use `VITE_NHSOT_URL`. Instead, we should update the quickstart to use the following everywhere instead:

```js
const nhost = new NhostClient({
  subdomain: import.meta.env.VITE_NHOST_SUBDOMAIN
  region: import.meta.env.VITE_NHOST_REGION
})
```

@kematzy can you help us with that?

kematzy:
> We should not use `VITE_NHSOT_URL`. Instead, we should update the quickstart to use the following everywhere instead:
> 
> ```js
> const nhost = new NhostClient({
>   subdomain: import.meta.env.VITE_NHOST_SUBDOMAIN
>   region: import.meta.env.VITE_NHOST_REGION
> })
> ```
> 
> @kematzy can you help us with that?

Yes, I'm happy to help with this over the weekend. 

FYI: I was following the Vue Quickstart docs, and only when I added the above ENV variable could I get my local dev version to work. This is why I added it to the docs.

elitan:
Yes, but that's because the initialization of the Nhost client was incorrect. We should use the code below everywhere to avoid using the `VITE_NHOST_URL`.

```js
const nhost = new NhostClient({
  subdomain: import.meta.env.VITE_NHOST_SUBDOMAIN
  region: import.meta.env.VITE_NHOST_REGION
})
```

kematzy:
Please see:
- [nhost/vue-quickstart pull request #1](https://github.com/nhost/vue-quickstart/pull/1)
- [https://github.com/nhost/nhost/pull/1372](https://github.com/nhost/nhost/pull/1372)

I hope these changes are acceptable.

elitan:
Perfect, will check those PRs. Closing this for now.

PR #1351, opened 2022-12-06
refactor: replace axios by cross-fetch

plmercereau:
Work in progress
To do:
- [ ] re-implement upload progress without Axios, but without regression
- [ ] wait for the [`graphql-request` PR](https://github.com/nhost/nhost/pull/1235) to be merged (will solve the Axios dependency in `nhost.graphql`)
- [ ] remove all references to Axios (for instance Vite config)
- [x] `@nhost/hasura-auth-js`
- [x] Dashboard


.changeset/cool-windows-greet.md
```diff
@@ -0,0 +1,7 @@
+---
+'@nhost/nhost-js': minor
+---
+
+Add new getters `nhost.graphql.httpUrl` and `nhost.graphql.wsUrl`
+
+`nhost.graphql.getUrl()` is now deprecated.
```
elitan, line 7: Why is this change set in this PR? Seems unrelated.

packages/hasura-auth-js/e2e/email.test.ts
```diff
@@ -1,12 +1,13 @@
...
+import fetch from 'cross-fetch'
 import { afterEach, describe, expect, it } from 'vitest'
-
 import { auth, getHtmlLink, mailhog } from './helpers'
 
 describe('emails', () => {
   afterEach(async () => {
     await auth.signOut()
+    // * Give hasura-auth enough time to process the the previous test
+    await new Promise((resolve) => setTimeout(resolve, 1000))
```
elitan, line 10: Process what exactly? I don't understand why this is needed.

elitan (changes requested):
Looks good. It would be nice if @szilarddoro took a second look at this PR.

.changeset/cool-windows-greet.md
plmercereau: As I made significant changes in `nhost.graphql`, I also added this change. When implementing `graphql-request` it had  a greater impact in the code, it is less the case here.

elitan: ah okey, I thought this was done in another PR but OK.

Don't we need another changeset for the replacement of axios?

plmercereau: we do - that's the last pending item in the to-do list of this PR

plmercereau: done

elitan: Ah, understood.

packages/hasura-storage-js/src/hasura-storage-api.ts
```diff
@@ -116,12 +109,9 @@ export class HasuraStorageApi {
...
   }
 
-  private generateAuthHeaders():
-    | { Authorization: string }
-    | { 'x-hasura-admin-secret': string }
-    | null {
+  private generateAuthHeaders(): HeadersInit | undefined {
     if (!this.adminSecret && !this.accessToken) {
-      return null
+      return undefined
```
szilarddoro, line 114: Why did you change it to `undefined`?

packages/hasura-storage-js/src/hasura-storage-client.ts
```diff
@@ -85,19 +84,10 @@ export class HasuraStorageClient {
       formData = params.formData
     }
 
-    const { fileMetadata, error } = await this.api.upload({
+    return await this.api.upload({
```
szilarddoro, line 87: We are not wrapping things in a try-catch here, so `await` on `return` is [unnecessary](https://eslint.org/docs/latest/rules/no-return-await).

```suggestion
    return this.api.upload({
```

packages/hasura-storage-js/src/machines/file-upload.ts
```diff
@@ -96,76 +96,41 @@ export const createFileUploadMachine = () =>
...
-                  error: response?.data?.error?.message || message
-                }
+                type: 'UPLOAD_PROGRESS',
+                progress: event.total ? Math.round((loaded * 100) / event.total) : 0,
+                loaded,
+                additions
               })
-            })
+            }
+          }).then(({ fileMetadata, error }) => {
```
szilarddoro, line 122: Can we use `async-await` instead of `then`?

packages/hasura-storage-js/src/machines/file-upload.ts
```diff
@@ -96,76 +96,41 @@ export const createFileUploadMachine = () =>
...
+            if (error) {
+              callback({ type: 'UPLOAD_ERROR', error })
+            } else {
```
szilarddoro, line 125: Empty block

packages/hasura-storage-js/src/utils/upload.ts
```diff
@@ -0,0 +1,118 @@
...
+      return { fileMetadata, error: null }
+    } catch (e) {
+      const error: ErrorPayload = {
+        status: 0,
+        message: (e as Error).message,
+        error: (e as Error).message
+      }
+      return { error, fileMetadata: null }
+    }
+  } else {
```
szilarddoro, line 81: This `else` seems to be unnecessary here

packages/hasura-storage-js/src/utils/upload.ts
```diff
@@ -0,0 +1,118 @@
...
+      if (!response.ok) {
+        const error: ErrorPayload = {
+          status: response.status,
+          message: await response.text(),
+          // * errors from hasura-storage are not codified
+          error: response.statusText
+        }
+        return { error, fileMetadata: null }
+      }
+      const fileMetadata: {
```
szilarddoro, line 60: We already have a type for this, don't we? I think `FileResponse` is the one we should use here. Although, I'm not sure if we should really cast it explicitly considering that this variable immediately assigned to `fileMetadata`.

szilarddoro (changes requested):
I have not checked the exact functionality yet. Will do later.

packages/hasura-storage-js/src/machines/file-upload.ts
plmercereau: we can't here, as the xstate service is running on callback mode. If it finds an `async` in the method, it will then run on promise mode, which we want to avoid

packages/hasura-storage-js/src/hasura-storage-api.ts
plmercereau: because `generateAuthHeaders()` is now used with the `headers` option of `fetch`, that accepts either an `Record<string,string>` or `undefined`

packages/hasura-storage-js/src/hasura-storage-api.ts
```diff
@@ -1,61 +1,51 @@
...
-          ...this.generateAuthHeaders(),
-          'Content-Type': 'multipart/form-data'
-        }
-      })
-
-      return { fileMetadata: res.data, error: null }
-    } catch (error) {
-      return { fileMetadata: null, error: error as Error }
-    }
+    return await fetchUpload(this.url, formData, {
```
szilarddoro, line 28:
```suggestion
    return fetchUpload(this.url, formData, {
```

packages/nhost-js/src/clients/graphql/index.ts
```diff
@@ -76,100 +63,128 @@ export class NhostGraphqlClient {
...
-    // add auth headers if any
-    const headers = {
-      ...this.generateAccessTokenHeaders(),
-      ...config?.headers,
-      'Accept-Encoding': '*'
-    }
-
+  request<T = any, V = Variables>(
+    document: RequestDocument | TypedDocumentNode<T, V>,
+    ...variablesAndRequestHeaders: V extends Record<any, never>
```
szilarddoro, line 68: This type is scary 😨 but I don't see a way of making it much easier to read.

szilarddoro (approved):
LGTM

PR #1352, opened 2022-12-07
Add a section on managing local seed data

wollerman:
I wasn't able to see any docs on seed data, but being able to reset them and manage them is really useful for local development!


elitan:
We have a section about it here: https://docs.nhost.io/database#seed-data

But maybe we could cross-mention it somehow?

Seed data is (for now) only used during local development, so I see your point.

elitan (approved):

elitan:
Thanks

PR #1381, opened 2022-12-14
docs update

elitan:


docs/docs/authentication/users.mdx
```diff
@@ -50,32 +50,32 @@ You're allowed to:
 
 ## Roles
 
-Each user can have one or multiple roles for API requests. You can see the roles of a user and set a default role in Nhost Console under **Users**.
+Each user has one **default role** and an array of **allowed roles**. These roles are used to resolve permissions for requests to [GraphQL](/graphql/permissions) and [Storage](http://localhost:3000/storage#permissions).
 
-Every GraphQL request is made with a specific role. This role will be used to resolve permissions when querying the database. In other words, every user can have multiple roles, but only one role will be applied for any given GraphQL request.
+When the user makes a request, only one role is used to resolve permissions. The default role is used if no role is excplicitly specified. Users can only make requests using the default role or one of the allowed roles.
```
elephant3, line 55: `excplicitly` => `explicitly`

docs/docs/graphql/permissions.mdx
```diff
@@ -30,107 +30,111 @@ The rule above make it so users can only select posts where the value of `user_i
...
 
-Every GraphQL request is resolved based on a **single role**. Roles are added in the Hasura Console when selecting a table and clicking **Permissions**.
-
-If the user is not signed in, the GraphQL API resolves permissions using the `public` role.
+Every GraphQL request resolves permissions using a **single role**.
 
 ### Default Role
 
-Every user have one **default role**. The default role is used to resolve permissions if no other role is specified using the `x-hasura-role` header in the GraphQL request. By default, the default role is `user` for signed-in users.
+Every user has one **default role**. The default role is used to resolve permissions if no role is specified using the `x-hasura-role` header in the GraphQL request. By default, the default role is `user` for authenticated users.
```
elephant3, line 123:
```suggestion
Every user has one **default role**. The default role is used to resolve permissions if no role is specified using the `x-hasura-role` header in the GraphQL request. The `user` is the default role for authenticated users.
```

docs/docs/authentication/users.mdx
```diff
@@ -50,32 +50,32 @@ You're allowed to:
 
 ## Roles
 
-Each user can have one or multiple roles for API requests. You can see the roles of a user and set a default role in Nhost Console under **Users**.
+Each user has one **default role** and an array of **allowed roles**. These roles are used to resolve permissions for requests to [GraphQL](/graphql/permissions) and [Storage](http://localhost:3000/storage#permissions).
```
guicurcio, line 53:
```suggestion
Each user has one **default role** and an array of **allowed roles**. These roles are used to resolve permissions for requests to [GraphQL](/graphql/permissions) and [Storage](/storage#permissions).
```

docs/docs/authentication/users.mdx
```diff
@@ -50,32 +50,32 @@ You're allowed to:
...
 
-Every GraphQL request is made with a specific role. This role will be used to resolve permissions when querying the database. In other words, every user can have multiple roles, but only one role will be applied for any given GraphQL request.
+When the user makes a request, only one role is used to resolve permissions. The default role is used if no role is explicitly specified. Users can only make requests using the default role or one of the allowed roles.
+
+You can manage users' default roles and allowed roles in the Dashboard at **Auth**.
 
 ### Default Role
 
-The default role is used when no role is specified in the GraphQL request. By default, users' default role is `user`.
+The default role is used when no role is specified in the request. By default, users' default role is `user`.
```
guicurcio, line 61: Same sentence as the paragraph above "The default role is used if no role is explicitly specified." it reads repetitively, can we rephrase this or remove it?

docs/docs/authentication/users.mdx
```diff
@@ -139,7 +135,7 @@ await nhost.auth.signUp({
 
 ## Import Users
 
-If you have users in a different system, you can import them into Nhost. When importing users you should insert the users directly into the database instead of using the authentication endpoints (`/signup/email-password`).
+If you have users in a different system, you can import them into Nhost. When importing users you should insert the users directly into the database instead of using the authentication endpoints (`/signup/email-password`) to avoid sending potential transactional emails.
```
guicurcio, line 138:
```suggestion
If you have users in a different system, you can import them into Nhost. When importing users you should insert the users directly into the database instead of using the authentication endpoints (`/signup/email-password`) to avoid sending unnecessary transactional emails.
```

docs/docs/graphql/permissions.mdx
```diff
@@ -30,107 +30,111 @@ The rule above make it so users can only select posts where the value of `user_i
...
+    }
+  }
+}
+```
+
+### Arrays
+
+You must add `[]` at the end of the path for permission variables that are arrays.
+
+**Example**: Let's say you have a permission variable `x-hasura-organization-ids`, the path should be, e.g., `user.profile.organizations.id[]`.
```
guicurcio, line 59: Is `oranizations` in plural in `user.profile.organizations.id\[\]\` correct?

guicurcio (approved):

docs/docs/authentication/index.mdx
```diff
@@ -5,7 +5,7 @@ sidebar_position: 1
 image: /img/og/authentication.png
 ---
 
-Nhost Authentication is a ready-to-use authentication service that is integrated with the [GraphQL API](/graphql) and its permission system from Hasura.
+Nhost Authentication is a ready-to-use authentication service that is integrated with the [GraphQL API](/graphql) and its permission system from Hasura. This allows you to easily add user authentication to your application without having to build and maintain your own authentication system.
```
szilarddoro, line 8: I wonder if “Nhost Authentication” should be called “Nhost Auth” here too. I saw you have changed it in another file.

szilarddoro (approved):

docs/docs/graphql/permissions.mdx
elitan: I don't see us using `oranizations` anywhere

docs/docs/authentication/users.mdx
elitan: I think it's OK. One is a summary, another is specific. It's not always that users read docs from start to finish so sometimes it's worth duplicating this info IMO.

docs/docs/authentication/index.mdx
elitan: I think we're using "Authentication" more than "Auth" so I'll use "Authentication" everywhere for now. 

But I think we should consider using "Auth" instead. That we can do in a follow-up PR.

PR #1382, opened 2022-12-14
feat(dashboard): Auth Management v2

guicurcio:
# feat(dashboard): Auth Management v2

Dashboard

- [x] Creates a new `common` component `Pagination` that is abstract enough to handle any pagination needs in our application (e.g. storage, database viewer).
- [x] Removes users/<userId> page ([userId].tsx) with configuration settings of the project users in favor  of opening a drawer when clicking on a user (as well as handling old URLs of an application pointing to a user configuration page.) Removes `AddUser.tsx` file, removes user's assets pertaining to old page.
- [x] New dialog types (`CREATE_USER`, `EDIT_USER`)
- [x] Structure GraphQL files to follow the same convention as components, utils, etc...
- [x] Handle edits to user roles.
- [x] Parse pagination state from URL (e.g. `users/2/` goes to the second page)

<details> <summary>Auth Management v2 preview</summary> 

https://user-images.githubusercontent.com/20285232/207598023-da551223-7f0b-4d18-8f6a-40a4a50c9d9f.mp4

</details>


dashboard/src/components/users/EditUserForm/EditUserForm.tsx
```diff
@@ -0,0 +1,472 @@
...
+  onDeleteUser,
+}: EditUserFormProps) {
+  const { onDirtyStateChange, openDialog } = useDialog();
+  const { currentApplication } = useCurrentWorkspaceAndApplication();
+
+  const { data } = useGetRolesQuery({
+    variables: { id: currentApplication.id },
+  });
+
+  const allAvailableProjectRoles = useMemo(
```
szilarddoro, line 92: Is `useMemo` really necessary here?

dashboard/src/components/users/EditUserForm/EditUserForm.tsx
```diff
@@ -0,0 +1,472 @@
...
+                component="span"
+                color="error"
+                size="small"
+                label="Banned"
+                className="self-center align-middle"
+              />
+            )}
+          </div>
+          <div>
+            <Select placeholder="Actions" hideEmptyHelperText>
```
szilarddoro, line 196: Let's use a `Dropdown` (using the `menu` prop) here instead of a `Select`. You have plenty of examples to take into consideration.

dashboard/src/components/users/EditUserForm/EditUserForm.tsx
```diff
@@ -0,0 +1,472 @@
...
+              <Chip
+                component="span"
+                color="error"
+                size="small"
+                label="Banned"
+                className="self-center align-middle"
+              />
+            )}
+          </div>
+          <div>
```
szilarddoro, line 195: I'm not sure if we really need this wrapper `div`. Seems unnecessary.

dashboard/src/components/users/EditUserForm/EditUserForm.tsx
```diff
@@ -0,0 +1,472 @@
...
+      },
+    });
+  }
+
+  const signInMethodsAvailable = user.email || user.userProviders.length !== 0;
+
+  return (
+    <FormProvider {...form}>
+      <Form
+        className="divide-y border-y"
```
szilarddoro, line 162: On larger screens, the form doesn't expand to full height. You have plenty of examples to take into consideration.

Screenshot 2022-12-19 at 9 59 51

dashboard/src/components/common/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+export type PaginationProps = DetailedHTMLProps<
+  HTMLProps<HTMLDivElement>,
+  HTMLDivElement
+> & {
+  /**
+   * Total number of pages.
+   */
+  totalNrOfPages: number;
+  elementsPerPage?: number;
```
szilarddoro, line 18: Missing documentation

dashboard/src/components/common/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,146 @@
...
+   */
+  onChangePage: (page: number) => void;
+  /**
+   * Props for component slots.
+   */
+  slotProps?: {
+    /**
+     * Props to be passed to the next button component.
+     */
+    nextButtonProps?: Partial<ButtonProps>;
```
szilarddoro, line 42: `...Props` is redundant for `slotProps`.

```suggestion
    nextButton?: Partial<ButtonProps>;
```

dashboard/src/components/common/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,146 @@
...
+    /**
+     * Props to be passed to the previous button component.
+     */
+    prevButtonProps?: Partial<ButtonProps>;
```
szilarddoro, line 46:
```suggestion
    prevButton?: Partial<ButtonProps>;
```

dashboard/src/components/common/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,146 @@
...
+   */
+  onPrevPageClick: VoidFunction;
+  /**
+   * Function to be called when navigating to the next page.
+   */
+  onNextPageClick: VoidFunction;
+  /**
+   * Props to be passed to the previous button component.
+   */
+  onChangePage: (page: number) => void;
```
szilarddoro, line 34:
```suggestion
  onPageChange: (page: number) => void;
```

You'll need to change references too.

dashboard/src/pages/[workspaceSlug]/[appSlug]/users/index.tsx
```diff
@@ -1,38 +1,379 @@
...
+  const [searchString, setSearchString] = useState<string>('');
+  const [currentPage, setCurrentPage] = useState(1);
+  const [deleteUser] = useRemoteAppDeleteUserMutation({
+    client: remoteProjectGQLClient,
+  });
+  const [updateUser] = useUpdateRemoteAppUserMutation({
+    client: remoteProjectGQLClient,
+  });
+  const limit = useRef(25);
+  const [nrOfPages, setNrOfPages] = useState(1);
```
Comment (szilarddoro) [inline dashboard/src/pages/[workspaceSlug]/[appSlug]/users/index.tsx:48]: I think we should also support storing the current page as a query parameter, just like we do on the Storage page.

szilarddoro (changes requested):
Please let me know if you have any questions that you want to discuss. Solid work btw 🙌🏼

elitan (changes requested):
- password input should be of type `password`.
- Error when creating user is not showing anywhere.
- Don't show "Unsaved changes" when canceling creating a new user.
- Use human readable "ago" text in the list. Use "yyyy-mm-dd hh:mm:ss" in detail.
- Remove "Sign In Methods" column
- Remove auto-generated avatar with letters. Use the actual value from the database.
- Space between avatar and name/email is too short. It should be 16px.
- Closing unsaved detail form should ask to discard.
- How are you able to render "Password is set"? We should not be able to know this because we should not query `passwordHash` from the database.
- Either we render all sign-in methods and mark what sign-in methods the user uses using the "active" chip. Or we render what sign-in methods the user has active and don't render the "active" chip.
- sign-in methods are not left-aligned.
- Default role select shows a gray area.
- cancel and save buttons should be sticky in the bottom.
- "Verified" text should not be bold.
- Having 30 users shows this on the first page: 1-25 of 50 users.

dashboard/src/pages/[workspaceSlug]/[appSlug]/users/index.tsx
Comment (elitan) [reply dashboard/src/pages/[workspaceSlug]/[appSlug]/users/index.tsx:66]: Yes, the current page should be a `?page=<page>` query parameter.

dashboard/src/components/users/UsersBody/UsersBody.tsx
```diff
@@ -0,0 +1,202 @@
...
+              />
+            </span>
+          </div>
+        </div>
+      )}
+
+      <List className="relative w-full h-full">
+        {users?.map((user) => (
+          <Fragment key={user.id}>
+            <ListItem.Root className="w-full h-[64px]">
```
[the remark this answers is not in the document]
guicurcio: Good catch, I've actually seen this happen to the app of one of our users.

dashboard/src/components/users/UsersBody/UsersBody.tsx
```diff
@@ -0,0 +1,202 @@
...
+                    <Avatar src={user.avatarUrl} />
+                  ) : (
+                    <Avatar className="relative inline-flex items-center justify-center w-10 h-10 overflow-hidden bg-gray-300 rounded-full">
+                      <span className="text-xs font-medium text-gray-600 uppercase">
+                        {user.displayName.slice(0, 2)}
+                      </span>
+                    </Avatar>
+                  )}
+                  <div className="grid items-center grid-flow-row">
+                    <Text className="text-sm+ font-medium">
```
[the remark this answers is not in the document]
guicurcio: sort-of, text-sm+ also adds the line-height, but \`leading-5\` is much more correct

dashboard/src/pages/[workspaceSlug]/[appSlug]/users/index.tsx
```diff
@@ -1,38 +1,379 @@
...
+          await axios.post(signUpUrl, {
+            email,
+            password,
+          });
+          await refetchProjectUsers();
+          closeDialog();
+        },
+      },
+      props: {
+        titleProps: { className: 'mx-auto' },
```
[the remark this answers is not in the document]
Comment (guicurcio) [reply dashboard/src/pages/[workspaceSlug]/[appSlug]/users/index.tsx]: Yes, I agree. I have removed this custom prop (and the others) I think it has to do with a minor mismatch between the design system and some of our components.

dashboard/src/pages/[workspaceSlug]/[appSlug]/users/index.tsx
```diff
@@ -1,38 +1,379 @@
...
+        },
+      },
+    });
+
+    await toast.promise(
+      banUser,
+      {
+        loading: user.disabled ? 'Unbanning user...' : 'Banning user...',
+        success: user.disabled
+          ? 'User unbanned succesfully.'
```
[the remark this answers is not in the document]
Comment (guicurcio) [reply dashboard/src/pages/[workspaceSlug]/[appSlug]/users/index.tsx]: I agree, I prefer the active voice as well. I'll make sure to stick by it from now on...

We had an issue about this in the old repo: nhost/console-next#1445 I think we should include as talking point alongside the custom props for dialog, inputs, etc...

dashboard/src/pages/[workspaceSlug]/[appSlug]/users/index.tsx
```diff
@@ -1,38 +1,379 @@
...
+            },
+            toastStyleProps,
+          );
+
+          await refetchProjectUsers();
+        },
+        primaryButtonColor: 'error',
+        primaryButtonText: 'Delete',
+        titleProps: { className: 'mx-auto' },
+        PaperProps: { className: 'max-w-lg mx-auto' },
```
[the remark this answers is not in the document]
Comment (guicurcio) [reply dashboard/src/pages/[workspaceSlug]/[appSlug]/users/index.tsx]: Removed it https://github.com/nhost/nhost/pull/1382#discussion_r1053151605

dashboard/src/components/users/EditUserForm/EditUserForm.tsx
guicurcio: Not even a bit, `getUserRoles` is pretty inexpensive to run and it doesn't depend from data outside itself (https://github.com/nhost/nhost/blob/4a57861354c1332a7043845d0308831565ba4993/dashboard/src/utils/settings/getUserRoles/getUserRoles.ts#L10) I'll actually take it out.

dashboard/src/components/users/UsersBody/UsersBody.tsx
```diff
@@ -0,0 +1,202 @@
...
+                      {user.displayName}
+                    </Text>
+                    <Text className="font-normal text-greyscaleGreyDark">
+                      {user.email}
+                    </Text>
+                  </div>
+                </div>
+
+                <Text
+                  color="greyscaleDark"
```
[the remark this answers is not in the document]
guicurcio: I don't mind... I don't think any different way is that useful actually (given that the `color` prop in our `Text` component takes basically any color.) For example, I would find a good compromise to settle into something like

a `color` prop of union type in our `Text` component with the main colors of our design system and then using Tailwind for custom one-off colors.

dashboard/src/components/users/UsersBody/UsersBody.tsx
```diff
@@ -0,0 +1,202 @@
...
+                          }.svg`}
+                          width={16}
+                          height={16}
+                        />
+                      }
+                    />
+                  ))}
+                </div>
+              </ListItem.Button>
+              <Dropdown.Root>
```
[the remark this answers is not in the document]
guicurcio: I did add this, but the dropdown component through \`secondaryAction\` takes up some width from the container (messing up the table header positions) have you had this happen to you? (it's not a problem in the Roles & Permissions page and Environment Variables given that at most we have 2 columns) I wonder if I should move the classnames a level up :thinking:

dashboard/src/pages/[workspaceSlug]/[appSlug]/users/index.tsx
Comment (guicurcio) [reply dashboard/src/pages/[workspaceSlug]/[appSlug]/users/index.tsx:35]: I'll add this

dashboard/src/pages/[workspaceSlug]/[appSlug]/users/index.tsx
```diff
@@ -1,38 +1,379 @@
...
+    await refetchProjectUsers();
+    closeDrawer();
+  }
+
+  function handleViewUser(user: RemoteAppUser) {
+    openDrawer('EDIT_USER', {
+      title: 'User Details',
+      payload: {
+        user,
+        onEditUser: handleEditUser,
```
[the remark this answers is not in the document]
Comment (guicurcio) [reply dashboard/src/pages/[workspaceSlug]/[appSlug]/users/index.tsx]: > We're better off sending an onSubmit handler that would close the drawer. Then mutations in the EditUserForm component could just simply call it to close the drawer or do perform "after effect".

This sounds good, especially the way that our refetch logic works given that we have two apollo client instances (and the query must use the same variables as the previous time it was run.) Thus, it would make more sense to add any such logic there.

dashboard/src/pages/[workspaceSlug]/[appSlug]/users/index.tsx
```diff
@@ -1,38 +1,379 @@
...
+        success: user.disabled
+          ? 'User unbanned succesfully.'
+          : 'User banned succesfully',
+        error: user.disabled
+          ? 'Error trying to unban user.'
+          : 'Error trying to ban user.',
+      },
+      { ...toastStyleProps },
+    );
+    await refetchProjectUsers();
```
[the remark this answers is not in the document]
Comment (guicurcio) [reply dashboard/src/pages/[workspaceSlug]/[appSlug]/users/index.tsx]: Unfortunately, we cannot easily do it the way we're doing it in the other pages like `Roles & Permissions` because here we are not fetching from our remote but from the user's app (second Apollo GQL client) and also using a \`dialog\` which lives outside the dom nodes from this, I tried this approach initially since it would greatly improve the code but when deleting /updating the user, the Apollo client cannot find the given query \`getRemoteUsersWhere\` this https://github.com/apollographql/apollo-client/issues/5419#issuecomment-598065442 is a good comment about the problem

If we import the query itself, we need to give it the variables for which it was previously run (passing down the props) if not, running the query as is, we would see inconsistent behavior when a user searchs for a user "sam," deletes it, and the query refetchs the initial page.

We could just update the cache directly after mutating a user, but I'm not so sure we all want that...

```typescript
 /**
  * Function to be called after a successful action.
  *
  * @example onSuccessfulAction={() => refetch()}
  * @example onSuccessfulAction={() => router.reload()}
  */
 onSuccessfulAction?: () => Promise<void> | void | Promise<T>;
```

I think something similar to your comment about `onSubmit` work best.

guicurcio:
> *   password input should be of type `password`.
> *   Error when creating user is not showing anywhere.
> *   Don't show "Unsaved changes" when canceling creating a new user.
> *   Use human readable "ago" text in the list. Use "yyyy-mm-dd hh:mm:ss" in detail.
> *   Remove "Sign In Methods" column
> *   Remove auto-generated avatar with letters. Use the actual value from the database.
> *   Space between avatar and name/email is too short. It should be 16px.
> *   Closing unsaved detail form should ask to discard.
> *   How are you able to render "Password is set"? We should not be able to know this because we should not query `passwordHash` from the database.
> *   Either we render all sign-in methods and mark what sign-in methods the user uses using the "active" chip. Or we render what sign-in methods the user has active and don't render the "active" chip.
> *   sign-in methods are not left-aligned.
> *   Default role select shows a gray area.
> *   Cancel and save buttons should be sticky in the bottom.
> *   "Verified" text should not be bold.
> *   Having 30 users shows this on the first page: 1-25 of 50 users.

From all of these, the following are done:

### Done:

\[x\] password input should be of type password.  
\[x\] Don't show "Unsaved changes" when canceling creating a new user.  
\[x\] Use human readable "ago" text in the list. Use "yyyy-mm-dd hh:mm:ss" in detail.  
\[x\] Remove "Sign In Methods" column  
\[x\] Remove auto-generated avatar with letters. Use the actual value from the database.  
\[x\] Space between avatar and name/email is too short. It should be 16px.  
\[x\] Closing unsaved detail form should ask to discard.  
\[x\] Either we render all sign-in methods and mark what sign-in methods the user uses using the "active" chip. Or we render what sign-in methods the user has active and don't render the "active" chip.  
\[x\] sign-in methods are not left-aligned.  
\[x\] Cancel and save buttons should be sticky in the bottom.  
\[x\] Having 30 users shows this on the first page: 1-25 of 50 users.  
\[x\] How are you able to render "Password is set"? We should not be able to know this because we should not query passwordHash from the database.  
\[x\] Default role select shows a gray area.

### Missing:

*    [ ]  Error when creating user is not showing anywhere.

### Not implemented

\> "Verified" text should not be bold.

This one is easy to fix, but I wonder if we should apply the same logic from what Szilard mentioned (https://github.com/nhost/nhost/pull/1382#discussion_r1053151605) right now the default class for the `CheckBox` is a bold text, so we must both make sure that if we add it as a default class name (that the other checkbox we're using we want the bold text) and if we do it as a one-off it looks different from all of our other checkboxes (similar to the Dialog & InputLabel). @szilarddoro wdyt?

dashboard/src/components/users/UsersBody/UsersBody.tsx
```diff
@@ -0,0 +1,202 @@
...
+                      component="span"
+                      color="default"
+                      size="small"
+                      label="Email & Password"
+                      sx={{
+                        paddingLeft: '0.55rem',
+                      }}
+                      icon={
+                        <Image
+                          src="/assets/Envelope.svg"
```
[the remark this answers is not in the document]
guicurcio: We've removed that specific case since we are only rendering the logos from the OAuth providers; I'll double-check all the other providers

dashboard/src/components/users/UsersBody/UsersBody.tsx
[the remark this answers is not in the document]
szilarddoro: I think we should define the basic colors in the theme / `Text` component and make one-off changes with Tailwind. We don't need to use a color prop for this IMO. We should set up variants properly.

dashboard/src/components/users/UsersBody/UsersBody.tsx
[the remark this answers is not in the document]
szilarddoro: That's weird. Could you attach a screenshot about this pls?

guicurcio:
*    [x]  Track the ban state of a user in the drawer
*    [x]  Button instead of an IconButton for changing a user's password
*    [x]   Check the v1 route for signing up users
*    [x]  Close the drawer after deleting a user
*    [x]  Error handling for creating users (`Alert` on top of the create button)
*    [x]  Remove pagination from users not found
*    [x]  Pagination through URL state
*    [x]  show banned users in the users table

elitan (changes requested):
- Password input field should be of type password (change password)
- Changing password is not working (wasn't this working before??)
- No error showing when changing the password is not working
- No need to ask for confirmation when canceling changing password.
- Signed up at in the users list shows "-", but should show "x [unit] ago".
- Banning and then unbanning a user is not working. The second operation does not go through correctly.

dashboard/src/components/common/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,156 @@
...
+  /**
+   * Function to be called when navigating to the previous page.
+   */
+  onPrevPageClick: VoidFunction;
+  /**
+   * Function to be called when navigating to the next page.
+   */
+  onNextPageClick: VoidFunction;
+  /**
+   * Props to be passed to the previous button component.
```
nunopato, line 39:
```suggestion
   * Function to be called when changing the page.
```

dashboard/src/components/common/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,156 @@
...
+          onClick={onNextPageClick}
+          {...slotProps?.nextButton}
+        >
+          Next
+          <ChevronRightIcon className="w-4 h-4" />
+        </Button>
+      </div>
+      <div className="flex flex-row items-center justify-end text-center gap-x-1">
+        <Text className="text-xs text-greyscaleGreyDark">
+          {currentPageNumber === 1 && currentPageNumber}
```
nunopato, line 143: the second flag here, `currentPageNumber`, is redundant. Or maybe you wanted to use another flag.

dashboard/src/components/users/CreateUserForm/CreateUserForm.tsx
```diff
@@ -0,0 +1,157 @@
...
+      if (error.response?.status === 409) {
+        setError('email', {
+          message: error.response.data.message,
+        });
+        return;
+      }
+      setCreateUserFormError(
+        new Error(error.response.data.message || 'Something went wrong.'),
+      );
+    }
```
nunopato, line 91: When I tried to create a new user, the loading spin would be there for a couple of seconds but nothing would happen. We lack the logic here, after the `catch` block, to show a success message and close the dialog.

dashboard/src/components/users/EditUserForm/EditUserForm.tsx
```diff
@@ -0,0 +1,425 @@
...
+                  e.stopPropagation();
+                  copy(user.id, 'User ID');
+                }}
+              >
+                <CopyIcon className="w-4 h-4" />
+              </IconButton>
+            </Text>
+
+            <InputLabel as="h3" className="col-span-1">
+              Created
```
nunopato, line 213:
```suggestion
              Created At
```

dashboard/src/components/common/Pagination/Pagination.tsx
guicurcio: This is correct I think: given that the left hand contains an equals operator it's converted to a condition -- we want to evaluate it in order to show the right-hand (which just happens to be the same one as used in the condition) removing the second one would result in the expression being a boolean and not a string

I could change it for `{currentPageNumber === 1 && 1}` for clarity, adding to the expression below (or abstract it to a function) if it's easier to understand 👍

dashboard/src/components/common/Pagination/Pagination.tsx
guicurcio: Good catch, thanks

guicurcio:
Just pushed the latest changes addressing Johan's review comments:

*    [x]  Password input field should be of type password (change password)
*    [x]  No need to ask for confirmation when canceling changing password.
*    [x]  Signed up at in the users list shows "-", but should show "x [unit] ago".
*    [x]  Banning and then unbanning a user is not working. The second operation does not go through correctly.
*    [x]  Changing password is not working (wasn't this working before??)
 - This was missing the correct GQL client with the user's app config details, so it was probably working before since the handler was moved to the form itself.
*    [x]  No error showing when changing the password is not working
 - Also added logic to check the confirm password which was missing.  .oneOf([Yup.ref('password')], 'Passwords do not match')

Addressed the code review comments by Nuno and Eugene's:
*    [x]   Closing the dialog & re-fetching the users table after creating a user.
*    [x]  Change icon colors and remove shadow in the user's table.

I've moved some the handlers one level down (now instead of passing the handlers to the payload, we just run an function after a successful submit) as suggested by Szilard this helps remove some unnecessary prop drilling.

Also added:
*    [x]   Handling odd pages in the URL (e.g. ?page=10023 goes to the first page.)
*    [x]   Handling users?user=<id> queries in the URL (we show the loader while we search for the user and display it on the users table. We could also run a side effect to open the drawer if the user is found 🤔 )

dashboard/src/pages/[workspaceSlug]/[appSlug]/users/index.tsx
```diff
@@ -1,38 +1,326 @@
...
+  useEffect(() => {
+    if (router.query.page && typeof router.query.page === 'string') {
+      const pageNumber = parseInt(router.query.page, 10);
+      if (nrOfPages >= pageNumber) {
+        setCurrentPage(pageNumber);
+      } else {
+        setCurrentPage(1);
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
Comment (guicurcio) [inline dashboard/src/pages/[workspaceSlug]/[appSlug]/users/index.tsx:97]: This line is here as updating `nrOfPages` was unnecessary running this side-effect.

I'll make a memoized value and delete it

dashboard/src/components/users/EditUserForm/EditUserForm.tsx
```diff
@@ -0,0 +1,471 @@
...
+          <section className="grid grid-flow-row grid-cols-4 gap-8 p-6">
+            <InputLabel as="h3" className="col-span-1">
+              User ID
+            </InputLabel>
+            <Text className="col-span-3 font-medium">
+              {user.id}
+              <IconButton
+                color="secondary"
+                variant="borderless"
+                className="ml-1"
```
elitan, line 245: can we add some padding to make this icon easier to click on?

elitan (approved):
OK, from a UI-test perspective. Good job!

dashboard/src/components/users/UsersBody/UsersBody.tsx
```diff
@@ -0,0 +1,372 @@
...
+                  </Dropdown.Content>
+                </Dropdown.Root>
+              }
+            >
+              <ListItem.Button
+                className="grid lg:grid-cols-6 grid-cols-1 cursor-pointer py-2.5 h-full w-full hover:bg-gray-100 focus:bg-gray-100 focus:outline-none motion-safe:transition-colors"
+                onClick={() => handleViewUser(user)}
+              >
+                <div className="grid grid-flow-col col-span-2 gap-4 place-content-start">
+                  <Avatar src={user.avatarUrl} className="border" />
```
szilarddoro, line 276: please add an `alt` attribute to avatars

szilarddoro:
Screenshot 2022-12-23 at 11 49 53
Screenshot 2022-12-23 at 11 50 36

A few remarks:

- I think we should increase the spacing between the header that contains the filter bar and the table, because it looks weird when there are no users, and also when there are users. The table header is just too close to the filter bar IMO.
- The avatars look very weird now. It's kind of like there should be something, but the network request failed for some reason. This can be seen as a faulty UI.

szilarddoro (changes requested):
A few more remarks:
- We should scroll to the top when a user is using the pagination.
- The spacing looks odd here: Screenshot 2022-12-23 at 13 58 33
- We are not showing a loading state on the "Save" button in the User Details drawer when the user clicks on it.

nunopato (approved):

dashboard/src/components/users/EditUserForm/EditUserForm.tsx
```diff
@@ -0,0 +1,464 @@
...
+              {...register('locale')}
+              id="locale"
+              variant="inline"
+              label="Locale"
+              slotProps={{ root: { className: 'truncate' } }}
+              fullWidth
+              error={!!errors.locale}
+              helperText={errors?.locale?.message}
+            >
+              <Option value="en">en</Option>
```
szilarddoro, line 364: Shouldn't this list be dynamic?

szilarddoro:
Screenshot 2023-01-02 at 12 44 21

I think the gap between the label and the icon is incorrect here.

dashboard/src/components/users/EditUserForm/EditUserForm.tsx
guicurcio: :thinking: Yes, technically this should be dynamic (pull from the `auth_locale_allowed_locales` column of the app), but we should be mindful that as of right now there's no way to edit this column (for example, on the settings page `/settings/authentication`) and most apps just come with `en` even though we also support `fr` @elitan @nunopato @plmercereau

guicurcio: FYI: This will also need an update for our back-end permissions (`select`)

elitan: You should be able to get this information from the remote project's backend, just like roles. It should work the same way, and it should be dynamic.

guicurcio: I'll also assume from the front-end that side that is a `AUTH_LOCALE_ALLOWED_LOCALES` a string of comma-separated locales (ref [https://github.com/nhost/hasura-auth/blob/main/docs/environment-variables.md#:~:text=AUTH_LOCALE_ALLOWED_LOCALES](https://github.com/nhost/hasura-auth/blob/main/docs/environment-variables.md#:~:text=AUTH_LOCALE_ALLOWED_LOCALES) @plmercereau)

elitan: It would be best to read it directly from the remote app's `auth.locales` table. Isn't that how you do it with roles? You read the data from `auth.roles` form the remote app?

If yes, do the same for `auth.locales`. If not, are there any reasons not to read roles from `auth.roles`?

szilarddoro:
Screenshot 2023-01-02 at 18 23 25

I mentioned this issue on Discord today. Once this is fixed, we can merge IMO.

dashboard/src/components/users/EditUserPasswordForm/EditUserPasswordForm.tsx
```diff
@@ -0,0 +1,156 @@
...
+          {...register('password')}
+          id="password"
+          type="password"
+          label="Password"
+          placeholder="Enter Password"
+          hideEmptyHelperText
+          error={!!errors.password}
+          helperText={errors?.password?.message}
+          fullWidth
+          autoComplete="off"
```
szilarddoro, line 123:
```suggestion
          autoComplete="off"
          autoFocus
```

dashboard/src/components/users/CreateUserForm/CreateUserForm.tsx
```diff
@@ -0,0 +1,168 @@
...
+        <Input
+          {...register('email')}
+          id="email"
+          label="Email"
+          placeholder="Enter Email"
+          hideEmptyHelperText
+          error={!!errors.email}
+          helperText={errors?.email?.message}
+          fullWidth
+          autoComplete="off"
```
szilarddoro, line 121:
```suggestion
          autoComplete="off"
          autoFocus
```

szilarddoro (approved):
LGTM

PR #1406, opened 2022-12-20
chore: deprecate axios

plmercereau:
Impacted methods:
- `nhost.functions.call`
- `nhost.graphql.request`

In both cases, their use is marked as deprecated unless there is a specific `{ useAxios: false }` last parameter.

When using `useAxios: false`:
- the only allowed option is `headers: Record<string,string>`
- the returned value matches values foreseen in the next major version:
  - `nhost.functions.call`: 
    - `error` is using the same standard error type as in `hasura-auth-js` and `hasura-storage-js`
    - `res` is using `{ status: number; statusText: string; data: T }`
  - `nhost.graphql.request`:
    - `error` is either using the standard error type, or `GraphQlError[]`


packages/nhost-js/src/utils/types.ts
```diff
@@ -73,6 +106,9 @@ export type FunctionCallResponse<T = unknown> =
     }
 
 export interface GraphqlResponse<T = object> {
-  errors?: object[]
+  errors?: GraphQLError[]
   data?: T
 }
+
+export type RestrictedFetchConfig = { headers?: Record<string, string> }
```
szilarddoro, line 113: Now that we are controlling the config, can't we just simply call it `NhostFetchConfig` or `NhostRequestConfig`?

packages/nhost-js/src/clients/functions.ts
```diff
@@ -54,6 +55,25 @@ export class NhostFunctionsClient {
     })
   }
 
+  /** @deprecated Axios will be replaced by cross-fetch in the near future. Only the headers configuration will be kept. */
+  async call<T = unknown, D = any>(
```
szilarddoro, line 59: Can't we just merge this and the other deprecated function to a single one?

For example:

```ts
async call<T = unknown, D = any>(
  url: string,
  data?: D,
  config?: (RestrictedFetchConfig | AxiosConfig) & { useAxios?: true }
): Promise<DeprecatedFunctionCallResponse<T>>
```

packages/nhost-js/src/utils/types.ts
plmercereau: yep

packages/nhost-js/src/clients/functions.ts
```diff
@@ -54,6 +55,19 @@ export class NhostFunctionsClient {
...
+    url: string,
+    data?: D,
+    config?: (NhostFetchConfig | AxiosConfig) & { useAxios?: true }
```
nunopato, line 62:
```suggestion
    config?: (NhostFunctionsConfig | AxiosConfig) & { useAxios?: true }
```

packages/nhost-js/src/clients/functions.ts
```diff
@@ -54,6 +55,19 @@ export class NhostFunctionsClient {
...
+  ): Promise<DeprecatedFunctionCallResponse<T>>
+
+  async call<T = unknown, D = any>(
+    url: string,
+    data: D,
+    config?: NhostFetchConfig & { useAxios: false }
```
nunopato, line 68:
```suggestion
    config?: NhostFunctionsConfig & { useAxios: false }
```

packages/nhost-js/src/clients/functions.ts
```diff
@@ -67,8 +81,13 @@ export class NhostFunctionsClient {
   async call<T = unknown, D = any>(
     url: string,
     data: D,
-    config?: AxiosRequestConfig
-  ): Promise<FunctionCallResponse<T>> {
+    { useAxios = true, ...config }: (AxiosConfig | NhostFetchConfig) & { useAxios?: boolean } = {}
```
nunopato, line 84:
```suggestion
    { useAxios = true, ...config }: (AxiosConfig | NhostFunctionsConfig) & { useAxios?: boolean } = {}
```

packages/nhost-js/src/clients/functions.ts
```diff
@@ -1,12 +1,13 @@
...
-
+import axios, { AxiosError, AxiosInstance, AxiosResponse, RawAxiosRequestHeaders } from 'axios'
 import { urlFromSubdomain } from '../utils/helpers'
-import { FunctionCallResponse, NhostClientConstructorParams } from '../utils/types'
+import {
+  AxiosConfig,
+  DeprecatedFunctionCallResponse,
+  FunctionCallResponse,
+  NhostClientConstructorParams,
+  NhostFetchConfig
```
nunopato, line 8:
```suggestion
  NhostFunctionsConfig
```

packages/nhost-js/src/clients/functions.ts
plmercereau: `NhostFetchConfig` is also used in `nhost.graphql.request`, I guess you want to rename it there too?
Following that logic, we should rather name the two aliases `NhostFunctionCallConfig` and `NhostGraphQlRequestConfig` as they are not parameters of the clients, but of a method of the clients

packages/nhost-js/src/clients/functions.ts
```diff
@@ -126,3 +176,6 @@ export class NhostFunctionsClient {
     return {}
   }
 }
+
+const dd = new NhostFunctionsClient({ url: 'https://functions.nhost.io' })
+dd.call('dwe')
```
nunopato, line 181:
```suggestion
```

packages/nhost-js/src/clients/functions.ts
nunopato: yes, I would also rename `NhostFetchConfig` in the `graphql` client. I can see how in the future we might have different options for functions vs graphql.

nunopato: > Following that logic, we should rather name the two aliases NhostFunctionCallConfig and

I am fine with that. In the same logic, `FunctionCallResponse` could become `NhostFunctionCallResponse`.

packages/nhost-js/src/utils/types.ts
```diff
@@ -53,6 +60,17 @@ export interface NhostClientConstructorParams
 }
 
 export type GraphqlRequestResponse<T = unknown> =
```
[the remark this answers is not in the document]
plmercereau: I moved the functions and graphql types in two distinct files and applied the above suggestions, except I did not prefix everything with `Nhost` to keep types consistent with our other client types

.changeset/kind-cycles-explain.md
```diff
@@ -0,0 +1,25 @@
+---
+'@nhost/hasura-auth-js': minor
+'@nhost/nhost-js': minor
+---
+
+Deprecate Axios
+
+Axios will be replaced by cross-fetch in the near future.
```
nunopato, line 8: We are changing t
```suggestion
Axios will be replaced by cross-fetch in the near future. The signature of 2 methods is changing because we are making the API more flexible. 
```

.changeset/kind-cycles-explain.md
```diff
@@ -0,0 +1,25 @@
...
+
+To prepare the transition, we recommend to adapt the following methods:
```
nunopato, line 10:
```suggestion
To prepare for the transition, we are deprecating the old signature for the following methods:
```

.changeset/kind-cycles-explain.md
```diff
@@ -0,0 +1,25 @@
...
+
+- `nhost.functions.call()`
+- `nhost.graphql.request()`
+
+Both methods are now marked as deprecated unless there is a specific `{ useAxios: false }` last parameter.
```
nunopato, line 15:
```suggestion
Both methods now accept an optional `useAxios` parameter that can be used to opt-in (`{ useAxios: false }`) to the new method signature. By default, `useAxios` is set to `true` so you can update it on your own time.
```

PR #1414, opened 2022-12-20
feat(dashboard): Permission Editor

### LINKED ISSUE #1427: Permission Editor
Implement permissions in the Dashboard.

szilarddoro:
**Preview URL:** https://dashboard-staging-git-feat-permission-editor-nhost.vercel.app/

Resolves #1427


dashboard/src/components/ui/v2/icons/FullPermissionIcon/FullPermissionIcon.tsx
```diff
@@ -0,0 +1,24 @@
+import type { IconProps } from '@/ui/v2/icons';
+import SvgIcon from '@mui/material/SvgIcon';
+
+function FullPermissionIcon({ sx, ...props }: IconProps) {
+  return (
+    <SvgIcon
+      width="20"
```
guicurcio, line 7: I feel like this is too small for a 1920x1080 display (while this is correct, and I have left a comment in our designs I wonder if we should just bump it) on the other hand, it looks good in smaller resolutions

dashboard/src/components/dataBrowser/EditPermissionsForm/EditPermissionsForm.tsx
```diff
@@ -0,0 +1,342 @@
...
+      </div>
+    );
+  }
+
+  if (rolesError) {
+    throw rolesError;
+  }
+
+  const availableRoles = [
+    'public',
```
guicurcio, line 115: This threw me off a bit... while hasura-auth sets the `HASURA_GRAPHQL_UNAUTHORIZED_ROLE` we initially do not create a role for it in hasura; I think we should show it both in the environment variables page and in the roles and permissions section (we have `user` as default but this is the default for authenticated users and no mention of the `public` role) this also could get stale if we allow users to change the unauthorized role given that we're relying in just a string (e.g. we don't have a column for this setting, and I'm not too sure if users could override it)

dashboard/src/components/ui/v2/Switch/Switch.stories.tsx
```diff
@@ -11,7 +11,7 @@ export default {
 const Template: ComponentStory<typeof Switch> = function Template(
   args: SwitchProps,
 ) {
-  return <Switch {...args} />;
+  return <Switch label="Accept Rules" {...args} />;
```
guicurcio, line 14: Shouldn't `label="Accept Rules"` be passed as `args` as this is an abstract component?

dashboard/src/components/ui/v2/Switch/Switch.tsx
```diff
@@ -1,13 +1,39 @@
...
 import { forwardRef } from 'react';
 
-export interface SwitchProps extends SwitchUnstyledProps {}
+export interface SwitchProps extends SwitchUnstyledProps {
+  /**
+   * Label to be displayed next to the checkbox.
+   */
+  label?: FormControlLabelProps['label'];
+  /**
+   * Props to be passed to the internal components.
```
guicurcio, line 17: We have two different comments for similar props:

`Props to be passed to individual component slots.` for [Radio.tsx](https://github.com/nhost/nhost/pull/1414/files#diff-cab2e6a6d80963fb89d93b52ac228ab8f2e2d9d60d3d2567547da562e922ac63) and `Props to be passed to the internal components.` here. Should we standarize it, or is there a reason for the difference here?

dashboard/src/components/dataBrowser/RuleGroupEditor/useRuleGroupEditor.ts
```diff
@@ -0,0 +1,19 @@
...
+}>({
+  schema: '',
+  table: '',
+});
+
+export default function useRuleGroupEditor() {
+  const context = useContext(RuleGroupEditorContext);
+
+  if (!context) {
+    throw new Error('useRuleGroupEditor must be used within a RuleGroupEditor');
```
guicurcio, line 15: Any reason to not name `RuleGroupEditorProvider` instead of `RuleGroupEditor` ?

elitan (approved):
It works as intended. I didn't look at the code.

🚀

guicurcio:
Since `X-Hasura-User-Id`

 is a `uuid` should we actually show this dropdown?   
  
FYI: I also left a few comments on Figma

dashboard/src/components/ui/v2/Switch/Switch.tsx
szilarddoro: Yep, we should standardize this. I'm leaning towards "individual component slots" or "each slot inside the XYZ component". The latter is being used by MUI.

dashboard/src/components/ui/v2/Switch/Switch.stories.tsx
szilarddoro: This is just a default label for the `Switch` that's shared between stories. Users can override it through the storybook.

dashboard/src/components/dataBrowser/EditPermissionsForm/EditPermissionsForm.tsx
szilarddoro: I thought that as an addition to the `admin` role, we have this by default set up in the background "automagically". @elitan would you mind sharing your opinion on this, please?

dashboard/src/components/ui/v2/icons/FullPermissionIcon/FullPermissionIcon.tsx
szilarddoro: The large version looks strange to me. Let's wait for @eugene-reuka 's response on this one.

dashboard/src/components/dataBrowser/RuleGroupEditor/useRuleGroupEditor.ts
szilarddoro: I don't think it makes sense because we are only allowing this hook to be accessed inside the boundaries of a `RuleGroupEditor`. We don't have a `RuleGroupEditorProvider`.

dashboard/src/components/ui/v2/icons/FullPermissionIcon/FullPermissionIcon.tsx
szilarddoro: I also think that this is a relatively easy task. I wouldn't necessarily block the release by waiting for this to be resolved.

elitan:
> This is a very good point. Unfortunately, this version doesn't support read-only mode yet. I wonder if we should support it now or create a follow-up PR for this. @elitan @nunopato what are your thoughts?

Good point indeed. Let's do it as a follow-up PR. Please create an issue or project so we don't forget.

PR #1444, opened 2022-12-27
Fix: convert non ISO-8859-1 file names

### LINKED ISSUE #822: React - Upload error when filename contain utf8 chars
Hi,

When i try to upload in react with the `useMultipleFilesUpload` (it will be the same with `useFileUpload`) the request isn't sent to the storage microservice and this error appear in the Interpretor event: `"Failed to execute 'setRequestHeader' on 'XMLHttpRequest': String contains non ISO-8859-1 code point."`

Example with special char: https://postimg.cc/SjvWZBCB
Example with normal filename: https://postimg.cc/Sjs6fS2L

It's seems related to axios, here the corresponding issue: https://github.com/axios/axios/issues/4556 (not resolved)

plmercereau:
Closes #822


elitan:
Can we support UTF-8 instead of ISO-8859-1?

plmercereau:
Maybe we can, but I couldn't. It is probably related to axios.
As we can't use special characters right now and this PR solves it transforming them, I would keep this PR, and once we moved away from Axios, I would try to keep any UTF character as is in a second PR

elitan:
If it's related to Axios, this PR (https://github.com/nhost/nhost/pull/1351) should fix it and I'd assume this current PR is not needed?

plmercereau:
> If it's related to Axios, #1351 should fix it and I'd assume this current PR is not needed?
#1351 works, but we will probably need to check the headers to make sure the names are kept in the exact same format.

it depends on how much time we want to move away from Axios. For instance, If we wait for three more months, I think it's worth fixing a bug for this period.

elitan:
Make sense. Is the `toIso88591` code only used for Axios and not cross-fetch?

plmercereau:
> Make sense. Is the `toIso88591` code only used for Axios and not cross-fetch?

yes

elitan (approved):

PR #1450, opened 2022-12-31
[chatgpt] added tests

elitan:


packages/hasura-auth-js/tests/utils.test.ts
```diff
@@ -1,82 +1,266 @@
...
+  })
+
+  test('should return the value of the specified parameter after the hash', () => {
+    expect(getParameterByName('param', 'http://example.com#param=value')).toBe('value')
+    expect(getParameterByName('param', 'http://example.com#other=other&param=value')).toBe('value')
+    expect(getParameterByName('param', 'http://example.com#param=value&other=other')).toBe('value')
+  })
 })
+
+// suite('removeParameterFromWindow', () => {
```
plmercereau, line 202: We should either explain why this part is commented, or remove it

plmercereau (changes requested):

packages/hasura-auth-js/tests/utils.test.ts
elitan: totally

elitan: got it working now with jsdom

plmercereau (approved):

PR #1452, opened 2023-01-02
fix: improve missing React provider error

### LINKED ISSUE #1391: Better error message when `<NhostProvider />` is absent.
**Describe the bug**
We're showing a non-helpful error message when the `<NhostProvier />` is not added correctly.

**To Reproduce**

Use our react/nextjs SDK without any `<NhostProvider />`.

Example:

CleanShot 2022-12-16 at 15 04 16@2x

I think we should return a helpful error message saying that the `<NhostProvier />` is not configured correctly.

plmercereau:
Closes #1391

### DIFF
--- a/.changeset/proud-hotels-pretend.md
+++ b/.changeset/proud-hotels-pretend.md
@@ -1,0 +1,5 @@
+---
+'@nhost/react': patch
+---
+
+Improve the error message when the application is not wrapped in `<NhostProvider></NhostProvider>`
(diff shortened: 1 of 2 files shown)


szilarddoro (approved):

plmercereau:
@elitan please let me know if it solves your initial problem: I did not get the same error as you.

szilarddoro:
Johan's error seems to be related to React missing as a dependency or something like that as `createContext` is only called inside `packages/react/src/provider.tsx`.

elitan (approved):

PR #1461, opened 2023-01-03
feat: image transformation parameters

plmercereau:
It is now possible to pass on image transformation parameters in `nhost.storage.getPublicUrl()`.
Available parameters:

- height
- width
- blur
- quality

For instance:

```ts
const url = nhost.storage.getPublicUrl({
  fileId: 'cd8eaca3-30a9-460e-b4d7-b4b7afc759c1',
  width: 800,
  blur: 20
})
```

missing part: test on the `quality` parameter, that doesn't seem to work correctly in hasura-storage


elitan (approved):
> missing part: test on the quality parameter, that doesn't seem to work correctly in hasura-storage

If we haven't created an issue in Hasura Storage yet, we should do so.

PR #1467, opened 2023-01-04
feat: add support for custom local subdomains

szilarddoro:


packages/nhost-js/src/utils/helpers.ts
```diff
@@ -1,7 +1,8 @@
 import { NhostClientConstructorParams } from './types'
 
 // a port can be a number or a placeholder string with leading and trailing double underscores, f.e. "8080" or "__PLACEHOLDER_NAME__"
-const LOCALHOST_REGEX = /^((?<protocol>http[s]?):\/\/)?(?<host>localhost)(:(?<port>(\d+|__\w+__)))?$/
+const LOCALHOST_REGEX =
```
szilarddoro, line 4: TBH, I'm not sure if my implementation makes sense as is.

Some context for you @plmercereau:

Now that `https://localdev.nhost.run` will be used by the local dashboard, I had to change these things here to prevent the error caused by the region check on line 51.

I didn't want to include `.nhost.run` in the subdomain configuration of the dashboard, because that's not how we would initialize an `NhostClient` in a project that's not our dashboard. I thought it makes more sense to just specify it as `localdev`, but then the RegExp looks a bit weird to me. I'd appreciate your thoughts on this.

plmercereau: > I didn't want to include `.nhost.run` in the subdomain configuration of the dashboard, because that's not how we would initialize an `NhostClient` in a project that's not our dashboard. I thought it makes more sense to just specify it as `localdev`

I don't understand this part

packages/nhost-js/src/utils/helpers.ts
```diff
@@ -25,7 +26,7 @@ export function urlFromSubdomain(
     throw new Error('Either `backendUrl` or `subdomain` must be set.')
   }
 
-  // check if subdomain is [http[s]://]localhost[:port]
+  // check if subdomain is [http[s]://]localhost[:port] or [http[s]://]localdev[:port]
   const subdomainLocalhostFound = subdomain.match(LOCALHOST_REGEX)
   if (subdomainLocalhostFound?.groups) {
     const { protocol = 'http', host, port = 1337 } = subdomainLocalhostFound.groups
```
plmercereau, line 32:
```suggestion
    const { protocol = 'http', host, port } = subdomainLocalhostFound.groups
```

goes with the following suggestion

packages/nhost-js/src/utils/helpers.ts
```diff
@@ -34,7 +35,17 @@ export function urlFromSubdomain(
...
       return urlFromEnv
     }
-    return `${protocol}://${host}:${port}/v1/${service}`
+
+    if (host === 'localhost') {
+      return `${protocol}://localhost:${port}/v1/${service}`
+    }
+
+    // we omit the port if it's the default port
+    if (port !== 1337) {
```
plmercereau, line 44:
```suggestion
    if (port) {
```

packages/nhost-js/src/utils/helpers.ts
szilarddoro: For `localhost`, we don't need the `.nhost.run` ending, but we do need it in every other case. Subdomain is passed to `NhostClient` as a simple subdomain without the `.nhost.run`. I just didn't want to make an exception here to pass it like `subdomain: 'localdev.nhost.run'` so I need additional checks inside the SDK and I need to add it manually for localdev.

plmercereau: Thanks, I got it.
One last thing that I don't get: why don't we need a port when using `localdev.nhost.run`? How traefik listen to ports 80/443?

dashboard/src/utils/env.ts
```diff
@@ -7,17 +7,29 @@ export const LOCAL_MIGRATIONS_URL = `http://localhost:${
...
+ * Subdomain of the Nhost project.
+ */
+export const SUBDOMAIN =
+  process.env.NEXT_PUBLIC_NHOST_LOCAL_SUBDOMAIN || 'localhost';
+
+/**
+ * Subdomain of the Nhost project with protocol.
+ */
+export const SUBDOMAIN_WITH_PROTOCOL =
+  SUBDOMAIN === 'localhost' ? SUBDOMAIN : `https://${SUBDOMAIN}`;
```
elephant3, line 25: it's confusing a bit, if it's subdomain with protocol, then maybe it should be `http://localhost` in case subdomain equals to `localhost`?

packages/nhost-js/src/utils/helpers.ts
elephant3: The former port `1337` becomes `443` and yet it will be configurable (in order to run multiple nhost apps)

elephant3 (commented):
I'll build a docker image and play with it locally in CLI. Will let you know how it works

dashboard/README.md
```diff
@@ -47,18 +47,22 @@ pnpm storybook
...
+| `NEXT_PUBLIC_NHOST_PLATFORM`                  | This should be set to `false` to connect the Nhost Dashboard to a locally running Nhost backend. Setting this to `true` turns off local development.                                            |
+| `NEXT_PUBLIC_NHOST_LOCAL_HASURA_PORT`         | This environment variable has been deprecated. Please use `NEXT_PUBLIC_NHOST_LOCAL_HASURA_CONSOLE_PORT` instead.                                                                                |
+| `NEXT_PUBLIC_NHOST_LOCAL_BACKEND_PORT`        | This environment variable has been deprecated. Please use `NEXT_PUBLIC_NHOST_LOCAL_GRAPHQL_PORT` instead.                                                                                       |
+| `NEXT_PUBLIC_NHOST_LOCAL_MIGRATIONS_PORT`     | This port should be configured when running `nhost [up/dev]` with the `--console-api-port <port` option. Used only if local development is enabled. Default: `9693`                             |
```
nunopato, line 56:
```suggestion
| `NEXT_PUBLIC_NHOST_LOCAL_MIGRATIONS_PORT`     | This port should be configured when running `nhost [up/dev]` with the `--console-api-port <port>` option. Used only if local development is enabled. Default: `9693`                             |
```

dashboard/Dockerfile
```diff
@@ -18,11 +18,17 @@ WORKDIR /app
 ENV NEXT_TELEMETRY_DISABLED 1
 ENV NEXT_PUBLIC_ENV dev
 ENV NEXT_PUBLIC_NHOST_PLATFORM false
+ENV NEXT_PUBLIC_NHOST_LOCAL_SUBDOMAIN localdev
```
nunopato, line 21: we decided to go with `local` instead of  `localdev`, if I remember correctly.

dashboard/docker-entrypoint.sh
```diff
@@ -3,13 +3,21 @@
 set -e
 
 # read ports from env variables or use defaults
-NEXT_PUBLIC_NHOST_MIGRATIONS_PORT="${NEXT_PUBLIC_NHOST_MIGRATIONS_PORT:=9693}"
-NEXT_PUBLIC_NHOST_HASURA_PORT="${NEXT_PUBLIC_NHOST_HASURA_PORT:=9695}"
-NEXT_PUBLIC_NHOST_LOCAL_BACKEND_PORT="${NEXT_PUBLIC_NHOST_LOCAL_BACKEND_PORT:=1337}"
+NEXT_PUBLIC_NHOST_LOCAL_SUBDOMAIN="${NEXT_PUBLIC_NHOST_LOCAL_SUBDOMAIN:=localdev}"
```
nunopato, line 6: same here.

dashboard/src/hooks/useCurrentWorkspaceAndApplication.tsx
```diff
@@ -41,7 +42,7 @@ export function useCurrentWorkspaceAndApplication(): UseCurrentWorkspaceAndAppli
           },
         ],
         deployments: [],
-        subdomain: 'localhost',
+        subdomain: 'local',
```
nunopato, line 45: I am not sure why here we're setting the subdomain to `local` but to `localhost` here https://github.com/nhost/nhost/pull/1467/files#diff-1151af892b0f19ccb8e4e69593896b1dfcf8fd9a400a87e9d1bf0e53adb7978cR41.

packages/nhost-js/src/utils/helpers.ts
```diff
@@ -25,16 +27,30 @@ export function urlFromSubdomain(
...
+
+    if (host === 'localhost') {
+      console.warn(
+        'The `subdomain` is set to "localhost". Support for this will be removed in a future release. Please use "local" instead.'
+      )
+
+      return `${protocol || 'http'}://localhost:${port || 1337}/v1/${service}`
+    }
+
+    // TODO: Change `localdev` to `local`
```
nunopato, line 48: what is the reason for not changing this already?

szilarddoro: I mentioned this at one of our meetings. For now, I wanted to keep it `localdev`, so it can be tested with the existing CLI. I'll change it to `local` everywhere eventually.

nunopato (approved):

elephant3 (approved):

PR #1470, opened 2023-01-04
fix(dashboard): Resetting the cache when signing out.

guicurcio:
# fix(dashboard): Resetting the cache when signing out.

Right now, when signing out you get an error regarding a missing field in our cache `workspace_members` this is because we sign out the user with `nhost.auth.signOut();` prior to redirecting and when we query again for the workspace the user has, it comes out as null (as there is no user) and then redirect.

### DIFF
--- a/dashboard/src/components/dashboard/AccountMenu.tsx
+++ b/dashboard/src/components/dashboard/AccountMenu.tsx
@@ -1,13 +1,11 @@
 import { ChangePasswordModal } from '@/components/applications/ChangePasswordModal';
-import { useWorkspaceContext } from '@/context/workspace-context';
-import { useUserDataContext } from '@/context/workspace1-context';
 import { Avatar } from '@/ui/Avatar';
 import { Modal } from '@/ui/Modal';
 import Button from '@/ui/v2/Button';
 import { Dropdown, useDropdown } from '@/ui/v2/Dropdown';
 import Text from '@/ui/v2/Text';
-import { emptyWorkspace } from '@/utils/helpers';
 import { nhost } from '@/utils/nhost';
+import { useApolloClient } from '@apollo/client';
 import { useUserData } from '@nhost/nextjs';
 import Image from 'next/image';
 import { useRouter } from 'next/router';
@@ -22,9 +20,8 @@ function AccountMenuContent({
 }: AccountMenuContentProps) {
   const user = useUserData();
   const router = useRouter();
+  const client = useApolloClient();
   const [clicked, setClicked] = useState(false);
-  const { setWorkspaceContext } = useWorkspaceContext();
-  const { setUserContext } = useUserDataContext();
   const { handleClose } = useDropdown();
 
   return (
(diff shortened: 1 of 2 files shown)


dashboard/src/components/dashboard/AccountMenu.tsx
```diff
@@ -23,21 +21,19 @@ function AccountMenuContent({
...
+  const client = useApolloClient();
 
   return (
-    <div className="relative grid w-account grid-flow-row gap-5 p-6">
+    <div className="relative grid grid-flow-row gap-5 p-6 w-account">
       <Button
         variant="borderless"
         color="secondary"
-        className="absolute top-6 right-4 grid grid-flow-col items-center gap-1 self-start font-medium"
+        className="absolute grid items-center self-start grid-flow-col gap-1 font-medium top-6 right-4"
```
szilarddoro, line 32: What tool are you using for this?

dashboard/src/components/dashboard/AccountMenu.tsx
```diff
@@ -23,21 +21,19 @@ function AccountMenuContent({
...
         onClick={async () => {
-          setWorkspaceContext(emptyWorkspace());
-          setUserContext({ workspaces: [] });
+          await router.push('/signin');
```
szilarddoro, line 34: I think we shouldn't change the order of the sign-out and the router.push, because the sign-in page would try to redirect us to the dashboard's home page when it sees that we are signed in.

dashboard/src/components/dashboard/AccountMenu.tsx
guicurcio: I have the `Headwind` VS code extension, which I think is sorting this classes (in a kind of arbitrary way? none are alphabetically) I did think that `prettier-plugin-tailwindcss` overrided this in our `precommit`

szilarddoro: I have removed Headwind as it conflicted with `prettier-plugin-tailwindcss`. I think we can roll with the Prettier plugin only, because that's IDE agnostic.

guicurcio: :+1: I have reseted this branch 2 commits prior and force pushed so we don't see this changes (also disabled the Headwind extension)

dashboard/src/components/dashboard/AccountMenu.tsx
guicurcio: OK after testing the re-order:  
  
This is what works the best and doesn't generate a flicker when signing out nor the cache errors:  
  
\`\`\`ts  
router.push('/signin');

**await** nhost.auth.signOut();

**await** client.resetStore();  
\`\`\`

This one generates the `workspace_member` field not found:

\`\`\`ts  
router.push('/signin');

**await** nhost.auth.signOut();

**await** client.resetStore();  
\`\`\`

This one `expecting a value for non-nullable variable: appId`:

```typescript
await client.resetStore();
nhost.auth.signOut();
await router.push('/signin');
```

szilarddoro: The first and second are looking identical to me, but thanks for the investigation. I see what you mean. Go with the one with the best result!

PR #1471, opened 2023-01-04
fix(dashboard): Refactor workspaces creation, fix 404 redirection (workspaces & applications)

### LINKED ISSUE #1502: Renaming project returns to 404
**To Reproduce**
1. Go to Settings -> Project Name
2. Change the project name
3. Click on "Save"
4. The user is redirected to https://app.nhost.io/404

guicurcio:
# fix(dashboard): Workspaces creation, new form, correct redirects.

Issues Fixed:
- [x] Creating a workspace with a name that errors kept the loading state (even if typing again.)
- [x] Creating a workspace, redirecting to the workspace doesn't redirect to 404 (we fetch the workspace data before redirecting.)
- [x] Deleting a workspace, redirecting, and updating the workspace sidebar.

Fixed #1502.


dashboard/src/components/home/EditJwtSecretForm/EditJwtSecretForm.tsx
```diff
@@ -0,0 +1,164 @@
...
+      try {
+        JSON.parse(value);
+        return true;
+      } catch (error) {
+        return false;
+      }
+    }),
+});
+
+export default function EditJwtSecretForm({
```
szilarddoro, line 63: Hmm, how come this form is added?

dashboard/src/components/home/CreateWorkspaceForm/CreateWorkspaceForm.tsx
```diff
@@ -0,0 +1,170 @@
...
+    resolver: yupResolver(validationSchema),
+  });
+
+  const {
+    register,
+    formState: { dirtyFields, isSubmitting, errors },
+  } = form;
+  const isDirty = Object.keys(dirtyFields).length > 0;
+
+  const currentUser = nhost.auth.getUser();
```
szilarddoro, line 84: You can make use of our `useUserData()` hook.

dashboard/src/components/home/EditJwtSecretForm/EditJwtSecretForm.tsx
szilarddoro: This I think is the same one we use on the Settings page.

dashboard/src/components/workspace/WorkspaceSection.tsx
```diff
@@ -15,7 +16,23 @@ export function WorkspaceSection() {
...
+
+                <Text variant="subtitle1" component="span">
+                  Invite team members to workspaces to work collaboratively.
+                </Text>
+              </span>
+            ),
+            // payload: {
+            //   disabled: true,
+            //   jwtSecret: data?.app?.hasuraGraphqlJwtSecret,
+            // },
```
szilarddoro, line 33:
```suggestion
```

szilarddoro (changes requested):

szilarddoro:
So good to see that we are using the same approach in more and more places 😎

dashboard/src/components/home/EditJwtSecretForm/EditJwtSecretForm.tsx
guicurcio: My bad... this one was the one I based the new form out of

dashboard/src/components/home/CreateWorkspaceForm/CreateWorkspaceForm.tsx
guicurcio: Just changed it, but as a note for us: 

`useUserData()` returns `any` while `nhost.auth.getUser();` returns a `User`

szilarddoro: Hmm, the type seems to be lost somewhere along the way, because in `packages/react/src/useUserData.ts` I see that the function returns a `User`. 🤔 

Do you have any ideas @plmercereau?

This is what I've found: `export declare const useUserData: () => import("@nhost/hasura-auth-js").User | null;`

Maybe we need to explicitly cast the return value to `User` to import it from `@nhost/nhost-js` instead of `@nhost/hasura-auth-js` which is not a dependency of `@nhost/react`?

dashboard/src/components/home/CreateWorkspaceForm/CreateWorkspaceForm.tsx
```diff
@@ -0,0 +1,189 @@
...
+        if (slug.length < 4 || slug.length > 32) {
+          return false;
+        }
+
+        return true;
+      },
+    ),
+});
+
+export default function CreateWorkspaceForm({
```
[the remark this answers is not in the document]
guicurcio: good point, this way it's similar also to the JWT edit. I'll add this :+1:

dashboard/src/components/home/CreateWorkspaceForm/CreateWorkspaceForm.tsx
guicurcio: Created an issue out of the comments: #1484

dashboard/src/components/workspace/WorkspaceHeader.tsx
```diff
@@ -112,9 +106,23 @@ export default function WorkspaceHeader() {
...
-                  onClick={() =>
-                    setChangeWorkspaceNameModal(!changeWorkspaceNameModal)
-                  }
+                  onClick={() => {
+                    openDialog('EDIT_WORKSPACE_NAME', {
+                      title: (
+                        <span className="grid grid-flow-row">
+                          <span>Change Workspace Name</span>
+                          <Text variant="subtitle1" component="span">
+                            Changing the workspace name will also affect the URL
```
guicurcio, line 115: Check copy & double-check that we want to also change the slug

dashboard/src/pages/[workspaceSlug]/[appSlug]/settings/general/index.tsx
```diff
@@ -69,6 +70,10 @@ export default function SettingsGeneralPage() {
   const { register, formState } = form;
 
   const handleProjectNameChange = async (data: ProjectNameValidationSchema) => {
+    await router.push({
```
Comment (guicurcio) [inline dashboard/src/pages/[workspaceSlug]/[appSlug]/settings/general/index.tsx:73]: This is the bit that solve the `404` flashing and the problem with changing the application name redirecting to 404 (which I just found out) basically, if we are in the process of mutating we change the browser URL to contain a query like `?mutating=true` this is because as soonest as we change the slug, the apollo cache realizes that there isn't a workspace/application with that slug and redirects to `404` .  
  
Alternatives would be to rewrite to the apollo [https://www.apollographql.com/docs/react/caching/cache-interaction/#using-graphql-queries](https://www.apollographql.com/docs/react/caching/cache-interaction/#using-graphql-queries) (which would be kind-of a more correct implementation) but we have been staying away from this given that if we move away from Apollo at some point, we would only have queries/mutations (it would solve it as long as a re-fetch doesn't happen in the meantime, given that our main hooks like `currentWorkspace..` fetchs the entire user workspace/app information.)   
  
Alternatives to keeping state in the URL would be to just use context so we don't show anything in the URL even if for a few seconds. Which may not be that bad since we also have to solve a few things regarding the last workspace used (so when creating a new workspace the user can use the last one he used instead of the same every time.) Though we may have to be careful about possible side-effects. 

wdyt? @szilarddoro

Comment (szilarddoro) [reply dashboard/src/pages/[workspaceSlug]/[appSlug]/settings/general/index.tsx]: This is a good one @guicurcio. I like this approach. Maybe one thing I would change is `mutating` to `updating` or something like that.

dashboard/src/components/workspace/WorkspaceHeader.tsx
szilarddoro: I think we should change the slug if we are doing it for projects as well. But maybe @nunopato / @elitan has a better answer to this.

guicurcio: > change the slug if we are doing it for projects as well.

Yep, that makes sense. Then we are good.

szilarddoro (approved):

Issue #1336, opened 2022-12-05
`@nhost/nextjs` should store the access token in a cookie to make it available during SSR. https://github.com/nhost/nhost/pull/1482


elitan:
Right now, `@nhost/nextjs` stores the refresh token in a cookie but the access token is stored in memory. 

I suggest we let `@nhost/nextjs` also to store the access token in a cookie so it can be available in `getServerSideProps` and other places on the server.

PR #1482, opened 2023-01-06
chore: store the session in a cookie to avoid refetching the jwt on every ssr call

plmercereau:
Store the session in a cookie so there is no need for an SSR page to ask for a new access token

Closes https://github.com/nhost/nhost/issues/1336

### DIFF
--- a/packages/nextjs/src/utils.ts
+++ b/packages/nextjs/src/utils.ts
@@ -1,5 +1,6 @@
-import { NhostSession } from '@nhost/react'
+import { NhostClient, NhostSession } from '@nhost/react'
 import fetch from 'cross-fetch'
+import Cookies from 'js-cookie'
 
 export const refresh = async (nhostUrl: string, refreshToken: string): Promise<NhostSession> => {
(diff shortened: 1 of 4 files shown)


packages/nextjs/src/utils.ts
```diff
@@ -12,3 +13,21 @@ export const refresh = async (nhostUrl: string, refreshToken: string): Promise<N
   if (result.ok) return result.json()
   else return Promise.reject(result.statusText)
 }
+
+export const NHOST_SESSION_KEY = 'nhostSession'
+
+export const setNhostSessionInCookie = (param: NhostClient | NhostSession | null) => {
+  const session = param && 'auth' in param ? param.auth.getSession() : param
+  if (session) {
```
szilarddoro, line 21: Maybe it's just nitpicking but changing this to

```ts
if (!session) {
  Cookies.remove(NHOST_SESSION_KEY)
  
  return
}

const { refreshToken, ...rest } = ...
```

makes the code so much nicer.

szilarddoro (approved):

Issue #1484, opened 2023-01-06
@nhost/nextjs SDK: `useUserData()` is not typed, returns `any` while nhost.auth.getUser(); returns a `User`


guicurcio:
Hmm, the type seems to be lost somewhere along the way, because in `packages/react/src/useUserData.ts` I see that the function returns a `User`. 🤔 

Do you have any ideas @plmercereau?

This is what I've found: `export declare const useUserData: () => import("@nhost/hasura-auth-js").User | null;`

Maybe we need to explicitly cast the return value to `User` to import it from `@nhost/nhost-js` instead of `@nhost/hasura-auth-js` which is not a dependency of `@nhost/react`?

_Originally posted by @szilarddoro in https://github.com/nhost/nhost/pull/1471#discussion_r1063224146_

plmercereau:
We can explicitly set the hook return type. We would probably need to add them to all our hooks. Will create a PR

PR #1485, opened 2023-01-06
fix: 🐛 add explicit types to React hooks and Vue composables

plmercereau:
closes https://github.com/nhost/nhost/issues/1484

### DIFF
--- a/.changeset/rare-eyes-jump.md
+++ b/.changeset/rare-eyes-jump.md
@@ -1,0 +1,5 @@
+---
+'@nhost/vue': patch
+---
+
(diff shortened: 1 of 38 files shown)


guicurcio (approved):
react stuff looks :100: (will download a branch and rebuild deps + leave a comment if anything is amiss) thanks pilou

szilarddoro:
Shouldn't we patch `@nhost/nextjs` as well?

plmercereau:
> Shouldn't we patch `@nhost/nextjs` as well?

The packages are linked, but I'm not sure it would add the changeset in the change log. I just updated the changeset

PR #1503, opened 2023-01-11
feat(dashboard): Retrigger Deployments

szilarddoro:
Resolves #1457. There is [a backend change](https://github.com/nhost/console-next/pull/1544) that adds the support for inserting deployments into the `deployments` table.

I unified the component we are using on the Overview and the Deployments page.

PR #1522, opened 2023-01-16
fix(dashboard): correct redeployment button

szilarddoro:
This is a follow-up PR for https://github.com/nhost/nhost/pull/1503. Also, fixes an issue where the "Live" badge was shown for every deployment that was the first on the page on the "Deployments" page.

Preview: https://dashboard-staging-git-fix-retrigger-deployment-status-nhost.vercel.app/

### DIFF
--- a/dashboard/src/components/ui/StatusCircle.tsx
+++ b/dashboard/src/components/ui/StatusCircle.tsx
@@ -1,9 +1,11 @@
-import clsx from 'clsx';
+import { twMerge } from 'tailwind-merge';
 
 export type DeploymentStatus =
   | 'DEPLOYING'
   | 'DEPLOYED'
   | 'FAILED'
+  | 'PENDING'
+  | 'SCHEDULED'
   | undefined
   | null;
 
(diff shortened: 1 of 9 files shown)


dashboard/src/components/ui/StatusCircle.tsx
```diff
@@ -1,9 +1,11 @@
-import clsx from 'clsx';
+import { twMerge } from 'tailwind-merge';
 
 export type DeploymentStatus =
   | 'DEPLOYING'
   | 'DEPLOYED'
   | 'FAILED'
+  | 'PENDING'
```
guicurcio, line 7: These three states match the `deploymentStatus` column in the `deployments` table:

<table><tbody><tr><td>DEPLOYING</td></tr><tr><td>DEPLOYED</td></tr><tr><td>FAILED</td></tr></tbody></table>

do the new ones 'PENDING' & 'SCHEDULED' are also new states in this column? @elephant3 I haven't seen them in staging, do we mutate a given deployment to go from, say, PENDING -> FAILED? I think we could also look up into creating these states as an enum-like similar to what we have with the `app_states`

elephant3: Yes, they are - https://github.com/nhost/watchtower/blob/ce804295c13a4d0067e20a941123ae2de77fb1ce/pkg/hasura/deployments_notificator.go#L18-L23

`PENDING` -> `FAILED` is possible, as well as `SCHEDULED` -> `FAILED` if they get stuck

guicurcio: Ah, got it. Thanks for clarifying. I'll open up an issue in console-next in the event we want to have something like the \`**app\_states\`** table to get a grasp for all possible states of a deployment :+1:

guicurcio (approved):

elitan (approved):

dashboard/src/components/ui/StatusCircle.tsx
7koma95: > Ah, got it. Thanks for clarifying. I'll open up an issue in console-next in the event we want to have something like the \`**app\_states\`** table to get a grasp for all possible states of a deployment :+1:

7koma95 (approved):

PR #1546, opened 2023-01-28
fix(dashboard): allowed roles

elitan:
The previous Roles section is renamed to Allowed Roles which is different from roles in general. Allowed roles are a subset of roles.


szilarddoro (changes requested):
Please create a changeset by running `pnpm changeset` in the root folder.

szilarddoro (approved):

PR #1547, opened 2023-01-28
fix(logs): fixed duplicated logs during live mode

elitan:
- Fixed a bug where live logs would show duplicated logs when the live mode was active.
- "To" now shows the current time (updates every second) when the live mode is active.


dashboard/src/components/logs/LogsHeader/LogsHeader.tsx
```diff
@@ -73,8 +75,23 @@ export default function LogsHeader({
     }
 
     onToDateChange(null);
+    setCurrentTime(new Date());
   }
 
+  useEffect(() => {
```
szilarddoro, line 81: I'd wrap the `LogsDatePicker` and the "Live" buttons together in a component that would be responsible for updating the time every second. I don't think we should re-render the whole header every second.

Basically lines 110-145 would become a new component with the state management included.

szilarddoro (changes requested):

szilarddoro:
Please create a changeset

szilarddoro:
I noticed that the auto-updating timer is changing its width when the live mode is enabled. We should probably set some fixed width there which is just slightly wider than the current one.

dashboard/src/pages/[workspaceSlug]/[appSlug]/logs/index.tsx
```diff
@@ -25,6 +25,7 @@ export default function LogsPage() {
     AvailableLogsServices.ALL,
   );
 
+  // create a client that sends http requests to Hausra but websocket requests to Bragi
```
Comment (guicurcio) [inline dashboard/src/pages/[workspaceSlug]/[appSlug]/logs/index.tsx:28]:
```suggestion
  // create a client that sends http requests to Hasura but websocket requests to Bragi
```

There are some comments in the function itself e.g. https://github.com/nhost/nhost/blob/eb3ba21afca6690ea173eba54d550f59d6a00e0c/dashboard/src/hooks/useRemoteApplicationGQLClientWithSubscriptions.tsx#L8-L13

though this appear on hover in an IDE

dashboard/src/pages/[workspaceSlug]/[appSlug]/logs/index.tsx
```diff
@@ -56,33 +57,41 @@ export default function LogsPage() {
...
+                log.service === prevLogs[0].service,
+            );
+
+          if (prevAndNewLogsAreTheSame) {
+            return prev;
           }
 
-          const newLogs = subscriptionData.data.logs;
-
+          // if the logs are not the same, it means we got new logs. We'll merge the new logs with the existing logs.
```
Comment (guicurcio) [inline dashboard/src/pages/[workspaceSlug]/[appSlug]/logs/index.tsx:92]:
```suggestion
          // if the logs are not the same, it means we've got new logs. We'll merge the new logs with the existing logs.
```

Also we've this format for multiline comments at most other files:

```
 /**
   * 
   * 
   */
```
which I prefer but I don't quite know if it makes sense for these kinds of lines (as opposed to functions, hooks, etc...)

dashboard/src/pages/[workspaceSlug]/[appSlug]/logs/index.tsx
Comment (szilarddoro) [reply dashboard/src/pages/[workspaceSlug]/[appSlug]/logs/index.tsx:28]: this is more like an inline comment on a variable, so it's fine IMO

dashboard/src/pages/[workspaceSlug]/[appSlug]/logs/index.tsx
Comment (szilarddoro) [reply dashboard/src/pages/[workspaceSlug]/[appSlug]/logs/index.tsx:92]: in this case, I think `//` makes sense. For hooks, etc. it makes sense to use the `/** ... */` syntax because the comment is then used by the IDE as you mentioned before

szilarddoro (approved):
Once Guido's suggestions are applied to the code, we're good to merge IMO

7koma95 (approved):

7koma95 (approved):

PR #1585, opened 2023-02-06
exponential backoff

plmercereau:
Set limits to refreshing the token on error

 When starting, the client was trying to refresh the token five times every second, then indefinitely every five seconds.
 It is now limited to 5 attempts at the following intervals: 5, 10, 20, 40, and 80 seconds. If all these attempts fail, the user state is signed out.

 Similarly, when refreshing the token failed, the client attempted to refresh the token every second.
 It is now limited to 5 attempts at the following intervals: 5, 10, 20, 40, and 80 seconds.

### DIFF
--- a/.changeset/old-wasps-yell.md
+++ b/.changeset/old-wasps-yell.md
@@ -1,0 +1,11 @@
+---
+'@nhost/hasura-auth-js': patch
+---
+
+Set limits to refreshing the token on error
+
+When starting, the client was trying to refresh the token five times every second, then indefinitely every five seconds.
+It is now limited to 5 attempts at the following intervals: 1, 2, 4, 8, and 16 seconds. If all these attempts fail, the user state is signed out.
+
+Similarly, when refreshing the token failed, the client was attempting to refresh the token every second.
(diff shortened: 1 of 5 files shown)


.changeset/old-wasps-yell.md
```diff
@@ -0,0 +1,11 @@
...
+'@nhost/hasura-auth-js': patch
+---
+
+Set limits to refreshing the token on error
+
+When starting, the client was trying to refresh the token five times every second, then indefinitely every five seconds.
+It is now limited to 5 attempts at the following intervals: 1, 2, 4, 8, and 16 seconds. If all these attempts fail, the user state is signed out.
+
+Similarly, when refreshing the token failed, the client was attempting to refresh the token every second.
+It is now limited to 5 attempts at the following intervals: 1, 2, 4, 8, and 16 seconds.
```
dbarrosop, line 11: I'd suggest starting in 5. This feels a bit too aggressive.

szilarddoro: I agree

szilarddoro (changes requested):

elitan (approved):

szilarddoro (approved):

PR #1609, opened 2023-02-11
1575 make it possible to manage roles

elitan:
## Default Allowed Roles
- [ ] Use Select/Option elements instead of text input. Only list available roles in the Select list.


elitan:
Will think of a better way to represent this in a single settings box.

PR #1617, opened 2023-02-15
Build types error different versions of NHost

fabioespinosa:
By making this change I have no issues with types using documentation here:

https://docs.nhost.io/reference/react/apollo

Otherwise following that same example and running yarn build to check types will throw an error.

```
Type error: Type 'import("/node_modules/@nhost/react/dist/client").NhostClient' is not assignable to type 'import("/node_modules/@nhost/nhost-js/dist/core/nhost-client").NhostClient'.
  Types of property 'auth' are incompatible.
    Type 'HasuraAuthClient' is missing the following properties from type 'HasuraAuthClient': api, onTokenChangedFunctions, onAuthChangedFunctions, refreshInterval, and 19 more.

  34 |     <NhostProvider nhost={nhost}>
  35 |       <NhostApolloProvider
> 36 |         nhost={nhost}
     |         ^
```

### DIFF
--- a/integrations/apollo/src/index.ts
+++ b/integrations/apollo/src/index.ts
@@ -11,7 +11,7 @@ import {
 import { setContext } from '@apollo/client/link/context'
 import { GraphQLWsLink } from '@apollo/client/link/subscriptions'
 import { getMainDefinition } from '@apollo/client/utilities'
-import { NhostClient } from '@nhost/nhost-js'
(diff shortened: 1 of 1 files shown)


fabioespinosa:
Nevermind, it seems I had to just delete node_modules and try again. Now working, but not sure what fixed it

szilarddoro:
It's highly likely that there was a version mismatch between the two packages.

Issue #1623, opened 2023-02-17
nhost/react module not loading in ReactNative 0.7 


lachu97:
I use "@nhost/react": "^1.13.4", library for my New React Native Project(version:0.71.3) while the android build is successful but in the metro Bundler iam getting the below error
![errors](https://user-images.githubusercontent.com/49357162/219546515-dddb1c41-223a-4c94-ab42-1a52487167b0.png)

**To Reproduce**
1. Make a New React native Project 
2. Then Run npm install @nhost/react library package 
3. In app.js file add NhostProvider as your root component and run it

**Expected behavior**
I want app to run properly 

**Desktop (please complete the following information):**
 - OS: Window 10
 - Android Phone Pixel 6

**Smartphone (please complete the following information):**
 - Device:Pixel 6
 - OS:Android 11
If I remove the nhost library its working properly.

beepsoft:
The same with RN/Expo: https://github.com/nhost/nhost/issues/1607

lachu97:
Ok but that's not a good fix, i guess there's a problem with the nhost/react module.Downvaluing ES version not comfortable in a very large codebase

beepsoft:
@lachu97 Yes, you are right. There are 2 problems currently with the nhost sdk. One is related to react native web (my patch solves that hopefully) and the other one is related to the way axios is used by the sdk.

lachu97:
@beepsoft Thanks hopefully it runs properly in the next version.

szilarddoro:
Hey @lachu97! I'm working on a fix that'll solve the issue you are experiencing. It is most likely related to Axios. Hopefully we can release that fix in the beginning of next week.

szilarddoro:
A quick follow-up on this one. We plan to release a new major version at the end of this month / the beginning of March, eliminating Axios. This means the fix will be delayed a little bit because my change would cause conflicts with https://github.com/nhost/nhost/pull/1348. Let's come back to this in a few days.

szilarddoro:
The new version without Axios has been released. Could you folks please give it a shot?

beepsoft:
Yes! 🥳  I can confirm that with `"@nhost/nhost-js": "^2.0.2"` it works on iOS.

However it still fails with RN web like here: https://github.com/nhost/nhost/issues/1607

lachu97:
[...]

Issue #1607, opened 2023-02-10
React Native Expo for ios/web fails with 'Module parse failed' when using @nhost/nhost-js 1.13.2


beepsoft:
Here's a repo to reproduce the error:

https://github.com/beepsoft/nhost-js-expo-web-problem2

When using the `@nhost/nhost-js` (1.13.2) in vanilla React Native Expo 46 or above for web  starting the app fails with an error like this in the web console:

```
wrapNativeSuper.js:31 Uncaught Error: Module parse failed: Unexpected token (285:143)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
|   };
| }
> const G = typeof window < "u", K = /* @__PURE__ */ new Map(), Pe = (s) => G && typeof localStorage < "u" ? localStorage.getItem(s) : K.get(s) ?? null, Ce = (s, e) => {
|   G && typeof localStorage < "u" ? e ? localStorage.setItem(s, e) : localStorage.removeItem(s) : e ? K.set(s, e) : K.has(s) && K.delete(s);
| }, Ne = (s, e) => {
    at ./node_modules/@nhost/hasura-auth-js/dist/index.esm.js (wrapNativeSuper.js:31:1)
    at __webpack_require__ (bootstrap:789:1)
    at fn (bootstrap:100:1)
    at Module.<anonymous> (index.esm.js:1:1)
    at ./node_modules/@nhost/nhost-js/dist/index.esm.js (index.esm.js:250:1)
    at __webpack_require__ (bootstrap:789:1)
    at fn (bootstrap:100:1)
    at ./App.tsx (App.tsx:1:1)
    at __webpack_require__ (bootstrap:789:1)
    at fn (bootstrap:100:1)
    at ./node_modules/expo/AppEntry.js (AppEntry.js:1:1)
    at __webpack_require__ (bootstrap:789:1)
    at fn (bootstrap:100:1)
    at 1 (log.js:59:1)
    at __webpack_require__ (bootstrap:789:1)
    at bootstrap:856:1
    at bootstrap:856:1
./node_modules/@nhost/hasura-auth-js/dist/index.esm.js	@	wrapNativeSuper.js:31
__webpack_require__	@	bootstrap:789
fn	@	bootstrap:100
(anonymous)	@	index.esm.js:1
./node_modules/@nhost/nhost-js/dist/index.esm.js	@	index.esm.js:250
__webpack_require__	@	bootstrap:789
[...]

PR #1632, opened 2023-02-17
Use es2019 build target

beepsoft:
This is a fix for https://github.com/nhost/nhost/issues/1623 and https://github.com/nhost/nhost/issues/1607

It seems that if the target to esbuild is es2019 it transforms the nullish coalescing (??) operator, which causes problem in React Native and Expo. Starting from es2020 it does not transforms those.

From the documentation I thought we should use es2020, but actually we need es2019 to make it work.

https://esbuild.docschina.org/content-types/#javascript

I am not sure of all the implications of the build targets, but I guess targeting older JS versions is overall safe.

### DIFF
--- a/config/vite.lib.config.js
+++ b/config/vite.lib.config.js
@@ -35,6 +35,7 @@ export default defineConfig({
     }
   },
   build: {
+    target: "es2019", // https://esbuild.docschina.org/content-types/#javascript
     sourcemap: true,
     lib: {
       entry,


szilarddoro:
I'm closing this in favor of #1665. Could you take a quick look at that PR @beepsoft?

There is a quite recent issue with RN Expo + Web though that is tracked here: https://github.com/expo/expo-cli/issues/4639. I also experienced this in the example project.

PR #1650, opened 2023-02-22
GraphQL client: fix snake_case bug and improve readme

plmercereau:
we need a changeset if we want to re-publish the readme on https://www.npmjs.com

### DIFF
--- a/.changeset/ninety-carrots-teach.md
+++ b/.changeset/ninety-carrots-teach.md
@@ -1,0 +1,5 @@
+---
+'@nhost/graphql-js': patch
+---
+
+Correct type inference on snake case operations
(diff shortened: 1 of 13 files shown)


szilarddoro:
We should extend the README with information on how to include the codegen script in the `package.json` file.

plmercereau:
> We should extend the README with information on how to include the codegen script in the `package.json` file.

true - it is not required when using yarn, but it makes things clearer for users that don't know the `yarn <a-bin-file>` command works.

elitan (approved):

elitan:
> we need a changeset if we want to re-publish the readme on https://www.npmjs.com/

We do what the README to be correct on npmjs.com

elitan (approved):

PR #1666, opened 2023-02-24
fix(nhost-js): use correct URL for functions requests

szilarddoro:


packages/nhost-js/src/clients/functions/index.ts
```diff
@@ -59,27 +59,32 @@ export class NhostFunctionsClient {
     body: D,
     config?: NhostFunctionCallConfig
   ): Promise<NhostFunctionCallResponse<T>> {
+    const requestUrl = `${this.url}${url.startsWith('/') ? url : `/${url}`}`
```
elitan, line 62: I think we can make this more readable.

elitan (approved):

packages/nhost-js/src/clients/functions/index.ts
szilarddoro: absolutely, done

PR #1670, opened 2023-02-26
Functions fix + tests

elitan:

PR #1673, opened 2023-02-27
Remove double forward slash "/" from function call

ctrlemo:
Remove double forward slash "/" from function call. Noticed today that after upgrading to the latest 2.0.4 @nhost/react package that function calls made with the Nhost client were being requested using double // forward slashes in the url.
http://localhost:1337/v1/functions//path/subpath

this happens regardless of which way I call it
nhost.functions.call("/path/subpath",.....
or
nhost.functions.call("path/subpath",.....

Small change but this should fix the issue.


elitan:
Nice catch!

Closed in favor of https://github.com/nhost/nhost/pull/1670

PR #1698, opened 2023-03-02
tests: hasura-storage-js

elitan:


packages/hasura-storage-js/src/utils/index.ts
```diff
@@ -7,7 +7,7 @@ export const appendImageTransformationParameters = (
   params: StorageImageTransformationParams
 ): string => {
   const queryParameters = Object.entries(params)
-    .map(([key, value]) => `${Array.from(key)[0]}=${value}`)
+    .map(([key, value]) => `${key[0]}=${value}`)
```
szilarddoro, line 10: I think this makes it more readable:

```suggestion
    .map(([key, value]) => `${key.charAt(0)}=${value}`)
```

szilarddoro (approved):
Just a minor improvement proposal

Issue #1576, opened 2023-02-02
Self Hosting


Bijit-Mondal:
**404 on self hosting**
I went through the documentation and use `docker-compose.yml` to install nhost on my cloud setup, but when ever I go to the public IP 404 page is coming as response. how to resolve

Samsonov0:
+

Bijit-Mondal:
Hey,have you set the .env variables

Samsonov0:
yes, I did cp .env.example .end and cp env.example .env.development, I tried to change the host address directly in docker-compose, climbed in dashboard configurations and everything like that. as I noticed, only dashboard doesn't start.  functions, for example, work fine

ngi:
maybe you should remove the "localhost" parts in traefik.http.routers.functions.rule in your docker-compose.yaml

Bijit-Mondal:
> maybe you should remove the "localhost" parts in traefik.http.routers.functions.rule in your docker-compose.yaml
have you succeed to install nhost
please help me out

zuohuadong:
+1

chaitugsk07:
+1

chaitugsk07:
+1

zuohuadong:
Has it made any progress?

dbarrosop:
Looking at the [example](https://github.com/nhost/nhost/tree/main/examples/docker-compose) and following the instructions in the `README.md` everything seems to work without any problem. Keep in mind, this is an example on how to run it on your own without using our cli, it will most likely need to be adapted for your existing setup. As you can imagine, we can't support every possible way users may want to deploy this as there are infinite possibilities here. If you are facing issues self-hosting it I'd suggest you contact your network administator or devops person so they can help you in your environment.

I am closing this issue as the example works, if someone has problems following the example to the letter feel free to open a new issue.

PR #1703, opened 2023-03-03
Updated docker compose

elitan:
fixes: #1576

### DIFF
--- a/examples/docker-compose/docker-compose.yaml
+++ b/examples/docker-compose/docker-compose.yaml
@@ -1,18 +1,5 @@
 version: '3.6'
 services:
-  traefik:
-    image: "traefik:v2.5"
-    container_name: "traefik"
-    command:
-      - "--api.insecure=true"
-      - "--providers.docker=true"
-      - "--providers.docker.exposedbydefault=false"
-      - "--entrypoints.web.address=:1337"
-    ports:
-      - "1337:1337"
-      - "9090:8080"
-    volumes:
-      - "/var/run/docker.sock:/var/run/docker.sock:ro"
   postgres:
     image: postgres
     restart: always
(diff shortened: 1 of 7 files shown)


examples/docker-compose/package.json
```diff
@@ -6,6 +6,8 @@
     "e2e": "vitest run"
   },
   "devDependencies": {
-    "cross-fetch": "^3.1.5"
+    "@nhost/nhost-js": "workspace:*",
```
szilarddoro, line 9:
```suggestion
    "@nhost/nhost-js": "*",
```

I think we want to use `*` here because users might want to copy the project outside of the repository where `workspace:*` is not available. I don't think this is a frequent use case, but I remember we talked about this in the past and decided to go with `*`.

szilarddoro (approved):
LGTM, but maybe we want to wait for @elephant3's opinion on this one too.

examples/docker-compose/docker-compose.yaml
```diff
@@ -132,16 +103,12 @@ services:
       SMTP_PORT: ${AUTH_SMTP_PORT:-1025}
       SMTP_PASS: ${AUTH_SMTP_PASS:-password}
       SMTP_USER: ${AUTH_SMTP_USER:-user}
-      SMTP_SECURE: "${AUTH_SMTP_SECURE:-false}"
+      SMTP_SECURE: '${AUTH_SMTP_SECURE:-false}'
```
elephant3, line 106: Is that resolved in single quotes?

elephant3 (commented):
LGTM, just one comment

Bijit-Mondal:
In this deployed version, when I try to login it showing 404 not found.

komali2:
Hello, I'm not sure if self-hosting is a desired pathway for this project anymore, but since it's the path I'm following, I'm curious why this branch was never merged? It seems to change port mappings and etc, was this the intended direction?

PR #1712, opened 2023-03-07
fix(nhost-js): correct return type in functions

### LINKED ISSUE #1702: nhost.functions.call api behaves different in latest nhost-react package
Noticed that in latest nhost-react package.
I can no longer pass **null** at data parameter or payload to the function.

```
  const { res, error } = await nhost.functions.call(
        "/your/path",
        null,
        {
          headers: { Authorization: Bearer ${myAccessToken} },
        }
      );

      if (res?.status === 200 && !error) {
        const { myProperty } = res.data;
```

Also noticed that the res (response) object data property is no longer parsed to javascript object. Instead it comes back  still encoded as a json string
it comes back as something like this.
`{\"myProperty\": "myPropertyValue"}`

I can no longer access it like res.data.myProperty or desctructure it `const { myProperty } = res.data;`

I had to revert back to "@nhost/react": "^0.12.4" in order for my code to work again. 😥

szilarddoro:
Fixes #1702

### DIFF
--- a/packages/nhost-js/src/clients/functions/index.ts
+++ b/packages/nhost-js/src/clients/functions/index.ts
@@ -56,7 +56,7 @@ export class NhostFunctionsClient {
    */
   async call<T = unknown, D = any>(
     url: string,
-    body: D,
+    body: D | null,
     config?: NhostFunctionCallConfig
   ): Promise<NhostFunctionCallResponse<T>> {
     const headers: HeadersInit = {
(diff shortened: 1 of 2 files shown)


packages/nhost-js/src/clients/functions/index.ts
```diff
@@ -80,7 +80,7 @@ export class NhostFunctionsClient {
 
       let data: T
 
-      if (result.headers.get('content-type') === 'application/json') {
+      if (result.headers.get('content-type')?.includes('application/json')) {
```
szilarddoro, line 83: the server sometimes returns this header as `application/json; <some-extra-info>` so we can't always assume that it strictly equals `application/json`

elitan (approved):

PR #1729, opened 2023-03-10
fix(nextjs): allow `subdomain`, `region` and service URLs

### LINKED ISSUE #1707: createServerSideClient still takes backendUrl
I'm trying to use `createServerSideClient` from the `nhost/nextjs` package, but it still takes `backendUrl` as a string which is no longer supported.

What should I pass in for `backendUrl` since I'm only given `region` and `subdomain`?

### LINKED ISSUE #1727: Deprecated backendUrl parameter in getNhostSession function of @nhost/nextjs package should be updated to accept alternative parameters.
The `getNhostSession` function in the `@nhost/nextjs` package only accepts the now deprecated `backendUrl`.

I _think_ that the `getNhostSession` should accept either `subdomain`/`region` or individual URLs for each endpoint (graphql / auth / storage / functions).

szilarddoro:
- Fixes #1707 
- Fixes #1727

### DIFF
--- a/examples/nextjs/pages/_app.tsx
+++ b/examples/nextjs/pages/_app.tsx
@@ -6,7 +6,6 @@ import { inspect } from '@xstate/inspect'
 import type { AppProps } from 'next/app'
 import Head from 'next/head'
 import NavBar from '../components/NavBar'
-import { BACKEND_URL } from '../helpers'
 import '../styles/globals.css?inline'
 
 const devTools = typeof window !== 'undefined' && !!process.env.NEXT_PUBLIC_DEBUG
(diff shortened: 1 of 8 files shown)


examples/nextjs/pages/_app.tsx
```diff
@@ -16,7 +15,7 @@ if (devTools) {
     iframe: false
   })
 }
-const nhost = new NhostClient({ backendUrl: BACKEND_URL, devTools })
+const nhost = new NhostClient({ subdomain: 'localhost', devTools })
```
szilarddoro, line 18: We should change every occurrence of `subdomain: 'localhost'` to `subdomain: 'local'` later

elitan (approved):

PR #1772, opened 2023-03-23
Update serverless-functions to link to event trigger docs

wollerman:
Based on https://github.com/nhost/nhost/discussions/1752#discussioncomment-5380702 adding a link so users know to look at the other section for more information on configuring functions.

### DIFF
--- a/docs/docs/serverless-functions.mdx
+++ b/docs/docs/serverless-functions.mdx
@@ -25,8 +25,6 @@ export default (req: Request, res: Response) => {
   res.status(200).send(`Hello ${req.query.name}!`)
 }
 ```
-    
-The same [environment variables that are used to configure event triggers](https://docs.nhost.io/database/event-triggers#format) can be used to authenticate regular serverless functions.
 
 :::info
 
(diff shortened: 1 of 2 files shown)


docs/docs/serverless-functions.mdx
```diff
@@ -25,6 +25,8 @@ export default (req: Request, res: Response) => {
   res.status(200).send(`Hello ${req.query.name}!`)
 }
 ```
+    
+The same [environment variables that are used to configure event triggers](https://docs.nhost.io/database/event-triggers#format) can be used to authenticate regular serverless functions.
```
szilarddoro, line 29: I think it would make more sense to add this to the "Environment Variables" section on this page.

wollerman: [That was my initial idea as well](https://github.com/nhost/nhost/discussions/1752#discussioncomment-5380251). I was following @dbarrosop 's [suggestion](https://github.com/nhost/nhost/discussions/1752#discussioncomment-5380702). Happy to adjust the docs either way!

dbarrosop: @szilarddoro you are probably more familiar with the docs so whatever you think is best :)

PR #1800, opened 2023-04-01
Update local URLs to new format

Glenas7:
With the change of local URLs to the subdomain format with "local" as subdiomain and no region, I believe these URLs should be updated here.


szilarddoro:
We have a few other occurrences of `http://localhost:1337` in `docs/docs/graphql/index.mdx` and `docs/docs/graphql/remote-schemas/stripe.mdx` as well. Could you please change those too?

szilarddoro (approved):

PR #1830, opened 2023-04-11
Add generateLinks instead of link and onError args

rikardwissing:
~Breaking change, but makes it more versatile.~
Not tested

Edit: There is a change in behavior when using customLink. But I think this is how it was intended to be implemented?


rikardwissing:
Feel free to make changes or make it backwards compatible.

rikardwissing:
@szilarddoro

integrations/apollo/src/index.ts
```diff
@@ -23,8 +22,7 @@ export type NhostApolloClientOptions = {
   fetchPolicy?: WatchQueryFetchPolicy
   connectToDevTools?: boolean
   cache?: InMemoryCache
-  onError?: RequestHandler
```
szilarddoro, line 26: We generally try to avoid breaking changes. Instead of dropping it right away, we could still keep `onError` and mark it as deprecated on the interface. We should still add it to the merged links for now. The same goes for `link`.

integrations/apollo/src/index.ts
```diff
@@ -23,8 +22,7 @@ export type NhostApolloClientOptions = {
...
-  link?: ApolloClient<any>['link']
+  generateLinks?: Function
```
szilarddoro, line 25: Is there a type for links in general that we can use for this function?

I'm thinking of something like this:
```ts
generateLinks?: (...links: ApolloLink[]) => void
```

szilarddoro (changes requested):

integrations/apollo/src/index.ts
rikardwissing: Sounds good to me I can make the change.

integrations/apollo/src/index.ts
rikardwissing: Sounds good, I'll make the change

rikardwissing:
Made some changes @szilarddoro 
Not tested so you'll have to do that 😄 

There is a change in behavior when using customLink. But I think this is how it was intended to be implemented?

integrations/apollo/src/index.ts
```diff
@@ -24,7 +25,8 @@ export type NhostApolloClientOptions = {
   connectToDevTools?: boolean
   cache?: InMemoryCache
   onError?: RequestHandler
-  link?: ApolloClient<any>['link']
+  link?: ApolloLink
+  generateLinks: (links: (ApolloLink | RequestHandler)[]) => (ApolloLink | RequestHandler)[]
```
szilarddoro, line 29:
```suggestion
  generateLinks?: (links: (ApolloLink | RequestHandler)[]) => (ApolloLink | RequestHandler)[]
```

I think we shouldn't necessarily force everyone to use it. Also, I'm not sure if this really needs to be a function and not just simply `(ApolloLink | RequestHandler)[]` instead.

integrations/apollo/src/index.ts
```diff
@@ -24,7 +25,8 @@ export type NhostApolloClientOptions = {
   connectToDevTools?: boolean
   cache?: InMemoryCache
   onError?: RequestHandler
```
szilarddoro, line 27:
```suggestion
  /**
   * @deprecated Please use `generateLinks` instead.
   */
  onError?: RequestHandler
```

Please note that depending on the result of the open discussion, we might want to change the deprecation message.

integrations/apollo/src/index.ts
```diff
@@ -124,6 +127,20 @@ export const createApolloClient = ({
       )
     : httpLink
 
+  const links = [];
+
+  if(onError) {
```
szilarddoro, line 132: This looks incorrectly formatted. Please run Prettier locally. I was under the impression that we had a git hook for this, but apparently, we don't 🤔

integrations/apollo/src/index.ts
rikardwissing: Yes, that's exactly the use case.

It's nice to be able to reuse the httplink and wslink that nhost manages for token renewal etc. But the order and all other links are nice to have full control over.

integrations/apollo/src/index.ts
```diff
@@ -124,6 +133,20 @@ export const createApolloClient = ({
...
+    links.push(onError);
+  }
+
+  if(customLink) {
+    links.push(customLink);
+  }
+
+  links.push(splitLink);
+
+  const link = from(generateLinks(links))
```
szilarddoro, line 148: Now that `generateLinks` is not mandatory, it's possible that this function is undefined. Could you please handle that case too?

szilarddoro (commented):
I'll do the formatting part once this last issue is fixed.

integrations/apollo/src/index.ts
rikardwissing: That's why I had it default to a function :) I'll remove that and check if it's set or not

szilarddoro: Oh, well. I didn't notice that, sorry. It's fine if you leave it like that.

integrations/apollo/src/index.ts
rikardwissing: Done, nice if you can handle the formatting 💯

szilarddoro (approved):

szilarddoro:
Thanks for your contribution @rikardwissing! 🙏🏼

rikardwissing:
> Thanks for your contribution @rikardwissing! 🙏🏼

My pleasure @szilarddoro!

PR #1837, opened 2023-04-12
Wait for valid token or sign out before establishing connection.

rikardwissing:
Some background: We are using nhost to develop a react-native app.

The issue we are having is that when we go back to the app after the token has been expired and a query is fired before nhost auth handles a token refresh, that query obviously fail.

This is the way I solved the issue. Maybe there's a nicer way to do it, maybe subscribing to token changes is better? Feedback is welcome if you have any ideas. But this should ensure that the connection waits to be established until we have a valid token (or are signed out).

Just threw this together now. Will test how it works for our issue tomorrow.

Update: works great for us and we don't have the issue anymore

### DIFF
--- a/integrations/apollo/src/index.ts
+++ b/integrations/apollo/src/index.ts
@@ -59,7 +59,7 @@ export const createApolloClient = ({
   let accessToken: AuthContext['accessToken'] | null = null
 
   const isTokenValid = () =>
-    accessToken?.value && accessToken?.expiresAt && accessToken.expiresAt > new Date()
+    !!accessToken?.value && !!accessToken?.expiresAt && accessToken?.expiresAt > new Date()
 
   const awaitValidTokenOrNull = () =>
     new Promise((resolve) => {
--- a/integrations/apollo/src/index.ts
+++ b/integrations/apollo/src/index.ts
@@ -65,7 +65,6 @@ export const createApolloClient = ({
     new Promise((resolve) => {
       // doing this as an interval to avoid race conditions that I imagine can happen if listening to token changes. Maybe there's a better way.
       const interval = setInterval(() => {
-        console.log('checking for token change')
         if (!accessToken || isTokenValid()) {
           clearInterval(interval)
           resolve(true)
--- a/integrations/apollo/src/index.ts
+++ b/integrations/apollo/src/index.ts
@@ -11,7 +11,7 @@ import {
 import { setContext } from '@apollo/client/link/context'
 import { GraphQLWsLink } from '@apollo/client/link/subscriptions'
 import { getMainDefinition } from '@apollo/client/utilities'
-import { NhostClient } from '@nhost/nhost-js'
+import { AuthContext, NhostClient } from '@nhost/nhost-js'
 
 import { createRestartableClient } from './ws'
 const isBrowser = typeof window !== 'undefined'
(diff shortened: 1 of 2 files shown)


integrations/apollo/src/index.ts
```diff
@@ -56,19 +56,38 @@ export const createApolloClient = ({
   const uri = backendUrl
   const interpreter = nhost?.auth.client.interpreter
 
-  let token: string | null = null
+  let accessToken: AuthContext['accessToken'] | null = null
 
-  function getAuthHeaders() {
+  const isTokenValid = () =>
+    accessToken?.value && accessToken?.expiresAt && accessToken.expiresAt > new Date()
```
szilarddoro, line 62:
```suggestion
    !!accessToken?.value && !!accessToken?.expiresAt && accessToken?.expiresAt > new Date()
```

There are cases (e.g: when `accessToken?.value` is falsy) where this function would not return a `boolean`.

integrations/apollo/src/index.ts
```diff
@@ -56,19 +56,38 @@ export const createApolloClient = ({
...
+
+  const awaitValidTokenOrNull = () =>
+    new Promise((resolve) => {
+      // doing this as an interval to avoid race conditions that I imagine can happen if listening to token changes. Maybe there's a better way.
+      const interval = setInterval(() => {
```
szilarddoro, line 67: Instead of this interval-based approach, maybe we could use something like this:

```ts
new Promise((resolve) => {
  interpreter?.onTransition(async (_state, event) => {
    if (event.type === 'TOKEN_CHANGED') {
      resolve(true)
    }
  })
})
```

if I understand this piece of code correctly, all we care about here is the time when the local `accessToken` variable changes

szilarddoro (changes requested):
Thanks @rikardwissing, these changes look good. Could you please give my suggestions a try?

integrations/apollo/src/index.ts
rikardwissing: Yes, but will TOKEN_CHANGED trigger when we get signed out as well or should we listen to SIGNED OUT as well? Cause i think we want the request to continue then as well but with the public role.

Wouldn't we want to clear this listener after we're done btw?

szilarddoro: Yes, the `TOKEN_CHANGED` event should be sent when the user is signed out.

We should clear the listener for sure. I just wanted to sketch out an alternative approach.

rikardwissing: Looking at this a bit more. Wouldn't this risk the onTransition in awaitValidTokenOrNull to be fired before the token variable is set in the other onTransition in this file? Then you might still use the invalid token even after the TOKEN_CHANGED have fired inside awaitValidTokenOrNull.

If this approach is implemented, then maybe we should just return the token directly from the onTransition state in awaitValidTokenOrNull and use that? Then you could also remove the need to store token in that let variable. But since it is a bit bigger changes I'm gonna leave that up to you.

I'm quite happy with the interval checking thing, since I don't want to make too big changes to this file.

szilarddoro: I just checked the state transitions. It is probably not the safest way to determine the latest access token so we're better off with this interval-based approach for now.

integrations/apollo/src/index.ts
```diff
@@ -56,19 +56,38 @@ export const createApolloClient = ({
...
 
-  function getAuthHeaders() {
+  const isTokenValid = () =>
+    !!accessToken?.value && !!accessToken?.expiresAt && accessToken?.expiresAt > new Date()
+
+  const awaitValidTokenOrNull = () =>
+    new Promise((resolve) => {
+      // doing this as an interval to avoid race conditions that I imagine can happen if listening to token changes. Maybe there's a better way.
+      const interval = setInterval(() => {
+        console.log('checking for token change')
```
szilarddoro, line 68:
```suggestion
```

szilarddoro (changes requested):
Once this `console.log` is removed, I think we're good to merge. Also, could you please generate a changeset that would describe your changes in a sentence? You can generate one by running`pnpm -w changeset`. A patch version bump would be fine.

rikardwissing:
@szilarddoro All done!

szilarddoro (approved):

PR #1849, opened 2023-04-18
add compute section to the docs

nunopato:


docs/docs/platform/compute.mdx
```diff
@@ -0,0 +1,28 @@
...
+---
+
+Compute resources are the fundamental units that represent the processing power and memory available to your Nhost projects. The primary compute resources are vCPU and RAM. This documentation outlines the key aspects of compute resources in the context of the Nhost Cloud Platform.
+
+
+## Shared vs Dedicated Compute
+
+All free and pro projects use shared vCPUs:
+
+- 2vCPUs / 1GiB for free projects
```
dbarrosop, line 14: should we outline the distribution between services?

docs/docs/platform/compute.mdx
```diff
@@ -0,0 +1,28 @@
...
+- 2vCPUs / 2GiB for pro projects
+
+This is fine if your apps mostly run at low to medium load, occasionally burst for brief periods of time, and can tolerate drops in performance. It is important to understand that the availability of CPU time is not guaranteed.
+
+### Dedicated Compute
+
+On the other hand, for high production workloads where time is important, or variable performance is not at all tolerable, you should consider configuring your project to use dedicated compute resources.
```
dbarrosop, line 21: > where time is important

do you mean latency?

docs/docs/platform/compute.mdx
```diff
@@ -0,0 +1,28 @@
...
+- 2vCPUs / 1GiB for free projects
+- 2vCPUs / 2GiB for pro projects
+
+This is fine if your apps mostly run at low to medium load, occasionally burst for brief periods of time, and can tolerate drops in performance. It is important to understand that the availability of CPU time is not guaranteed.
+
+### Dedicated Compute
+
+On the other hand, for high production workloads where time is important, or variable performance is not at all tolerable, you should consider configuring your project to use dedicated compute resources. 
+
+By choosing dedicated compute, you can be sure that performance is guaranteed and can match the unique demands of your projects.
```
dbarrosop, line 23: not sure I like this sentence. What about something like:

With dedicated compute, resources are guaranteed for your project so you don't have to contend for them.

In addition to the resources fully dedicated to the project, apps are allowed to burst if demand requires it and resources are available. If properly sized, dedicated resources should guarantee the performance of your application while allowing for occassional burts.

dbarrosop:
maybe mention the 1:2 ration and the 250/128 steps?

docs/docs/platform/compute.mdx
nunopato: done

docs/docs/platform/compute.mdx
nunopato: good catch

docs/docs/platform/compute.mdx
nunopato: updated it to your suggestion

PR #1903, opened 2023-05-05
Fixed linking to wrong destination in docs

hrmoller:
On the docs page describing login methods [`Security Keys (WebAuthn)`](https://docs.nhost.io/authentication/sign-in-methods) is linking to `/authentication/sign-in-with-phone-number-sms`, however it should be linking to `/authentication/sign-in-methods`.

This PR fixes that.


szilarddoro (approved):
Good catch! Thanks

PR #2082, opened 2023-06-29
fix(dashboard): use dynamic validation schema for password editing

### LINKED ISSUE #2081: Wrong client validation on password length when updating from nhost dashboard
I'm using the nhost dashboard and I want to update a user's password. 
I just updated the max psw length from 9 to 3 chars.
However, when I try to update the user's password I'm getting a client error validation saying that the psw should be at least 8 chars long. 

I guess if you want to have client-side validation make sure it matches whatever the max length for the password is.. :) 

Image

szilarddoro:
Fixes #2081

### DIFF
--- a/.changeset/kind-horses-act.md
+++ b/.changeset/kind-horses-act.md
@@ -1,0 +1,5 @@
+---
+'@nhost/dashboard': patch
+---
+
+fix(users): use the password length from the config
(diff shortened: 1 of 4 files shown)


onehassan (approved):
LGTM

PR #2105, opened 2023-07-06
Update serverless-functions.mdx

st3phan:
Please correct me if I'm wrong but I believe the endpoints in the docs are wrong.


dbarrosop (approved):

dbarrosop:
You are 100% right, thanks for this.

PR #2136, opened 2023-07-23
fix: back to previous limit

tawsbob:
A Quick fix to improve the user experience at users page which the limit is currently one and make worse to manage the user

### DIFF
--- a/.changeset/fuzzy-cooks-hide.md
+++ b/.changeset/fuzzy-cooks-hide.md
@@ -1,0 +1,5 @@
+---
+'@nhost/dashboard': patch
+---
+
+fix(auth): fix users pagination limit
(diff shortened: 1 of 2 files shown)


dbarrosop:
Would you mind adding a changeset? You have the instructions [here](https://github.com/nhost/nhost/blob/main/DEVELOPERS.md#changesets). If you are unable we can add it ourselves but it will certainly expedite things if you can do it yourself.

tawsbob (commented):
thanks!

tawsbob (commented):
@dbarrosop and @onehassan do we need to change the pkg version?

onehassan:
@tawsbob once this PR is merged, the version will be bumped automatically and we will have a new release

PR #2182, opened 2023-08-18
feat: add sveltekit example

### LINKED ISSUE #903: Feature Request: Svelte and SvelteKit SDK
The Svelte community is growing faster than you think.

More and more people are pouring down into the awesomeness of this framework.

It would really help if we could get an official SDK with well-written documentation on how to use it.

onehassan:
fixes #903


.gitignore
```diff
@@ -20,6 +20,7 @@ coverage/
 dist/
 umd/
 lib/
```
dbarrosop, line 22: maybe it's better to remove `lib` (and any folder you don't recognize)

dbarrosop (approved):

PR #2197, opened 2023-08-23
fix(hasura-auth-js): make sure CodifiedError works on non v8 browsers

### LINKED ISSUE #2187: react-apollo-example: verify use of useAddSecurityKey
Seems to work fine but adding a second key doesn't seem to work. We need to check if it's a frontend, sdk or backend issue.

onehassan:
fixes #2187

### DIFF
--- a/.changeset/thin-candles-float.md
+++ b/.changeset/thin-candles-float.md
@@ -1,0 +1,5 @@
+---
+'@nhost/hasura-auth-js': patch
+---
+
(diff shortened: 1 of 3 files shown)


onehassan:
Now it correctly throws this error

image

onehassan:
[More](https://github.com/w3c/webauthn/issues/1566) on `InvalidStateError`

onehassan:
The main issue is that `Error.captureStackTrace` doesn't work on non V8 browsers such as Safari

PR #2214, opened 2023-09-01
feat(dashboard): update usage metrics

### LINKED ISSUE #2205: update limits in the dashboard
1. Make sure limits are up to date.
2. Add egress limits
3. Add lambda limits

onehassan:
fixes #2205

### DIFF
--- a/dashboard/src/features/projects/overview/components/OverviewUsage/OverviewUsage.tsx
+++ b/dashboard/src/features/projects/overview/components/OverviewUsage/OverviewUsage.tsx
@@ -96,7 +96,7 @@ export function OverviewUsageMetrics() {
     remoteAppMetricsData?.filesAggregate?.aggregate?.sum?.size || 0;
   const totalStorage = currentProject?.plan?.isFree
     ? 1 * 1000 ** 3 // 1 GB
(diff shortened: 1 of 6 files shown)


dashboard/src/features/projects/overview/components/OverviewUsage/OverviewUsage.tsx
```diff
@@ -105,14 +105,24 @@ export function OverviewUsageMetrics() {
   // metrics for functions
   const usedFunctions = functionsInfoData?.app.metadataFunctions.length || 0;
   const totalFunctions = currentProject?.plan?.isFree ? 10 : 50;
+  const functionsDuration = projectMetrics?.functionInvocations.value || 0;
```
dbarrosop, line 108: I think this should be:

```suggestion
  const functionsDuration = projectMetrics?.functionsDuration.value || 0;
```

also, keep in mind the value returned is in seconds.

dbarrosop:
We should probably remove egress from the top boxes as it is now redundant.

onehassan:
> We should probably remove egress from the top boxes as it is now redundant.

In that case should I change the grid to 2 columns like this
image

dbarrosop (approved):

Issue #2216, opened 2023-09-01
Sign in with apple is doesn't work with nHost


thoth-seshat:
**Describe the bug**
I'm attempting to use Sign in with Apple with nhost. When I attempt to call: `nhost.auth.signIn({ provider: "apple" });` I get an error: 

```
Possible Unhandled Promise Rejection (id: 1):
TypeError: Cannot set property 'href' of undefined
```

I've configured the nhost client as directed in the docs with an identifier, callback urls, team ids etc. 

**To Reproduce**
React Native with expo v48
1. Go through the nhost docs to configure sign in with apple, make sure your dashboard is all configured
2. Set up an nhost client via their js sdk 
3. Attempt to sign in with apple ```nhost.auth.signIn({ provider: "apple" });```
4. See the following error:
```Require cycles are allowed, but can result in uninitialized values. Consider refactoring to remove the need for a cycle.
 WARN  Possible Unhandled Promise Rejection (id: 0):
TypeError: Cannot set property 'href' of undefined
TypeError: Cannot set property 'href' of undefined
```

**Expected behavior**
I'd expect to see an option to log in with apple.

**Screenshots**
<img width="920" alt="Screenshot 2023-09-01 at 3 44 59 PM" src="https://github.com/nhost/nhost/assets/122938275/d39427b0-089c-4564-8837-9032d0cdd75f">

**Desktop (please complete the following information):**
 - OS: iOS Simulator 14
 - Version React Native 48

**Additional context**
I've been having a discussion in Discord and was told I'm using the React SDK instead of the JS one however, this is not the case. I'm directly using the JS sdk now.

I've also pulled my config from the nhost dashboard locally and that doesn't seem to help. I'm able to get Auth working fine with Firebase Auth but want to keep everything in nhost. I'll also note, I'd probably just use the API to set everything up but there is no documentation on how authentication works in nhost (that I'm aware of) it's more like, go code diving in our repository and figure it out). which is starting to take it's toll on productivity since  I can't really just look things up. 

Anyway, I've lost the better part of week on this problem and so am trying here. Has anyone been able to get login working with React Native and Nhost?

georgii-ivanov:
+1 for this issue.

Looks like they don't support React Native officially. IMO, they should have a fallback using react-native/expo Linking.

This issue is 2 mo. old and it's pretty sad for me it has still no reaction. Especially, when Apple enforces us to support Apple Sign In option along with others like Google Sign In.

pkreipke:
[...]

PR #2217, opened 2023-09-02
Allow settting isBrowser manually

thoth-seshat:
#2216 2216

Id like to manually set this since I'm able to pass this in via my app easier.


onehassan:
This issue was resolved in [PR #2716](https://github.com/nhost/nhost/pull/2716), which addresses the problem mentioned

PR #2252, opened 2023-09-20
chore: added codeql to CI

dbarrosop:


onehassan (approved):
Can we add a changeset in here as well

PR #2309, opened 2023-10-10
fix: integrations: apollo: set accessToken to null after TOKEN_CHANGED event on sign-out

### LINKED ISSUE #2297: Apollo: cached/public queries do not return data after `nhost.auth.signOut()`
**Describe the bug**
I have this query, with public permissions, that is run (with Apollo's `useQuery`) while not being logged in.
Then after I sign in and then sign out, that query is no longer returning data and hangs in a loading state.

When I disable [this line](https://github.com/nhost/nhost/blob/236996a903fe880e5b1ac8d6f54fe309dfb91940/integrations/apollo/src/index.ts#L204C7-L204C46) in the `@nhost/react-apollo` pacakge, the public query returns data normally after a `signOut()`.
However, queries that require an access token obviously no longer work.

Am I perhaps on the trail of a possible bug? Or is this just an implementation error on my side.

Here's a screen recording that shows the issue in more detail:
https://www.dropbox.com/scl/fi/ii4hke3k9tjqtavpw28z4/2023.10.04-Nhost-apollo-cache-issue.mp4?rlkey=ofdkga3skgvwvb277a52x6nzw&dl=0

**To Reproduce**
1. Make sure you are logged out
2. Do a simple query using useQuery
3. Do a sign in
4. Do a sign out
5. Go to the page with the simple query (from step 2) and see that the query is not returning data.

**Expected behavior**
useQuery should return data after a `nhost.auth.signOut()`.

**Packages information**
* @apollo/client: 3.8.4
* @nhost/nextjs: 13.37
* @nhost/nhost-js: 2.2.16
* @nhost/react: 2.0.31
* @nhost/react-apollo: 5.0.35

onehassan:
resolve #2297

### DIFF
--- a/integrations/apollo/src/index.ts
+++ b/integrations/apollo/src/index.ts
@@ -187,7 +187,10 @@ export const createApolloClient = ({
 
   interpreter?.onTransition(async (state, event) => {
     if (['SIGNOUT', 'SIGNED_IN', 'TOKEN_CHANGED'].includes(event.type)) {
-      if (event.type === 'SIGNOUT') {
+      if (
+        event.type === 'SIGNOUT' ||
+        (event.type === 'TOKEN_CHANGED' && state.context.accessToken === null)
+      ) {
         accessToken = null
 
         try {
--- a/.changeset/violet-onions-swim.md
+++ b/.changeset/violet-onions-swim.md
@@ -1,0 +1,5 @@
+---
+'@nhost/apollo': patch
+---
+
+fix: apollo-integration: reset accessToken to null after sign-out


dbarrosop (approved):

integrations/apollo/src/index.ts
```diff
@@ -187,7 +187,10 @@ export const createApolloClient = ({
 
   interpreter?.onTransition(async (state, event) => {
     if (['SIGNOUT', 'SIGNED_IN', 'TOKEN_CHANGED'].includes(event.type)) {
-      if (event.type === 'SIGNOUT') {
+      if (
+        event.type === 'SIGNOUT' ||
+        (event.type === 'TOKEN_CHANGED' && state.context.accessToken === null)
```
st3phan, line 192: I think this fix isn't working. When I debug this I see that `accessToken` is an object. So shouldn't we check the `value` here? Like this:

```ts
(event.type === 'TOKEN_CHANGED' && state.context.accessToken.value === null)
```

PR #2397, opened 2023-12-02
Update docker-compose and .env-example

lectrician1:
The docker-compose example is severely outdated and the past configuration used a docker image for the dashboard that did not allow you to configure the URLs to the various API endpoints if you self-hosted your own dashboard publicly. 

The newest dashboard image allows you to do this so the docker-compose has been updated to use this image and the env variables have been updated accordingly.

Other variables have been updated in the docker-compose to support self-hosting a public instance.

A commented traefik configuration in the docker-compose for the dashboard service also allows the user to configure basic auth to protect a publicly-facing dashboard.


dbarrosop:
Thanks! Forgot to mention you need to run it in the root of the project. Merging.

komali2:
Happy CNY. Is the CI something others such as myself can help with? Seems to be vercel-side and I can't find any CI docs on the subject

examples/docker-compose/docker-compose.yaml
```diff
@@ -140,8 +145,16 @@ services:
     volumes:
       - ./data/mailhog:/maildir
   dashboard:
-    image: nhost/dashboard:0.7.4
-    ports:
-      - "3030:3000"
+    image: nhost/dashboard:0.21.1
+    env_file: .env
```
dbarrosop, line 149: please, provide a working `.env` file and make sure `pnpm e2e`. Thanks!

dbarrosop:
The issue is that `pnpm e2e` fails. I dropped a comment. Thanks and sorry this fell through the cracks.

examples/docker-compose/.env.example
```diff
@@ -1,5 +1,31 @@
...
+PROXY_HOST=proxy.my-nhost.com
+# Change to http://${PROXY_HOST}:1337 if using localhost
+PROXY_URL=https://${PROXY_HOST}
+
+# Public (if making your Nhost dashboard publicly accessible) or private (change to localhost) hostname for the Nhost dashboard for the proxy to allow requests from
+NHOST_HOST=example-url.my-nhost.com
+
+# Environment variables for the Nhost Dashboard. See dashboard/.env.example for default values. The defualt values are changed to work with our traefik setup created with the docker-compose
+
+# URL for the migrations API running from running `hasura-console` locally. It runs on port 9693 so change to http://localhost:9693 if you are running a private instance
```
komali2, line 21: It seems the default port for `hasura console` is 9695, not 9693, is that unique to my system?

lectrician1:
Should the pnpm lock be committed or gitignored or just not added?
image

examples/docker-compose/docker-compose.yaml
lectrician1: current .env.sample works.

lectrician1:
well e2e works on my system. looks like problem with github actions

examples/docker-compose/docker-compose.yaml
dbarrosop, line 149:
```suggestion
    env_file: .env.example
```

examples/docker-compose/.env.example
```diff
@@ -1,5 +1,31 @@
+# Change all these variables before first creating the containers
 POSTGRES_PASSWORD=secret-pg-password-never-use-this-value
-HASURA_GRAPHQL_ADMIN_SECRET=nhost-admin-secret
-HASURA_GRAPHQL_JWT_SECRET='{"type":"HS256", "key":"5152fa850c02dc222631cca898ed1485821a70912a6e3649c49076912daa3b62182ba013315915d64f40cddfbb8b58eb5bd11ba225336a6af45bbae07ca873f3","issuer":"hasura-auth"}'
+HASURA_GRAPHQL_ADMIN_SECRET=change-me
+# The jwt secret key can be generated with `openssl rand -hex 32`
+HASURA_GRAPHQL_JWT_SECRET='{"type":"HS256", "key":"change-this-to-different-32-char-value","issuer":"hasura-auth"}'
```
dbarrosop, line 5: this needs to be reverted or the CI won't work

dbarrosop:
> well e2e works on my system. looks like problem with github actions

It doesn't work for me though. Just checking out your branch and running `pnpm e2e` leads to various issues:

1. First the `.env` file reference in the docker-compose file doesn't exist.
2. Changing the reference to `.env.example` leads to issues with the jwt secret being too short
3. After fixing that things still don't work, I suspect due to the values of `NHOST_HOST`, `PROXY_HOST` and `PROXY_URL`.

lectrician1:
> First the .env file reference in the docker-compose file doesn't exist.

We expect the user to copy the `.env` file from the `.env.example` per the docker-compose README.

The e2e tests currently run/don't run on my computer because the `afterAll` test in `docker-compose.test.ts` is called right before the `it()` tests and runs at the same time as the `beforeAll`. This leads is to complete before the `it()` tests leading to the `graphql-engine` container going down/becoming unhealthy before all the `it()` tests complete.

Commenting 
```js
afterAll(async () => {
      // * Stop docker compose
      await promisifiedExec('docker compose -f docker-compose.yaml --env-file .env.example down')
      }, 5 * 60 * 1000)
```

allows the tests to pass. `afterAll` seems like it's only supposed to run after all the tests have executed but it's currently not doing that.

dbarrosop:
> allows the tests to pass. afterAll seems like it's only supposed to run after all the tests have executed but it's currently not doing that.

`afterAll` runs after all the tests have run. If that's not the case for you it most likely means `beforeAll` failed and the tests are being skipped. You can verify on the branch:

```
(main) ❯ pnpm e2e

> @nhost-examples/docker-compose@0.0.7 e2e /Users/dbarroso/workspace/nhost/nhost/examples/docker-compose
> vitest run


 RUN  v0.32.4 /Users/dbarroso/workspace/nhost/nhost/examples/docker-compose

 ✓ test/docker-compose.test.ts (5) 72454ms
   ✓ docker-compose should start, work and stop (5) 72454ms
     ✓ Hasura
     ✓ Hasura auth (retry x44)
     ✓ Hasura storage
     ✓ Serverless functions
     ✓ Dashboard

 Test Files  1 passed (1)
      Tests  5 passed (5)
   Start at  08:02:07
   Duration  72.66s (transform 23ms, setup 0ms, collect 29ms, tests 72.45s, environment 0ms, prepare 43ms)
```

However, on your branch things don't work for the reasons I explained here:

https://github.com/nhost/nhost/pull/2397#issuecomment-1938176555

dbarrosop:
Ok, with the suggestion I just added to your `docker-compose.yaml` `pnpm e2e` works fine without any change.

lectrician1:
Sooo the current code is good?

dbarrosop:
Yes, if you accept my suggestion I will do a last check to make sure `pnpm e2e` works and if it does I will merge it. Thanks!

dbarrosop:
Great job, thanks a lot for the contribution.

dbarrosop (approved):

PR #2450, opened 2024-01-04
remove migration step from ai guide

dbarrosop:


onehassan (approved):
should we add a changeset as well ?

nunopato (approved):

PR #2506, opened 2024-02-01
fix (docs): indicate that custom domains for postgres doesn't require configuration

dbarrosop:


docs/platform/custom-domains.mdx
```diff
@@ -34,14 +34,13 @@ fqdn = ['hasura.custom-domain.com']
...
 
-[[postgres.resources.networking.ingresses]]
-fqdn = ['postgres.custom-domain.com']
-
 [[functions.resources.networking.ingresses]]
 fqdn = ['functions.custom-domain.com']
 ```
 
-For Run services, typically in nhost-service.toml specific to the service:
+For PostgreSQL there is nothing to configure on Nhost's side so as long as you have a CNAME properly configured it should work.
```
quentin-decre, line 41: I do not understand this. How do you know that postgres.develop.jarvi.tech should point to postgres service ?

docs/platform/custom-domains.mdx
```diff
@@ -34,14 +34,13 @@ fqdn = ['hasura.custom-domain.com']
...
+
+For Run services, typically in an `nhost-service.toml` specific to the service:
```
quentin-decre, line 44: Also maybe, in the doc, a tab or link redirecting to how to setup custom domains in the the sdk would be useful

docs/platform/custom-domains.mdx
```diff
@@ -34,14 +34,13 @@ fqdn = ['hasura.custom-domain.com']
 [[auth.resources.networking.ingresses]]
 fqdn = ['auth.custom-domain.com']
 
-[[postgres.resources.networking.ingresses]]
-fqdn = ['postgres.custom-domain.com']
-
 [[functions.resources.networking.ingresses]]
 fqdn = ['functions.custom-domain.com']
 ```
```
quentin-decre, line 39: What about storage ?

docs/platform/custom-domains.mdx
dbarrosop: We don't care. As long as you connect to the right IP (which will depend on your correct CNAME configuration) and you have the correct credentials and database you are good. With the rest of the services we need to know your domain or choice to add it to the load balancer and to generate TLS certificates but here that's not needed.

nunopato (approved):

PR #2507, opened 2024-02-01
feat: hasura-storage-js: add authenticated download of files

nunopato:


packages/hasura-storage-js/src/hasura-storage-api.ts
```diff
@@ -92,6 +95,38 @@ export class HasuraStorageApi {
     return { fileMetadata, error: null }
   }
 
+  async downloadFile(params: StorageDownloadFileParams): Promise<StorageDownloadFileResponse> {
+    try {
+      const { fileId, range, width: w, height: h, quality: q, blur: b } = params
+
+      const headers = this.generateAuthHeaders()
+      if (range) headers.append('Range', range)
```
dbarrosop, line 103: maybe it's better to just allow extra headers for the request, that way people can pass other headers related to etags, if-modified, etc... which are also supported.

nunopato: updated to allow any headers

dbarrosop (approved):

onehassan (approved):

Issue #2348, opened 2023-10-31
Next JWT token expired when using SSR


AlexanderMoroz:
So after discussion with @dbarrosop in discord help channel I've decided to open an Issue here

**Describe the bug**
I'm investigation integration next + nhost with SSR enabled (pages router). I've setup our app the same way as in examples - <NhostProvider> + <NhostApolloProvider> + getNhostSession from nhost/nextjs

In general it works really well! Thanks guys for all your work. 

But sometimes (and it's even not clear when and why) I recieve errors about JWT token being expired, both on client and server. Previously we were using next but without getNhostSessions so all auth part was working only on client and it was working flawlessly

error: ApolloError: Could not verify JWT: JWTExpired
**To Reproduce**
1. Go to start page.
2. Wait till refresh token expired

**Expected behavior**
Token updates automatically

**Screenshots**

**Desktop (please complete the following information):**
 - OS: macOS 14.0 (23A344)
 - Browser: Chrome: 118.0.5993.117 
 - Versions: 
   -  "next": "13.3.0",
   -  "@nhost/apollo": "^5.2.19",
   - "@nhost/nextjs": "^1.13.33",
   -  "@nhost/nhost-js": "^2.2.12",
   -  "@nhost/react": "^2.0.27",
   -  "@nhost/react-apollo": "^5.0.36",

**Code**

ApptProvider in `_app.tsx`
```tsx
export const Provider: FC<ProviderProps> = ({
  children,
  nhostSession,
  transitionConfig = {
    type: 'spring',
    bounce: 0.01,
  },
}) => {
  return (
    <NhostProvider nhost={nhost} initial={nhostSession}>
      <ApolloProvider>
        <MotionConfig transition={transitionConfig}>{children}</MotionConfig>
      </ApolloProvider>
    </NhostProvider>
  );
};
```

ApolloProvider used above:
```tsx
import { FC, PropsWithChildren } from 'react';
import { NhostApolloProvider } from '@nhost/react-apollo';
import { nhost } from '@htch/shared-services/nhost';
import { NhostApolloClientOptions } from '@nhost/apollo';

export const ApolloProvider: FC<
  PropsWithChildren<Omit<NhostApolloClientOptions, 'nhost'>>
> = ({ children, ...props }) => {
  return (
    <NhostApolloProvider nhost={nhost} {...props}>
      {children}
    </NhostApolloProvider>
  );
};
```

getNhostSession:
```
[...]

PR #2533, opened 2024-02-12
fix: JWT expired bug

onehassan:
fixes: https://github.com/nhost/nhost/issues/2348


packages/hasura-auth-js/src/machines/authentication/machine.ts
```diff
@@ -687,9 +687,16 @@ export const createAuthMachine = ({
         isAutoRefreshDisabled: () => !autoRefreshToken,
         refreshTimerShouldRefresh: (ctx) => {
           const { expiresAt } = ctx.accessToken
+
           if (!expiresAt) {
             return false
           }
+
+          // This happens when either the computer goes to sleep or when Chrome descides to suspend the tab
+          if (expiresAt.getTime() < Date.now()) {
```
dbarrosop, line 696: I think we should have a margin, i.e., 60s. Otherwise, you can end in a situation where the token was valid in this line but on the next one it has expired.

PR #2564, opened 2024-02-27
chore (examples/docker-compose): clarification on greyed-out options in the dashboard when self-hosting

lectrician1:
Should answer questions like these: https://discord.com/channels/552499021260914688/1212064280170463243

### DIFF
--- a/.changeset/afraid-ducks-hang.md
+++ b/.changeset/afraid-ducks-hang.md
@@ -2,4 +2,4 @@
 '@nhost-examples/docker-compose': minor
 ---
 
-chore: clarification on greyed-out options in the dashboard when self-hosting```
+chore: clarification on greyed-out options in the dashboard when self-hosting
--- a/.changeset/afraid-ducks-hang.md
+++ b/.changeset/afraid-ducks-hang.md
@@ -2,4 +2,4 @@
 '@nhost-examples/docker-compose': minor
 ---
 
-Note some features not available with docker-compose self hosted
+chore: clarification on greyed-out options in the dashboard when self-hosting```
--- a/.changeset/afraid-ducks-hang.md
+++ b/.changeset/afraid-ducks-hang.md
@@ -1,0 +1,5 @@
+---
+'@nhost-examples/docker-compose': minor
+---
+
+Note some features not available with docker-compose self hosted
--- a/examples/docker-compose/README.md
+++ b/examples/docker-compose/README.md
@@ -2,6 +2,8 @@
 
 Here is an example on how to reproduce the Nhost stack from a docker-compose file.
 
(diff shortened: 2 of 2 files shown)


examples/docker-compose/README.md
```diff
@@ -2,6 +2,8 @@
 
 Here is an example on how to reproduce the Nhost stack from a docker-compose file.
 
+Note that the Nhost instance produced will not allow you to do all the things available on Nhost Cloud. Some tabs like Settings may be greyed-out and you will need to configure them using envinoment variables or your own system. Additionally, the [Nhost CLI](https://github.com/nhost/cli) is [unable to be used](https://github.com/nhost/cli/issues/837#issuecomment-1963499492) with self-hosted instances.
```
dbarrosop, line 5: I feel the current wording mischaracterizes a bit the issue and some people may be led to believe we are criplling the self-hosted product, which isn't the case. Maybe a better wording could be:

```suggestion
NOTE: You may notice that some options in the dashboard are greyed-out. Those options are extra services provided by the Nhost Cloud and, thus, not available when self-hosting.
```

lectrician1: Could the note about the CLI not being available be kept?

dbarrosop: I don't think that comment is accurate. The CLI is a development tool. You are free to use it whether you are self-hosting or not. Obviously, you are going to have implement your own deployment pipelines and configuration management system that converts the output of the CLI to your self-hosted instance but there is nothing preventing you from using it.

lectrician1: Well I would believe the work involved to get your own system configured to correctly interface with the CLI would be quite the amount of work when nobody has done it before... So I don't think it's not fair to say that the CLI is "compatible".

.changeset/afraid-ducks-hang.md
```diff
@@ -0,0 +1,5 @@
+---
+'@nhost-examples/docker-compose': minor
+---
+
+Note some features not available with docker-compose self hosted
```
dbarrosop, line 5:
```suggestion
chore: clarification on greyed-out options in the dashboard when self-hosting```

File: .changeset/afraid-ducks-hang.md [hunk 3]
```diff
@@ -0,0 +1,5 @@
+---
+'@nhost-examples/docker-compose': minor
+---
+
+chore: clarification on greyed-out options in the dashboard when self-hosting```
```
dbarrosop, line 5:
```suggestion
chore: clarification on greyed-out options in the dashboard when self-hosting
```

dbarrosop:
Thanks! Merging!

dbarrosop (approved):

PR #2615, opened 2024-03-15
feat: react-apollo: add example of how to connect github to an existing account

onehassan:
fixes https://github.com/nhost/nhost/issues/2582


dbarrosop (approved):

examples/react-apollo/src/profile/connect-socials.tsx
```diff
@@ -0,0 +1,55 @@
...
+  `
+
+  const { data } = useAuthQuery<{
+    authUserProviders: AuthUserProviders[]
+  }>(AUTH_USER_PROVIDERS, {
+    pollInterval: 5000,
+    fetchPolicy: 'cache-and-network'
+  })
+
+  const connectToGithubLink = `${github}&connect=${jwt}`
```
dbarrosop, line 33: shouldn't we have something in the SDK for this? I.e. instead of `nhost.auth.signin({github})` something like `nhost.auth.connect_social({github})`

packages/hasura-auth-js/src/hasura-auth-client.ts
```diff
@@ -132,6 +133,45 @@ export class HasuraAuthClient {
...
+   *    redirectTo: window.location.href
+   *   }
+   * })
+   * ```
+   *
+   * @docs https://docs.nhost.io/reference/javascript/auth/connect-provider
+   */
+  async connectProvider(
+    params: SignInWithProviderParams
+  ): Promise<SignInResponse & { providerUrl?: string; provider?: string }> {
```
dbarrosop, line 154: should we change the response type? I am going to change the backend to just respond with "OK" and given that you are hardcoding the response `session: null, mfa: null` it feels like we need a different type

packages/hasura-auth-js/src/types/options.ts
```diff
@@ -48,7 +48,9 @@ export interface DeanonymizeOptions extends RegistrationOptions {
   password?: string
 }
 
-export interface CommonProviderOptions extends RegistrationOptions, RedirectOption {}
+export interface CommonProviderOptions extends RegistrationOptions, RedirectOption {
+  connect?: boolean
```
dbarrosop, line 52: should this be added to CommonProviderOptions? It is only used in the connect endpoint

dbarrosop (approved):

Issue #2696, opened 2024-05-09
add "headers" option to missing methods in the js sdk


dbarrosop:
This is mostly relevant to functions, storage and graphql SDKs.

Check which methods are missing the headers option and add it.

Bonus: if possible, maybe add a global `setHeaders` so you can set global headers for the client.

PR #2697, opened 2024-05-09
feat: sdk: add `setHeaders` for setting global headers and make sure all client calls have a `headers` arg

onehassan:
resolves https://github.com/nhost/nhost/issues/2696

### DIFF
--- a/packages/nhost-js/src/clients/nhost.ts
+++ b/packages/nhost-js/src/clients/nhost.ts
@@ -90,4 +90,20 @@ export class NhostClient {
     // this.functions.setAdminSecret(newValue)
     // this.graphql.setAdminSecret(newValue)
   }
+
+  /**
+   * Use `nhost.setRole` to set the user role for all subsequent graphql, storage and functions calls
+   *
+   * @example
+   * ```ts
+   * nhost.setRole('admin')
+   * ```
+   *
+   * @docs https://docs.nhost.io/reference/javascript/set-role
+   */
+  setRole(role: string) {
+    this.graphql.setHeaders({ 'x-hasura-role': role })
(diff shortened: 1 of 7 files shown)


dbarrosop (approved):

packages/nhost-js/src/clients/nhost.ts
```diff
@@ -90,4 +90,20 @@ export class NhostClient {
...
+   *
+   * @example
+   * ```ts
+   * nhost.setRole('admin')
+   * ```
+   *
+   * @docs https://docs.nhost.io/reference/javascript/set-role
+   */
+  setRole(role: string) {
+    this.graphql.setHeaders({ 'x-hasura-role': role })
```
dbarrosop, line 105: what happens if the user does the following?

```
n.setRole("my-role")
n.storage.setHeaders({"x-something": "blah"}) 
```

will the role still be sent? It won't, will it? We should :)

onehassan: Yes, good catch 👍 - will adjust

packages/nhost-js/src/clients/nhost.ts
```diff
@@ -90,4 +90,38 @@ export class NhostClient {
...
+    this.storage.setHeaders({ 'x-hasura-role': role })
+    this.functions.setHeaders({ 'x-hasura-role': role })
+  }
+
+  /**
```
dbarrosop, line 110: similar comment to above

xmlking:
@onehassan   why do we need to pass argument for unsetRole() function ? `e.g., nhost.unsetRole('admin')`
Is it just removing `x-hasura-role` header?

dbarrosop:
@xmlking copy&paste mistake, will fix

PR #2716, opened 2024-05-16
fix(hasura-auth-js): resolve social sign-in issue in React Native

onehassan:
fix https://github.com/nhost/nhost/issues/2216

### DIFF
--- a/.changeset/plenty-beers-warn.md
+++ b/.changeset/plenty-beers-warn.md
@@ -2,4 +2,4 @@
 '@nhost/hasura-auth-js': patch
 ---
 
-fix: resolve social sign-in issue in React Native
+fix: improve environment/browser detection to support React Native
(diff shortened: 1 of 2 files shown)


dbarrosop (approved):

dbarrosop:
I think a better changelog would be "improve environment/browser detect to include react-native"

onehassan:
> improve environment/browser detect to include react-native

Thanks will apply

hadnet:
So now oauth is now working using React Native?

dbarrosop:
it will when we release this fix. We are finishing up an example/documentation as we speak so stay tuned :)

hadnet:
> it will when we release this fix. We are finishing up an example/documentation as we speak so stay tuned :)

Any updates on this?

PR #2718, opened 2024-05-20
fix: update `GetAllWorkspacesAndProjects` query polling to use exp backoff

onehassan:


dashboard/src/pages/index.tsx
```diff
@@ -7,41 +7,55 @@ import { Button } from '@/components/ui/v2/Button';
...
+  // keep polling for workspaces until there is a workspace available.
+  // We do this because when a user signs up a workspace is created automatically
+  // and the serverless function can take some time to complete.
+  const { data, startPolling, stopPolling, networkStatus } =
     useGetAllWorkspacesAndProjectsQuery({
       skip: !user,
+      notifyOnNetworkStatusChange: true,
+      onError: () => {
+        // When there's an error (graphql, network error) apply an exponential backoff strategy
+        setPollInterval((prevInterval) => {
```
dbarrosop, line 33: what is the initial value for `prevInterval`?

dbarrosop: and how does it get updated?

onehassan: the initial value (1s) is defined on L22 `const [, setPollInterval] = useState(1_000)`
`useState` [allows me to pass a function to have access to the previous state](https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state), and the value we return from that functions is the new value: `return newInterval` L36

dbarrosop (approved):

PR #2725, opened 2024-05-24
feat (dashboard): postgress major and minor verions

idsintehittapa:


dashboard/src/components/form/ControlledAutocomplete/ControlledAutocomplete.tsx
```diff
@@ -40,21 +40,26 @@ function ControlledAutocomplete(
   const form = useFormContext();
   const { field } = useController({
     ...(controllerProps || {}),
-    name: controllerProps?.name || name || '',
+    name: controllerProps?.name || name || '' || 'version',
     control: controllerProps?.control || control,
+    defaultValue: { label: '', value: '' },
```
idsintehittapa, line 45: I had to add this otherwise the `field.value` would be undefined and we would get a warning going from a uncontrolled to a controlled state

PR #2735, opened 2024-05-30
fix (dashboard): validate postgres version in ai service settings page

### LINKED ISSUE #2733: ai service settings page fails to validate postgres 14.11 version
if you are running postgres version `14.11-20240515-1` and try to enable the AI service in the settings page it will fail with error:

> In order to enable the AI service you need to update your database version to 14.6-20231018-1 or newer.

We should do this check only if the postgres image starts with `14.6`, any other version (14.11, 15.x, 16.y) should pass the check automatically.

dbm03:
fixes #2733

### DIFF
--- a/dashboard/src/features/ai/settings/utils/isPostgresVersionValidForAI/isPostgresVersionValidForAI.ts
+++ b/dashboard/src/features/ai/settings/utils/isPostgresVersionValidForAI/isPostgresVersionValidForAI.ts
@@ -10,9 +10,9 @@ export default function isPostgresVersionValidForAI(
   const MIN_POSTGRES_VERSION_SUPPORTING_AI = '14.6-20231018-1';
 
   if (/^14\.6-/.test(postgresVersion)) {
-     return postgresVersion >= MIN_POSTGRES_VERSION_SUPPORTING_AI
+    return postgresVersion >= MIN_POSTGRES_VERSION_SUPPORTING_AI;
   }
-  
+
   // Note: No need to account for versions less than 14.6
-  return true
+  return true;
 }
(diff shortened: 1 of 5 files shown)


dashboard/src/features/ai/settings/utils/isPostgresVersionValidForAI/isPostgresVersionValidForAI.ts
```diff
@@ -0,0 +1,38 @@
+/**
+ * Check if the given postgres version is valid for enabling AI in the project
+ *
+ * @param postgresVersion - Postgres version used in the project.
+ * @returns Whether is valid for enabling AI.
+ */
+export default function isPostgresVersionValidForAI(
```
dbarrosop, line 7: can this be simplified to:

```
const MIN_POSTGRES_VERSION_SUPPORTING_AI = '14.6-20231018-1';
const majorMinorSupportedVersion = MIN_POSTGRES_VERSION_SUPPORTING_AI.split('-')

const [majorMinorVersion] = postgresVersion.split('-')

if (postgresMinor === MIN_POSTGRES_MINOR_VERSION_SUPPORTING_AI) {
   return postgresVersion >= MIN_POSTGRES_VERSION_SUPPORTING_AI
}
return true
```

dbarrosop: > can clients use a postgres version lower than 14.6-xxxx?

no, `14.6` is the minimum they can use

onehassan (approved):

dbarrosop (approved):

PR #2780, opened 2024-07-03
fix (dashboard): improve project health error handling, add poll interval

### LINKED ISSUE #2776: dashboard: overview project health modal doesn't show error info
Service state modal has a transient state where it might show the error `undefined is not an object (evaluating r.length)` instead of the expected accordion for each service

dbm03:
Fixes #2776


dashboard/src/features/projects/common/hooks/useServiceStatus/useServiceStatus.ts
```diff
@@ -0,0 +1,100 @@
...
+      skip: !isPlatform || !currentProject,
+      pollInterval: 0,
+      skipPollAttempt: () => !isVisible,
+    });
+
+  // Fetch when mounted
+  useEffect(() => {
+    refetch();
+    return () => {
+      stopPolling();
```
onehassan, line 43: I think this is redundant here, see Line 77

dbm03: Not sure, if there's no Line 77, then useEffect wouldn't stop polling after cleanup (when `options.shouldPoll=true`), right?

Maybe it's clearer to rewrite as:
```
useEffect(() => {
    if (options.shouldPoll) {
      if (
        [auth, storage, postgres, hasura, ai, ...Object.values(run)].some(
          (service) => service?.state !== ServiceState.Running,
        )
      ) {
        // Poll more frequently if a service is not running
        startPolling(options.pollInterval || 10000);
      } else {
        // Poll less frequently when all services are running
        startPolling(60000);
      }
    }
    

    return () => stopPolling();
  }
  ```

onehassan: in that case the cleanup will be done on L43 right — That will be executed when the component unmounts anyway

dbm03: Oh I see your point now, I'll update it :+1:

onehassan (approved):
LGTM

PR #2820, opened 2024-08-12
feat (docs): added subdomain/region information

dbarrosop:


docs/guides/cli/subdomain.mdx
```diff
@@ -0,0 +1,95 @@
...
+<Warning>
+If you are trying to connect to your local environment from an external device or VM make sure that:
+
+- The IP address you are using is reachable from this device/VM
+- That your firewall isn't blocking requests
+</Warning>
+
+<Warning>
+If you are testing a social provider don't forget you will need to configure the callback URL to match the subdomain/region you are using.
+</Warning>
```
onehassan, line 74: maybe it would be useful to show an example of the callback URL like so
`https://192-168-1-108.auth.local.nhost.run/v1/signin/provider/google/callback`

onehassan (commented):
My only concern here is that we are removing the details where we a user needs to use `10.0.2.2` for Android and/or figure out their ip address to use for iOS when running on a  physical device

docs/guides/cli/subdomain.mdx
dbarrosop: I think that will just clutter the docs and will probably lead to some users copy and pasting (we already had an example yesterday). A reminder will probably make them have to think about it rather than follow blindly and paste the wrong thing.

dbarrosop:
> My only concern here is that we are removing the details where we a user needs to use 10.0.2.2 for Android and/or figure out their ip address to use for iOS when running on a physical device

I think that's preferable. If we need that kind of information that should be in the react native guide or similar (which maybe we need to update?) but I don't think we should have the dnsmasq page or a dedicated page for figuring it out the IP of people's devices.

docs/guides/cli/subdomain.mdx
dbarrosop: https://github.com/nhost/nhost/issues/2821

onehassan: sounds good 👌

onehassan (approved):

PR #2988, opened 2024-11-01
fix (dashboard): configuring "Disable New Users" modifies the wrong setting

### LINKED ISSUE #2912: dashboard: configuring "Disable New Users" modifies the wrong setting
Under settings -> authentication we have the setting "Disable new Users" which seems to disable sign ups rather than the `disableNewUsers` setting. We should fix this and add disable sign ups now that we are in the process as I think we don't have it.

dbm03:
Resolves #2912

### DIFF
--- a/dashboard/src/features/orgs/projects/authentication/settings/components/DisableSignUpsSettings/DisableSignUpsSettings.tsx
+++ b/dashboard/src/features/orgs/projects/authentication/settings/components/DisableSignUpsSettings/DisableSignUpsSettings.tsx
@@ -105,10 +105,10 @@ export default function DisableSignUpsSettings() {
         }
       },
       {
-        loadingMessage: 'Disabling new user sign ups...',
-        successMessage: 'New user sign ups have been disabled successfully.',
+        loadingMessage: 'Disabling new users sign ups...',
+        successMessage: 'New users sign ups have been disabled successfully.',
         errorMessage:
-          'An error occurred while trying to disable new user sign ups.',
+          'An error occurred while trying to disable new users sign ups.',
       },
     );
   };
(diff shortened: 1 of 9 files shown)


dashboard/src/features/orgs/projects/authentication/settings/components/DisableSignUpsSettings/DisableSignUpsSettings.tsx
```diff
@@ -0,0 +1,136 @@
...
+export type DisableSignUpsFormValues = Yup.InferType<typeof validationSchema>;
+
+export default function DisableSignUpsSettings() {
+  const { openDialog } = useDialog();
+  const isPlatform = useIsPlatform();
+  const { maintenanceActive } = useUI();
+  const localMimirClient = useLocalMimirClient();
+  const { project } = useProject();
+  const [updateConfig] = useUpdateConfigMutation({
+    refetchQueries: [GetAuthenticationSettingsDocument],
```
onehassan, line 34:
```suggestion
```

can we remove this as this will trigger a reload - we should rely on either res

onehassan (commented):
left a comment

dashboard/src/features/orgs/projects/authentication/settings/components/DisableSignUpsSettings/DisableSignUpsSettings.tsx
dbm03: On it

dbarrosop:
There is something wrong in this PR, the setting doesn't seem to work. I have the following config in a test project:

```toml
[auth.signUp]
disableNewUsers = false
enabled = true
```

and the dashboard shows:

Screenshot 2024-11-06 at 09 23 25

dbarrosop (approved):

onehassan (approved):

PR #3139, opened 2025-01-16
fix (dashboard): undefined is not an object in logs page

dbm03:


dashboard/src/pages/orgs/[orgSlug]/projects/[appSubdomain]/logs.tsx
```diff
@@ -102,6 +109,10 @@ export default function LogsPage() {
   );
 
   useEffect(() => {
+    if (!project || loadingProject) {
+      return () => {};
```
Comment (robertkasza) [inline dashboard/src/pages/orgs/[orgSlug]/projects/[appSubdomain]/logs.tsx:113]: This clean up function does nothing. I do not know why this useEffect needs an empty clean up function, since they do nothing we should remove them.
Could you remove them and simply use return;? (The last return(line 135) is not needed at all.)

robertkasza (approved):

PR #3145, opened 2025-01-19
feat(dashboard): filter files by id

### LINKED ISSUE #2777: dashboard: Storage dashboard should allow filtering by ID
A lot of the time, I want to use the dashboard to find specific storage files for debugging or other purposes. The dashboard has a "filter & search" field but it seems to only search names. I had hoped/expected it could filter or search by ID (or other fields, but ID is critical).

Add a filter for ID or at least allow the current search field to search IDs.

**Describe alternatives you've considered**
No other alternatives other than making calls to the storage API manually, which negates the purpose of the dashboard for easy data viewing.

**Additional context**
Screenshot 2024-07-02 at 1 51 34 PM

antoinecaputo:
From the Storage page, this allows finding a specific file by its ID.

As it's a UUID type, we can't run a comparison until the input is in a valid UUID format.
Otherwise, it will fail with `invalid input syntax for type uuid`.

**Filtering by name:**
filtering-by-name

**Filtering by ID:**
filtering-by-id

Closes #2777


dbarrosop:
Hello,
thanks for the PR very much appreciated, I just tested it and works as advertised but I think the behavior needs a bit of tweaking:

1. I think we should rephrase "Filter and search..." with "Filter and search by file ID..." so it is clear only ID is supported. Otherwise people would be tempted to try adding a content type or some metadata
2. If I just type `x` all content disappears, I think we shouldn't query the backend unless the value in the input form is a valid uuid. We should validate the input to make sure it is a valid uuid, show  and only then query the backend. Alternatively, add a "search" button and query only when the user presses "enter" or clicks on the button.


Thanks again for the PR.

dbarrosop:
Ok, after looking at this properly caffeinated LGTM. Thanks a lot for the contribution.

dbarrosop (approved):

PR #3193, opened 2025-02-07
fix (dashboard): regex search with service filter

### LINKED ISSUE #3154: dashboard: Logs page, can't hit search when you have already selected a filter by service type
**To Reproduce**
1. Go to Logs
2. Click on `All Services` filter
3. Select `Backups`
4. Type a regex and clicking `Search` doesn't update anything

dbm03:
Fixes #3154


dashboard/src/features/orgs/projects/logs/components/LogsHeader/LogsHeader.tsx
```diff
@@ -49,35 +52,23 @@ export default function LogsHeader({
...
-
   const { data, loading: loadingServiceLabelValues } =
     useGetServiceLabelValuesQuery({
       variables: { appID: project?.id },
       skip: !project?.id,
     });
 
-  useEffect(() => {
+  const serviceLabels = useMemo<{ label: string; value: string }[]>(() => {
     if (!loadingServiceLabelValues && data) {
```
robertkasza, line 62: I think the The loading is not needed at all. If there is data we create the labels based on the LOGS_SERVICE_TO_LABEL and data and we return an empty array if there is no data yet. If you want it to be a little bit more readable than this : `data?.getServiceLabelValues ?? []`

We could do something like this: 

```
const serviceOptions = useMemo(() => {
    // use isEmptyValue to make it more readable
    if (isEmptyValue(data)) {
      return [];
    }
    const options = [
      {
        label: LOGS_SERVICE_TO_LABEL[AvailableLogsService.ALL],
        value: AvailableLogsService.ALL,
      },
      ...data.getServiceLabelValues.map((l) => ({
        label: LOGS_SERVICE_TO_LABEL[l] ?? l,
        value: l,
      })),
    ];
    // You need to move the whole part form line 118 to here if you'd like to use useMemo 
    return options.map(({ value, label }) => (
      <Option key={value} value={value} className="text-sm+ font-medium">
        {label}
      </Option>
    ));
  }, [data]);

(...)

```
If you'd like to use useMemo you need serviceOptions to be only change when data is changing. The way it is used in line 118 makes it that every render a new array is created because of the spread operator and the map function. Basically canceling out the useMemo.
And use it like this in the jsx: {serviceOptions}

Also the ControlledSelect component needs to be used with React.memo otherwise the component is rerendered every time.

dbm03: I updated to use `!Array.isArray(data?.getServiceLabelValues)` and the memoization for `serviceOptions`.

Thanks for the suggestions!

dashboard/src/features/orgs/projects/logs/components/LogsHeader/LogsHeader.tsx
```diff
@@ -49,36 +52,33 @@ export default function LogsHeader({
...
-        grafana: 'Grafana',
-        'job-backup': 'Backup Jobs',
-        ai: 'AI',
-      };
-
-      setServiceLabels(
-        labels.map((l) => ({ label: labelMappings[l] ?? l, value: l })),
-      );
+  const serviceOptions = useMemo<JSX.Element[]>(() => {
+    if (!Array.isArray(data?.getServiceLabelValues)) {
```
robertkasza, line 61: In my opinion these type of checks are hard to read. You need to notice that data can be **undefined** and if it is not undefined that it might have a getServiceLabelValues prop which could be an array or something else we do not know, but the type suggests that it can only be an array of strings. And you still need to negate the whole thing because of the **exclamation mark**.

That is why I suggested the isEmptyValue(might have to rename it to isEmpty) function. I think it is easier to understand, because it is closer to a sentence, than the other one. 

If data is empty then do this....  |  if(isEmpty(data)) {...} These two are very similar and might be easier to understand what the code does.

Also we need this check because the data object is  initially **undefined** if we are not sure that the getServiceLabelValues than we can check that too, but it should not be necessary I think it will be an empty array if there are no services.

A simpler and more readable code can lead to less defects and easier to maintain. This is just my opinion you can take it or leave it. I can live with this check too. :)

robertkasza (approved):

PR #3210, opened 2025-02-19
fix (dashboard): can change minor database version with unhealthy project

### LINKED ISSUE #3174: dashboard: minor database version change shouldn't be blocked by project's health

dbm03:
Fixes #3174


dashboard/src/features/orgs/projects/database/settings/components/DatabaseServiceVersionSettings/DatabaseServiceVersionSettings.tsx
```diff
@@ -206,12 +206,16 @@ export default function DatabaseServiceVersionSettings() {
     !applicationPaused &&
     !applicationPausing &&
     !applicationUpdating;
+
   const isMajorVersionDirty = formState?.dirtyFields?.majorVersion;
   const isMinorVersionDirty = formState?.dirtyFields?.minorVersion;
   const isDirty = isMajorVersionDirty || isMinorVersionDirty;
 
-  const versionFieldsDisabled =
+  const majorVersionFieldDisabled =
```
dbarrosop, line 214: interesting, why are we blocking on `applicationUpdating`? it isn't a big deal, just wondering. Also, what is `maintenanceActive`?

dbm03: We're showing this disclaimer when the app status is ApplicationStatus.Updating.

image

Should we allow changing the field even while updating? (the Save button can still be disabled)

`maintenanceActive` is an ENV variable that seems like we're never setting in deployment anymore. I removed it from this file in the last commit

dashboard/src/features/orgs/projects/database/settings/components/DatabaseServiceVersionSettings/DatabaseServiceVersionSettings.tsx
```diff
@@ -206,12 +206,16 @@ export default function DatabaseServiceVersionSettings() {
...
     applicationUpdating || applicationUnhealthy || maintenanceActive;
+  const minorVersionFieldDisabled = applicationUpdating || maintenanceActive;
```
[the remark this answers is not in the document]
dbm03: I updated the minor version restrictions to be consistent with auth/storage pages

dashboard/src/features/orgs/projects/database/settings/components/DatabaseServiceVersionSettings/DatabaseServiceVersionSettings.tsx
dbarrosop: yeah, just remove the warning and allow changing versions. We already have the "Updating" at the top

dbm03: I've removed the warning and allowed to change versions while `ApplicationStatus.Updating | ApplicationStatus.Migrating` in the last commit

dbarrosop (approved):

dbarrosop:
This seems broken, now I can't update the version at all. I did the following:

1. Set some random postgres version, this will cause the change to fail eventually
2. Now I can't do anything, the major version is locked (which is fine) but the minor doesn't do anything, if I select a different version it doesn't unlock the save button.

PR #3247, opened 2025-03-19
chore (dashboard): update PiTR

robertkasza:
- add tests
- add price to PiTR settings
- update link in PiTR settings
- add note with recommendation about restore
- add link to PiTR docs to restore page
- small bug fixes


dashboard/src/features/orgs/projects/backups/components/common/PointInTimeBackupInfo/BackupScheduledInfo.tsx
```diff
@@ -29,9 +15,9 @@ function BackupScheduledInfo({ onClose, orgSlug, subdomain }: Props) {
       <p>Your backup has been scheduled successfully and will start shortly.</p>
       <p>
         To follow its process go to the{' '}
-        <LogsLink href={`/orgs/${orgSlug}/projects/${subdomain}/logs`}>
+        <TextLink href={`/orgs/${orgSlug}/projects/${subdomain}/logs`}>
           Logs page
-        </LogsLink>{' '}
+        </TextLink>{' '}
         and select the service &quot;Backup Job&quot; to see the restore logs.
```
dbm03, line 21: Just a tiny detail, the Logs page selector shows `Backup Jobs` instead of `Backup Job`

`&quot;Backup Jobs&quot;`

dashboard/src/features/orgs/projects/backups/components/ImportBackupTabContent/ImportBackupTabContent.test.tsx
```diff
@@ -0,0 +1,268 @@
...
+    await user.click(await screen.getByText('13'));
+
+    const hoursInput = await screen.getByLabelText('Hours');
+    await user.type(hoursInput, '18');
+
+    const updatedDateTimeButton = await screen.getByRole('button', {
+      name: /UTC/i,
+    });
+    expect(updatedDateTimeButton).toHaveTextContent(
+      '13 Mar 2025, 18:00:05 (UTC+02:00)',
```
dbm03, line 200: I believe this is going to break when the `Europe/Helsinki` timezone switches to Eastern European Summer Time. That is, `Europe/Helsinki` goes from UTC+02:00 to UTC+03:00 but this "absolute" date remains the same.

dbm03 (approved):
Please double check that tests won't break with the european daylight time adjustment. Other than that LGTM!

dashboard/src/features/orgs/projects/backups/components/ImportBackupTabContent/ImportBackupTabContent.test.tsx
robertkasza: Because the date itself is in UTC+02:00 this test should not break when we enter DST. The offset in the button's text is generated with the selected date and not the current date, so it should be okay. If it breaks than it is a bug.

There is a test for checking if the offset is correctly updated when jumping between DST and ST.
https://github.com/nhost/nhost/pull/3247/files#diff-c7076012eb33d6f60049710638b5ad19c2f310b8c250c79f1905be7e0a30b00aR137

Issue #3244, opened 2025-03-18
dashboard: mobile navbar covered by project/orgs navigation


dbm03:
When you open the top right corner hamburger button navbar in a small screen (mobile) it gets covered by the project/orgs navigation sidebar

**Screenshots**

![Image](https://github.com/user-attachments/assets/5dc5e902-9a73-4f49-82ee-eb3c7d249cc3)

PR #3249, opened 2025-03-20
bug: mobile-nav-update-z-index-increase

naaa760:
Fix: Mobile Navigation Menu Z-Index Issue

#3244

image

What
Fixed mobile navigation drawer being covered by project/organization sidebar when opened on small screens.

Why
Users were unable to access mobile menu options as the navigation drawer was appearing underneath the project sidebar, making it inaccessible.

How
Increased the z-index value of the mobile navigation drawer from 39 to 1200 in MobileNav.tsx to ensure proper layering of UI components.

Changes
Updated z-index in MobileNav.tsx from z-[39] to z-[1200]
The mobile menu now correctly displays above all other UI elements
Improved accessibility of navigation options on mobile devices

Testing
1. Open the dashboard on a mobile device or narrow the browser window
2. Click the hamburger menu in the top right corner
3. Verify mobile navigation drawer appears above the project sidebar
Confirm all menu options are accessible and clickable

### DIFF
--- a/dashboard/src/components/layout/MobileNav/MobileNav.tsx
+++ b/dashboard/src/components/layout/MobileNav/MobileNav.tsx
@@ -106,7 +106,7 @@ export default function MobileNav({ className, ...props }: MobileNavProps) {
         anchor="right"
         open={menuOpen}
         onClose={() => setMenuOpen(false)}
-        className="z-[39] w-full sm:hidden"
+        className="z-[1200] w-full sm:hidden"
         hideCloseButton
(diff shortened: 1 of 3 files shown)


dbm03:
Thanks for your contribution! I noticed that in Chrome, the drawer is covering the header bar, which prevents me from closing it.

Ideally, the header navigation should remain visible while only the main container gets hidden when opening the mobile navigation drawer.

Let me know if you need any clarification - thanks again!

PR #3250, opened 2025-03-23
Fix Mobile Navigation Drawer Positioning

naaa760:
>> Description

This PR addresses an issue where the mobile navigation drawer was covering the header bar in Chrome, making it impossible for users to close the drawer or access header navigation elements.

>> Changes Made

- Modified the MobileNav component to position the drawer and its backdrop below the header

- Added proper positioning using both CSS classes and inline styles to ensure cross-browser compatibility

- Set the drawer height to be calculated relative to the header height

- Updated the authentication implementation to use the Nhost client directly instead of the problematic @nhost/next.js 

>> Technical Details

- Used CSS variables with fallbacks (var(--header-height, 64px)) to ensure consistent positioning

- Applied positioning to both the backdrop and drawer paper components

- Implemented a direct signOut function using the Nhost client

### DIFF
--- a/dashboard/src/components/database/TableForm.tsx
+++ b/dashboard/src/components/database/TableForm.tsx
@@ -1,0 +1,64 @@
+import { useState } from 'react';
+
+// PostgreSQL has a limit of 63 bytes for identifiers
+const MAX_TABLE_NAME_LENGTH = 63;
+
+const TableForm = ({ onSubmit, initialValues, ...props }) => {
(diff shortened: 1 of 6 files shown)


dbarrosop:
Hello, thanks for the contribution. Is this a duplicate of #3249 ?

dbm03:
Closing as this was solved in #3341. Thank you for your contribution

Issue #3273, opened 2025-03-31
chores: Upgrade package versions, configuration


robertkasza:
- ~~Upgrade **typescript** to latest version~~
-  ~~Make the [prettierrc.js](https://github.com/nhost/nhost/blob/main/.prettierrc.js) the default config for prettier in the whole repo~~ 
    - ~~compare dashboard's and the root's config~~
    - ~~add `prettier-plugin-organize-imports` to the root config~~
    - ~~extend the default config if its needed e.g: dashboard- > `prettier-plugin-tailwindcss`~~
    - ~~un the format on the whole repo~~
    - ~~make sure that `prettier-plugin-tailwindcss` and `prettier-plugin-organize-imports` is not installed twice (root and dashboard)~~
    - ~~update **changeset** to latest version and use prettier false config~~
- ~~Upgrade **eslint** to latest version~~ 
- use biome.js for linting and formatting

PR #3278, opened 2025-04-01
upgraded packages

Ankitkr5858:
https://github.com/nhost/nhost/issues/3273


dbarrosop:
Hello,
thanks for the contribution, unfortunately this task is a bit more complex/involved as this will require updating plenty of places in the codebase or upgrading the package will break things. I will be closing this PR as this is a large task that has many parts and I think is best if we tackle it ourselves. Thanks again and sorry.

Regards

PR #3299, opened 2025-04-10
feat (dashboard): remote schemas

dbm03:


dashboard/src/features/orgs/projects/remote-schemas/utils/convertIntrospectionToSchema.ts
```diff
@@ -0,0 +1,10 @@
+import type { IntrospectRemoteSchemaResponse } from '@/utils/hasura-api/generated/schemas';
+import { buildClientSchema, type GraphQLSchema } from 'graphql';
+
+export default function convertIntrospectionToSchema(
+  introspectionData: IntrospectRemoteSchemaResponse,
+): GraphQLSchema | null {
+  return introspectionData?.data
+    ? buildClientSchema(introspectionData.data as any)
```
robertkasza: Please do not cast to **any**, it is almost never a good idea. Please check  why `directives` is missing from the generated type.

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaPreview.tsx
```diff
@@ -0,0 +1,178 @@
...
+            variant="secondary"
+            size="sm"
+            disabled={!searchTerm.trim() || isSearching}
+            loading={isSearching}
+          >
+            {isSearching ? 'Searching...' : 'Find'}
+          </Button>
+        </form>
+
+        {/* Search Message */}
```
robertkasza: Please remove them if they are not needed.

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaPreview.tsx
```diff
@@ -0,0 +1,178 @@
...
+          </Alert>
+        )} */}
+      </div>
+
+      <div className="p-4">
+        <RemoteSchemaTree
+          ref={treeRef}
+          className="min-h-[400px]"
+          schema={schema}
+          rootFields={['query', 'mutation', 'subscription']}
```
robertkasza: I tried a couple of remote schemas from here:
https://www.apollographql.com/blog/8-free-to-use-graphql-apis-for-your-projects-and-demos

It looks like we only display the tree if the query type is 'Query', 'mutation' or 'subscription' but there seem to be a lot more type which hasura supports and we should to

Could you check it please?

dashboard/src/features/orgs/projects/remote-schemas/components/BaseRemoteSchemaForm/BaseRemoteSchemaForm.tsx
```diff
@@ -0,0 +1,182 @@
...
+   * Optional slot to override the default GraphQL customizations section.
+   */
+  graphQLCustomizationsSlot?: React.ReactNode;
+}
+
+export const baseRemoteSchemaValidationSchema = Yup.object({
+  name: Yup.string().required('This field is required.'),
+  comment: Yup.string(),
+  definition: Yup.object({
+    url: Yup.string().required('This field is required.'),
```
robertkasza: There is no error message for the URL on the UI, if it's missing. Also it would be probably a good idea to validate if it is an actual URL or not (This is nice to have).

dashboard/src/features/orgs/projects/remote-schemas/components/BaseRemoteSchemaForm/BaseRemoteSchemaForm.tsx
```diff
@@ -0,0 +1,182 @@
...
+        'Either value or value from environment variable must be provided',
+        (obj) => {
+          const { value, value_from_env } = obj || {};
+          const hasValue = value?.trim() !== '';
+          const hasEnvValue = value_from_env?.trim() !== '';
+          return hasValue || hasEnvValue;
+        },
+      ),
+    ),
+    timeout_seconds: Yup.number().required('This field is required.'),
```
robertkasza: Also, no error message for this field. I think the whole definition related data need to be checked whether there is an error message or not.

dashboard/src/features/orgs/projects/remote-schemas/components/BaseRemoteSchemaForm/AdditionalHeadersEditor.tsx
```diff
@@ -0,0 +1,204 @@
+import { Box } from '@/components/ui/v2/Box';
```
robertkasza, line 1: Might be the wrong file, but this looks bad: 
image

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaBrowserSidebar/RemoteSchemaBrowserSidebar.tsx
```diff
@@ -0,0 +1,502 @@
...
+        {...props}
+      >
+        <RetryableErrorBoundary>
+          <RemoteSchemaBrowserSidebarContent
+            onSidebarItemClick={handleSidebarItemClick}
+          />
+        </RetryableErrorBoundary>
+      </Box>
+
+      <IconButton
```
robertkasza, line 488: The button is partially hidden
image

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaPreview.tsx
```diff
@@ -0,0 +1,168 @@
...
+      setIsSearching(true);
+
+      try {
+        const path = await treeRef.current.findItemPath(searchTerm);
+
+        if (path && path.length > 0) {
+          // Expand tree to show the found item
+          await treeRef.current.expandToItem(path);
+
+          // Focus and select the found item
```
robertkasza: I think this comment is not needed. Same for the above one.

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaPreview.tsx
```diff
@@ -0,0 +1,168 @@
...
+  const [isSearching, setIsSearching] = useState(false);
+
+  const schema = useMemo(() => {
+    if (introspectionData) {
+      return convertIntrospectionToSchema(introspectionData);
+    }
+    return null;
+  }, [introspectionData]);
+
+  const handleSearch = useCallback(
```
robertkasza: These useCallbacks do not make sense to me. The components that they are used with are not using React.memo. So why are we using it?

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaPreview.tsx
```diff
@@ -0,0 +1,168 @@
+import { Input } from '@/components/ui/v2/Input';
```
robertkasza, line 1: Could you tell me why we are using the v2 UI components?

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaPreview.tsx
```diff
@@ -0,0 +1,168 @@
...
+        <div className="mb-3 flex items-center justify-between">
+          <div>
+            <h3 className="text-sm font-medium">Schema Preview</h3>
+            <p className="mt-1 text-xs text-muted-foreground">
+              Browse the GraphQL schema structure
+            </p>
+          </div>
+        </div>
+
+        {/* Search Input */}
```
robertkasza: This comment is unnecessary as well.

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaPreview.tsx
```diff
@@ -0,0 +1,168 @@
...
+          </Button>
+        </form>
+      </div>
+
+      <div className="p-4">
+        <RemoteSchemaTree
+          ref={treeRef}
+          className="min-h-[400px]"
+          schema={schema}
+          rootFields={['query', 'mutation', 'subscription']}
```
robertkasza: Could you tell me  the reasoning behind to make `rootFields` a prop? The value seems to be always the same.

dashboard/src/features/orgs/projects/common/hooks/useGetMetadataResourceVersion/useGetMetadataResourceVersion.ts
```diff
@@ -0,0 +1,57 @@
...
+        process.env.NEXT_PUBLIC_ENV === 'dev'
+          ? getHasuraAdminSecret()
+          : project?.config?.hasura.adminSecret!;
+
+      return fetchExportMetadata({ appUrl, adminSecret });
+    },
+    {
+      ...queryOptions,
+      enabled: !!(
+        project?.subdomain &&
```
robertkasza: I think we do not need this check this hook is only used in drawers(or whatever it is called), so the project will be loaded, by the time this hook is mounted. I think the `project?.config?.hasura.adminSecret` is always set, if we are on the platform then project will have its own secret and when we are using the cli the local application's adminsecret is set here: https://github.com/nhost/nhost/blob/main/dashboard/src/features/orgs/utils/local-dashboard/index.ts#L40

dashboard/src/features/orgs/projects/remote-schemas/components/BaseRemoteSchemaRelationshipForm/sections/FieldToColumnMapSelector.tsx
```diff
@@ -0,0 +1,124 @@
...
+export default function FieldToColumnMapSelector({
+  sourceSchema,
+  disabled,
+}: FieldToColumnMapSelectorProps) {
+  const form = useFormContext<DatabaseRelationshipFormValues>();
+
+  const tableInfo = form.watch('table');
+  const { schema, name: table } = tableInfo;
+
+  // Watch the selected source type to get its fields
```
robertkasza: Please go through your PR and remove these unnecessary comments.

dashboard/src/features/orgs/projects/remote-schemas/components/BaseRemoteSchemaRelationshipForm/sections/FieldToColumnMapSelector.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const selectedSourceType = form.watch('sourceType');
+
+  // Introspect the source remote schema to get its types
```
robertkasza: unnecessary comment

dashboard/src/features/orgs/projects/remote-schemas/components/BaseRemoteSchemaRelationshipForm/sections/FieldToColumnMapSelector.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const { data: introspectionData } = useIntrospectRemoteSchemaQuery(
+    sourceSchema,
+    {
+      queryOptions: {
+        enabled: !!sourceSchema,
+      },
+    },
+  );
+
+  // Extract fields from the selected source type
```
robertkasza: unnecessary comment

dashboard/src/features/orgs/projects/remote-schemas/components/BaseRemoteSchemaRelationshipForm/sections/FieldToColumnMapSelectorItem.tsx
```diff
@@ -0,0 +1,197 @@
...
+  disabled,
+}: FieldToColumnMapSelectorItemProps) {
+  const form = useFormContext<DatabaseRelationshipFormValues>();
+
+  const [sourceFieldOpen, setSourceFieldOpen] = useState(false);
+  const [referenceColumnOpen, setReferenceColumnOpen] = useState(false);
+
+  const fieldMappings = form.watch('fieldMapping');
+
+  // Function to get available source fields for a specific row index
```
robertkasza: unnecessary comment

dashboard/src/features/orgs/projects/remote-schemas/components/BaseRemoteSchemaRelationshipForm/sections/RemoteSchemaRelationshipForm.tsx
```diff
@@ -0,0 +1,235 @@
...
+      targetRemoteSchema: defaultValues?.targetRemoteSchema || '',
+      sourceType: defaultValues?.sourceType || '',
+      targetField: defaultValues?.targetField || '',
+      mappings: defaultValues?.mappings || [],
+    },
+  });
+
+  const { isSubmitting } = form.formState;
+
+  // Fetch all remote schemas for the target schema dropdown
```
robertkasza: unnecessary comment

dashboard/src/features/orgs/projects/remote-schemas/components/BaseRemoteSchemaRelationshipForm/sections/RemoteSchemaRelationshipForm.tsx
```diff
@@ -0,0 +1,235 @@
...
+  const { data: remoteSchemas, status: remoteSchemasQueryStatus } =
+    useGetRemoteSchemas();
+
+  // Introspect the source remote schema to get its types
```
robertkasza: unnecessary comment

dashboard/src/features/orgs/projects/remote-schemas/components/BaseRemoteSchemaRelationshipForm/sections/RemoteSchemaRelationshipForm.tsx
```diff
@@ -0,0 +1,235 @@
...
+  const { data: sourceIntrospectionData } = useIntrospectRemoteSchemaQuery(
+    sourceSchema,
+    {
+      queryOptions: {
+        enabled: !!sourceSchema,
+      },
+    },
+  );
+
+  // Watch the selected target remote schema
```
robertkasza: unnecessary comment

dashboard/src/features/orgs/projects/remote-schemas/components/BaseRemoteSchemaRelationshipForm/sections/RemoteSchemaRelationshipForm.tsx
```diff
@@ -0,0 +1,235 @@
...
+  const targetRemoteSchemaValue = form.watch('targetRemoteSchema');
+
+  // Introspect the target remote schema to get its fields
```
robertkasza: unnecessary comment

dashboard/src/features/orgs/projects/remote-schemas/components/BaseRemoteSchemaRelationshipForm/sections/SchemaToArgumentMapSelector.tsx
```diff
@@ -0,0 +1,258 @@
...
+  disabled,
+}: SchemaToArgumentMapSelectorProps) {
+  const form = useFormContext<RemoteSchemaRelationshipFormValues>();
+
+  // Watch form values to reactively get field arguments
+  const selectedSourceType = form.watch('sourceType');
+  const selectedTargetRemoteSchema = form.watch('targetRemoteSchema');
+  const selectedTargetField = form.watch('targetField');
+
+  // Introspect the source remote schema to get its types and fields
```
robertkasza: unnecessary comment

dashboard/src/features/orgs/projects/remote-schemas/components/BaseRemoteSchemaRelationshipForm/sections/SchemaToArgumentMapSelector.tsx
```diff
@@ -0,0 +1,258 @@
...
+        enabled: !!sourceSchema,
+      },
+    },
+  );
+
+  // Introspect the target remote schema to get field arguments
+  const { data: targetIntrospectionData } = useIntrospectRemoteSchemaQuery(
+    selectedTargetRemoteSchema,
+    {
+      queryOptions: {
```
robertkasza: unnecessary comment

robertkasza (commented):
Please remove you should not need a comment to tell what the next line of code or block of code does.

dashboard/src/features/orgs/projects/common/hooks/useGetMetadataResourceVersion/useGetMetadataResourceVersion.ts
dbm03: Just to double-check, does this also mean that we can remove the check that chooses between `getHasuraAdminSecret()` and `project?.config?.hasura.adminSecret` in all queries? Including the hooks in the database page (although it would be part of a different PR).

Like:

https://github.com/nhost/nhost/pull/3299/files#diff-2ff47b78a00542df539fa5234f9b7ea5d2f0334731a8f39f31b57a85863e0950R36-R39

```ts
  const adminSecret =
        process.env.NEXT_PUBLIC_ENV === 'dev'
          ? getHasuraAdminSecret()
          : project?.config?.hasura.adminSecret!;
```

into

```ts
const adminSecret = project?.config?.hasura.adminSecret!;
```

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaPreview.tsx
robertkasza: discussed offline.

dashboard/src/features/orgs/projects/remote-schemas/components/EditRemoteSchemaPermissionsForm/RemoteSchemaRolePermissionsEditorForm.tsx
```diff
@@ -0,0 +1,911 @@
...
+                      Root Operations
+                    </Text>
+                    <div className="space-y-4 rounded border p-4">
+                      {rootTypes.map((schemaType) => (
+                        <div key={schemaType.name} className="space-y-2">
+                          <Text className="font-semibold text-blue-600">
+                            {schemaType.name.replace('type ', '')} Operations
+                          </Text>
+                          <div className="pl-4">
+                            <Accordion type="multiple" className="space-y-1">
```
robertkasza, line 581: image

There is a warning in this component because accordion and the checkbox both renders a button. Could you check if you can get rid of the warning or not? Since they are both "third party" component it's okay if we can't get rid of the warning but we should try at least.

dashboard/src/features/orgs/projects/remote-schemas/components/BaseRemoteSchemaRelationshipForm/sections/DatabaseRelationshipForm.tsx
```diff
@@ -0,0 +1,231 @@
...
+                  <Tooltip title="This will be used as the field name in the source type.">
+                    <InfoIcon
+                      aria-label="Info"
+                      className="h-4 w-4"
+                      color="primary"
+                    />
+                  </Tooltip>
+                </FormLabel>
+                <FormControl>
+                  <Input
```
robertkasza, line 140: image

The red border is missing.

dashboard/src/features/orgs/projects/remote-schemas/components/EditRemoteSchemaRelationships/sections/RelationshipTableCell.tsx
```diff
@@ -0,0 +1,115 @@
...
+        definition.to_source.field_mapping || {},
+      ) as string[];
+
+      return (
+        <span className="flex items-center gap-1 text-sm text-foreground">
+          <span className="truncate">{tableName}</span>
+          {columns.length > 0 ? <span className="px-1">/</span> : null}
+          {columns.map((col, index) => (
+            <span
+              key={`to-source-col-${tableName}-${col}`}
```
robertkasza: There is a possibility that we are going to have duplicate keys here. The user can selects the same column twice but for a different source field. 

For example selects the id columns for fieldA and fieldB. The UI will allow it, I do not know why anybody would do it but it can be done.

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaTree.test.tsx
```diff
@@ -0,0 +1,406 @@
...
+
+const schema = buildSchema(sdl);
+
+describe('RemoteSchemaTree', () => {
+  it('renders a tree for a basic schema and matches snapshot', () => {
+    render(<RemoteSchemaTree schema={schema} />);
+
+    const tree = screen.getByRole('tree', { name: 'Remote Schema Tree' });
+    expect(tree).toMatchInlineSnapshot(`
+      <div
```
robertkasza: I do not like snapshot tests in general, but in this case it does not make sense at all. The whole component is a third party package and we are taking a snapshot of that component checking if that renders its content well. We should not care about how `react-complex-tree` renders its content. If we want to check if certain things are rendered we can use `getBy...` or `findBy...` methods to check if the component renders the correct tree.

We might use snapshot testing  for checking if the correct styles applied based on the theme. (shallow rendering might work in this case, because we only care about the wrapper div).

In the `RemoteSchemaTree` component the `findItemPath` function seems complex, it might worth write a test for. In this case the test could be also used as "documentation", so that I(and other developers) do not need to check `react-complex-tree` documentation.

And we had problem with schemas that had different rootType, might worth writing a test for cases where we pass schemas with different rootTypes and the schema is rendered, but do not check it with snapshots.

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaTree.tsx
```diff
@@ -0,0 +1,192 @@
...
+        } else {
+          searchableText = String(item.data);
+        }
+
+        if (searchableText.toLowerCase().includes(searchTerm.toLowerCase())) {
+          return [...currentPath, itemId];
+        }
+
+        let result: string[] | null = null;
+        item.children?.some((childId) => {
```
robertkasza: This should be a foreach or a simple filter instead of some. What does this piece of code(i mean the some) supposed to do?

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaTree.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+    if (typeof item.data === 'string') {
+      return item.data;
+    }
+    return String(item.data);
+  };
+
+  return (
+    <div
+      className={`${className} ${theme.palette.mode === 'dark' ? 'rct-dark' : ''}`}
```
robertkasza: The `cn` utility function should be used to apply a class conditionally, like this:
`cn(className,  {'rct-dark':  theme.palette.mode === 'dark'} )`

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaTree.tsx
```diff
@@ -0,0 +1,192 @@
...
+  focusItem: (itemId: string) => void;
+  selectItems: (itemIds: string[]) => void;
+  focusTree: () => void;
+}
+
+export const RemoteSchemaTree = forwardRef<
+  RemoteSchemaTreeRef,
+  RemoteSchemaTreeProps
+>(({ schema, className }, ref) => {
+  const treeRef = useRef<any>(null);
```
robertkasza: Let's use the correct type for this ref.

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaTree.tsx
```diff
@@ -0,0 +1,192 @@
...
+  const environmentRef = useRef<any>(null);
```
robertkasza: Does this ref necessary? I think it is not used in this component anywhere. If it is not needed let's remove it.

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaTree.tsx
```diff
@@ -0,0 +1,192 @@
...
+      }),
+    [schema],
+  );
+
+  const [focusedItem, setFocusedItem] = useState<string>();
+  const [expandedItems, setExpandedItems] = useState<string[]>(['root']);
+  const [selectedItems, setSelectedItems] = useState<string[]>([]);
+
+  const findItemPath = useCallback(
+    async (
```
robertkasza: Could you explain it to me why this function needs to be async?

dashboard/src/features/orgs/projects/remote-schemas/components/BaseRemoteSchemaRelationshipForm/sections/DatabaseRelationshipForm.tsx
dbm03: Should be fixed altogether with the other comboboxes in the last commit https://github.com/nhost/nhost/pull/3299/commits/12bab5e8f2220473dbaef457f8d0f14cfd9b65ab, thx

dashboard/src/features/orgs/projects/remote-schemas/components/EditRemoteSchemaPermissionsForm/RemoteSchemaRolePermissionsEditorForm.tsx
dbm03: I managed to fix it in the commit https://github.com/nhost/nhost/commit/12bab5e8f2220473dbaef457f8d0f14cfd9b65ab, by replacing the checkbox with its primitive and adding the `asChild` prop and a span element below. Thx

dashboard/src/features/orgs/projects/remote-schemas/components/EditRemoteSchemaRelationships/sections/RelationshipTableCell.tsx
dbm03: I've added `sourceField` to the `id` prop in  https://github.com/nhost/nhost/pull/3299/commits/aa9607a9c5e0643a0c1643da05db2c0c11e038bc

Thank you

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaTree.tsx
dbm03: The body is synchronous, so it doesn't have to be async (probably async was there because of messing with debouncing search before). I'm updating it in the last commit thank you

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaTree.tsx
dbm03: Removed in last commit https://github.com/nhost/nhost/pull/3299/commits/358bc55ecde8fa44dee28485dccfa0aa76420cae. 👍

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaTree.tsx
dbm03: Fixed in https://github.com/nhost/nhost/commit/358bc55ecde8fa44dee28485dccfa0aa76420cae

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaTree.tsx
dbm03: Thx, I'm replacing this code block here https://github.com/nhost/nhost/pull/3299/commits/5b4fb5852427a42168daf60688f3ab321ccdbd08 by:

```ts
let foundPath: string[] | null = null;
// eslint-disable-next-line no-restricted-syntax
for (const childId of item.children ?? []) {
  foundPath = searchInItem(childId, [...currentPath, itemId]);
  if (foundPath) {
    break;
  }
}
return foundPath;
```

The idea is to find the first occurrence in `searchInItem` of a node that matches our `searchTerm` in lowercase, so that it can target any type of node in the tree: field names, data types...

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaTree.tsx
dbm03: Fixed in https://github.com/nhost/nhost/pull/3299/commits/5b4fb5852427a42168daf60688f3ab321ccdbd08, thx

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaTree.test.tsx
dbm03: I've added the following tests in the commit https://github.com/nhost/nhost/pull/3299/commits/41925714baf4c158940b372efab0a3c35e20daea:

- Check that the `.rct-dark` class is getting applied when you switch to dark theme.
- No snapshot tests, replaced some of them with `querySelector` method.
- Tests for some scenarios in `findItemPath`, which could serve as documentation for this function.

robertkasza: Not sure about the querySelectors, generally these kind of test should be written from a user POV. For example in the UI the user will see texts, buttons, inputs etc. So we, should find these elements based on their roles and texts.(findByText, findByRole) Or when it is really hard to find the elements then we can use **data-testid** attribute for that. 

But we can use querySelector, do not need to replace them.

dashboard/src/features/orgs/projects/common/hooks/useGetMetadataResourceVersion/useGetMetadataResourceVersion.ts
robertkasza: Probably yes, but we will do that some other time.

PR #3324, opened 2025-05-05
ajustes

carlosrobertodevops:


dbarrosop:
Hello, thanks for the PR. Unfortunately I am not sure I understand the motivation of the change. I would suggest to start by reverting any change that isn't strictly necessary and then explain why the change is needed in the self-hosted example.

dbarrosop:
Closing as we had to ban this user due to being disrespectful to members of the community.

PR #3330, opened 2025-05-08
fix (examples/nextjs-server-components): proper use of onError in middleware

dimaMachina:
onError should be returned since it can contain `NextResponse.redirect` like in the example  https://github.com/nhost/nhost/blob/e87505c564f81b2e3464fd4c7ab8623fe76d6579/examples/quickstarts/nextjs-server-components/src/middleware.ts#L5-L7


dbarrosop (approved):

dbarrosop:
Thanks for the contribution!

Issue #3308, opened 2025-04-20
update to nextjs 15


srghma:
1. `getServerSideProps` is deprecated - https://docs.nhost.io/reference/nextjs/get-nhost-session
2. version should be updated - https://github.com/nhost/nhost/blob/1bc2fabe599c9ecb643dc25b32f32715df7913c8/examples/nextjs/package.json#L27
3. `_app` and `_document` is now `_layout` - https://github.com/nhost/nhost/blob/main/examples/nextjs/pages/_app.tsx

srghma:
https://github.com/nhost/nhost/blob/main/packages/nextjs/src/create-server-side-client.ts should be rewritten to use https://nextjs.org/docs/app/api-reference/functions/cookies istead of js-cookie package

srghma:
also possible to remove 'cookies' from ClientStorageType and use 'custom' in

```
export const localStorageGetter = (
  clientStorageType: ClientStorageType,
  clientStorage?: ClientStorage
): StorageGetter => {
```

but this is optional (why? to make js-cookies dep from hasura-auth-js package)

dbarrosop:
Hi, thanks a lot for bringing this to our attention and for the analysis. I don't think we will bother adding nextjs15 support to the current SDK though. We are currently working a new SDK that should be easier to maintain and extend:

https://github.com/nhost/sdk-experiment

There is currently a nextjs15 example there:

https://github.com/nhost/sdk-experiment/tree/main/demos/nextjs-ssr-demo

I would suggest you to help out there reviewing it and providing feedback and testing the new SDK. The SDK is not yet ready, we are still working on it and making changes based on feedback.

PR #3337, opened 2025-05-18
feat(nextjs): #3308 update to nextjs 15 -> use next/headers cookies in create-server-side-client

srghma:


dbarrosop:
Unfortunately this is a breaking change, in addition, for the reasons explained in #3308 it is probably not worth the effort.

PR #3346, opened 2025-05-24
fix(examples/docker-compose): fix postgres persistent data storage path

PrinzKenny1:
With the current path it won't persist the data across restarts, so here's a fix for it :P


dbarrosop:
Thanks!

PR #3369, opened 2025-06-17
fix (dashboard): request new mfa ticket after error when signing in

robertkasza:


dbarrosop (approved):

dashboard/src/features/auth/SignIn/SignInWithEmailAndPassword/components/SignInWithEmailAndPassword.tsx
```diff
@@ -1,15 +1,31 @@
 import useOnSignUpWithPasswordHandler from '@/features/auth/SignIn/SignInWithEmailAndPassword/hooks/useOnSignInWithEmailAndPasswordHandler';
```
dbm03: I believe we should change the name of the `useOnSignUpWithPasswordHandler` -> `useOnSignInWithEmailAndPasswordHandler`, to the one we're actually importing. I confused it with the other `useOnSignUpWithPasswordHandler` hook for a moment

PR #3390, opened 2025-07-14
chore (dashboard): turn on null checks

robertkasza:


dashboard/src/utils/callAll/callAll.ts
```diff
@@ -1,7 +1,9 @@
 /**
  * Calls all the functions in the array with the given arguments.
  */
-export default function callAll(...fns: ((...args: any[]) => any)[]) {
+export default function callAll(
```
dbm03: I think it would be ideal to replace this with the following code snippet:

https://gist.github.com/lifeiscontent/6696dce07d2da474fc6b9fcc48288bbe

```ts
const callAll = <Args extends readonly unknown[]>(
  ...fns: readonly (((...args: Args) => void) | undefined)[]
) => (...args: Args): void => fns.forEach((fn) => fn?.(...args));

export default callAll;
```

I believe using `any` removes the type inference in the editor

dashboard/src/pages/orgs/[orgSlug]/projects/[appSubdomain]/hasura.tsx
```diff
@@ -32,7 +32,7 @@ export default function HasuraPage() {
   const hasuraUrl =
     process.env.NEXT_PUBLIC_ENV === 'dev' || !isPlatform
       ? `${getHasuraConsoleServiceUrl()}`
-      : generateAppServiceUrl(project?.subdomain, project?.region, 'hasura', {
+      : generateAppServiceUrl(project!.subdomain, project!.region, 'hasura', {
```
Comment (dbm03) [inline dashboard/src/pages/orgs/[orgSlug]/projects/[appSubdomain]/hasura.tsx:35]: Are we really sure that project is not `null` here? What if there's an `error` object in the `useProject` hook? It's probably fine most of the time, as we're checking the loading state.

I'm just thinking about a case where the server could return an error on the getProject query, after `loading` is set to `false`.

dashboard/src/pages/github-app-installation/index.tsx
```diff
@@ -75,6 +77,7 @@ function Page() {
   }
 
   if (state.error) {
+    // eslint-disable-next-line @typescript-eslint/no-throw-literal
```
dbm03, line 80: I believe disabling eslint here is no longer necessary. I am not getting any warning in my IDE after removing the comment

dashboard/src/features/orgs/projects/services/components/ServiceForm/ServiceForm.tsx
```diff
@@ -144,7 +145,7 @@ export default function ServiceForm({
   const getFormattedConfig = (values: ServiceFormValues) => {
     // Remove any __typename property from the values
     const sanitizedValues = removeTypename(values) as ServiceFormValues;
```
dbm03, line 147: It could be nice to have better typing in the `removeTypename` function. I believe it will reduce some of the complexity from the type assertions that we're using.

Maybe something like the following works, inside `helpers.ts`:

https://github.com/nhost/nhost/blob/a4fbc5869b2d56c6b970dea8a4fe3d0b1e0f4771/dashboard/src/utils/helpers/helpers.ts#L94

```ts
/**
 * Utility type that recursively removes '__typename' properties from types
 */
type RemoveTypename<T> = T extends readonly (infer U)[]
  ? readonly RemoveTypename<U>[]
  : T extends (infer U)[]
    ? RemoveTypename<U>[]
    : T extends object
      ? {
          [K in keyof T as K extends '__typename' ? never : K]: RemoveTypename<
            T[K]
          >;
        }
      : T;

/**
 * Recursively removes the property '__typename' from a JavaScript object and its nested objects and arrays.
 * Preserves the original type structure while removing '__typename' properties.
 */
export const removeTypename = <T>(obj: T): RemoveTypename<T> => {
  if (!obj || typeof obj !== 'object') {
    return obj as RemoveTypename<T>;
  }

  if (Array.isArray(obj)) {
    return obj.map((item) => removeTypename(item)) as RemoveTypename<T>;
  }

  const newObj: Record<string, unknown> = {};

  Object.keys(obj).forEach((key) => {
    if (key !== '__typename') {
      newObj[key] = removeTypename((obj as Record<string, unknown>)[key]);
    }
  });

  return newObj as RemoveTypename<T>;
};
```

dashboard/src/pages/orgs/[orgSlug]/projects/[appSubdomain]/hasura.tsx
Comment (robertkasza) [reply dashboard/src/pages/orgs/[orgSlug]/projects/[appSubdomain]/hasura.tsx:35]: Project can't be null here. Becasue of this: https://github.com/nhost/nhost/blob/main/dashboard/src/pages/orgs/%5BorgSlug%5D/projects/%5BappSubdomain%5D/index.tsx#L7


What has worked before probably will work after this too. By turning on this config it will not suddenly break the dashboard. Typescript just will warn you if you trying to do something that yous shouldn't.

But good thinking.

dashboard/src/pages/github-app-installation/index.tsx
robertkasza: My IDE, is showing an error. I will leave it here, it will be removed anyway, once we migrate to biome.js

dashboard/src/features/orgs/projects/services/components/ServiceForm/ServiceForm.tsx
robertkasza: Let's try and not add to more to this PR and focus on the strict null check. By implementing this I add at least 10 other Typescript error.

That said it is probably a good idea to have type other than any for the result. I created an issue to later address this

https://github.com/nhost/nhost/issues/3413

dashboard/src/features/orgs/hooks/useFinishOrgCreation/useFinishOrgCreation.ts
```diff
@@ -36,37 +36,37 @@ function useFinishOrgCreation({
...
           async () => {
-            const {
-              data: { billingPostOrganizationRequest },
-            } = await postOrganizationRequest({
+            const response = await postOrganizationRequest({
               variables: {
                 sessionID: session_id as string,
               },
             });
+            if (response.data) {
```
dbm03: I believe we should check `if (response.data?.billingPostOrganizationRequest) {` here instead.

(Same as in https://github.com/nhost/nhost/blob/f19943c2e18ab91dc400abde70589cc76336fb66/dashboard/src/features/orgs/components/members/components/NotificationsTray/NotificationsTray.tsx#L93)

dashboard/src/features/orgs/projects/ai/DeleteAutoEmbeddingsModal/DeleteAutoEmbeddingsModal.tsx
```diff
@@ -29,24 +30,39 @@ export default function DeleteAutoEmbeddingsModal({
...
+        project.region,
+        'graphql',
+      );
+      return new ApolloClient({
+        cache: new InMemoryCache(),
+        link: new HttpLink({
+          uri: serviceUrl,
+          headers: {
+            'x-hasura-admin-secret':
+              process.env.NEXT_PUBLIC_ENV === 'dev'
```
dbm03, line 50: I think this `NEXT_PUBLIC_ENV === 'dev'` check here is redundant, as it should be already covered by the `if` statement above

dashboard/src/features/orgs/projects/services/components/ServiceForm/ServiceForm.tsx
dbm03: Thank you for creating the issue, I'm resolving this one

dbm03 (approved):
LGTM! thank you. I left only a couple more comments in the review

dashboard/src/features/orgs/projects/ai/DeleteAutoEmbeddingsModal/DeleteAutoEmbeddingsModal.tsx
robertkasza: The `NEXT_PUBLIC_ENV !== 'dev'` check should be removed from the if statement above, because this way we only use the admin secret from the config, no matter which environment we're using.

PR #3398, opened 2025-07-22
feat: dashboard: new onboarding

nunopato:


dashboard/src/lib/segment.ts
```diff
@@ -9,7 +7,7 @@ export const analytics = AnalyticsBrowser.load(
     writeKey: process.env.NEXT_PUBLIC_ANALYTICS_WRITE_KEY!,
   },
   {
-    disable: !isPlatform() || isDevOrStaging()
+    // disable: !isPlatform() || isDevOrStaging()
```
robertkasza: Do not forget to uncomment this!

dashboard/src/pages/index.tsx
```diff
@@ -22,17 +22,23 @@ export default function IndexPage() {
       }
 
       if (orgs) {
-        const orgFromLastSlug = orgs.find((o) => o.slug === lastSlug);
+        // check if user has no organizations (first-time user)
+        // should we render this even if user is not coming in for the first time?
```
robertkasza: This line should be removed.

dashboard/src/features/orgs/components/CreateOrgFormDialog/CreateOrgFormDialog.tsx
```diff
@@ -8,6 +8,18 @@ import {
   DialogTrigger,
 } from '@/components/ui/v3/dialog';
 import { Input } from '@/components/ui/v3/input';
+import {
```
robertkasza, line 11: The starter plan is not disabled if the user has a free org already, since the user can only have one free organization, I think we should disable it.

nunopato: done 👍

dashboard/src/lib/segment.ts
nunopato: done 👍

dashboard/src/pages/index.tsx
nunopato: done 👍

PR #3424, opened 2025-08-14
feat (dashboard): Allow composite keys

robertkasza:


dbarrosop (approved):

dashboard/src/features/orgs/projects/database/dataGrid/components/EditTableForm/EditTableForm.tsx
```diff
@@ -144,10 +152,17 @@ export default function EditTableForm({
   ]);
 
   async function handleSubmit(values: BaseTableFormValues) {
+    const primaryKey = values.primaryKeyIndices.reduce<string[]>(
```
dbm03: I believe this can be simplified with a `.map()` function instead of a `reduce`. Something like:

```ts
const primaryKey = values.primaryKeyIndices.map<string>(
  (index) => values.columns[index].name,
);
```

dashboard/src/features/orgs/projects/database/dataGrid/hooks/useUpdateTableMutation/prepareUpdateTableQuery.ts
```diff
@@ -108,28 +109,43 @@ export default function prepareUpdateTableQuery({
...
-  const currentPrimaryKey = originalColumns.find((column) => column.isPrimary);
+  const currentPrimaryKeys = originalColumns.filter(
+    (column) => column.isPrimary,
+  );
+
+  const currentPrimaryKeyNames = currentPrimaryKeys
+    .map((pk) => `${pk.name}`)
+    .join('');
+  const updatedPrimaryKeys = updatedTable.primaryKey.join('');
+  if (currentPrimaryKeyNames !== updatedPrimaryKeys) {
```
dbm03: I think there's an edge case that isn't covered by this check. There can be collisions between different composite primary keys.

For instance:

If the current Primary Key is `['user', 'id', 'type']`
`currentPrimaryKeyNames := 'useridtype'`

And you try to update the Primary Key to `['userid','type']`
`updatedPrimaryKeys := 'useridtype'`

Submitting the form won't save the new primary key

PR #3434, opened 2025-08-19
fix (dashboard): GitHub verify email redirect

dbm03:


dashboard/src/pages/github/verify.tsx
```diff
@@ -0,0 +1,99 @@
+import { NavLink } from '@/components/common/NavLink';
```
robertkasza: Why do we want to create a new page for verifying emails? Can not we reuse the one we have already?

Just remove the `useEffect` from the email/verify page (we do not need it anyway) and based on whether there is an email in the query or not you show the button or the form.

something like this:
https://gist.github.com/robertkasza/5ba3af520de6cc8fd53045c6164c3308

dashboard/src/features/auth/SignIn/SignInWithEmailAndPassword/hooks/useResendVerificationEmail.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export default function useResendVerificationEmail() {
+  const nhost = useNhostClient();
+  const [loading, setLoading] = useState(false);
+
+  const resendVerificationEmail = async (email: string) => {
+    setLoading(true);
+
+    try {
+      await nhost.auth.sendVerificationEmail({ email: email as string });
```
robertkasza: The casting is unnecessary.

dashboard/src/providers/Auth/AuthProvider.tsx
```diff
@@ -66,6 +68,22 @@ function AuthProvider({ children }: PropsWithChildren) {
...
+        if (error === 'unverified-user') {
+          removeRefreshTokenFromQuery();
+          await push('/github/verify');
+        } else {
+          const description =
+            typeof errorDescription === 'string'
+              ? errorDescription
+              : 'An error occurred during the sign-in process. Please try again.';
+          toast.error(description, getToastStyleProps());
+          removeRefreshTokenFromQuery();
```
robertkasza: Let's rename this function because it does more than just removes the refreshToken from the query maybe `removeQueryParamsFromURL` or something like that.

dashboard/src/pages/github/verify.tsx
dbm03: I've just added these changes you suggested in my last commit. Thank you!

dbarrosop (approved):

PR #3536, opened 2025-09-30
fix (dashboard): fix flaky e2e tests

robertkasza:
### Checklist

- [ ] No breaking changes
- [ ] Tests pass


dashboard/e2e/database/create-table.test.ts
```diff
@@ -245,75 +245,78 @@ test('should not be able to create a table with a name that already exists', asy
...
     columns: [
       { name: 'id', type: 'uuid', defaultValue: 'gen_random_uuid()' },
+      { name: 'second_id', type: 'uuid', defaultValue: 'gen_random_uuid()' },
       { name: 'name', type: 'text' },
     ],
   });
 
-  // create table
+  expect(page.locator('div[data-testid="id"]')).toBeVisible();
+  expect(page.locator('div[data-testid="second_id"]')).toBeVisible();
```
dbm03: Why don't we do `await expect` here instead? Wouldn't this mean that if for some reason the next step of navigating to the `/public/${tableName}` URL is faster, then the `div` elements with testid  `id` and `second_id` would no longer be visible and then throw an error?

robertkasza: I forgot about it, or rather I just did not notice that I should await the expect. I added it but since these are after the navigation the scenario that you mentioned is not possible because we are awaiting the navigation and only after the navigation we check for the badges.

dbm03: Thanks, I resolve the comment

PR #3558, opened 2025-10-03
fix(dashboard): improve remote schema preview search

dbm03:


dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaPreview.tsx
```diff
@@ -57,19 +72,70 @@ export default function RemoteSchemaPreview({
...
+      setHasSearched(false);
     }
   };
 
+  useEffect(() => {
+    const onKeyDown = (e: KeyboardEvent) => {
+      const target = e.target as HTMLElement | null;
+      const inTree =
+        treeContainerRef.current?.contains(target as Node) ?? false;
+      const isSearchInput = target === searchInputRef.current;
```
robertkasza: Based on the usage of this variable this should be: `const isNotSearchInput = target !== searchInputRef.current;` and use without negating the value.

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaPreview.tsx
```diff
@@ -41,14 +58,12 @@ export default function RemoteSchemaPreview({
...
-        treeRef.current.focusItem(foundItemId);
+      const allPaths = treeRef.current.findAllItemPaths(searchTerm);
+      setMatches(allPaths);
+      setHasSearched(true);
+      if (allPaths.length > 0) {
+        setMatchIndex(0);
+        await navigateToMatch(allPaths, 0);
       }
     } catch (searchError) {
       console.error('Search error:', searchError);
```
robertkasza: What kind of errors can be thrown here? Is it okay not to show the error on the UI?

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaTree.tsx
```diff
@@ -72,32 +74,30 @@ export const RemoteSchemaTree = forwardRef<
...
-      ): string[] | null => {
+        acc: string[][],
+      ): void => {
         const item = treeData[itemId];
         if (!item) {
-          return null;
+          return;
         }
 
         const searchableText = getText(item.data);
```
robertkasza, line 73: The serchableText looks someting like this: `undefinedcreated: String`. Searching for **undefined** will return everything as a match:
image

Also, I think we should move the getText function out of the useCallback/component since it is a utility function and it is not using component related data.

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaTree.tsx
```diff
@@ -72,32 +74,30 @@ export const RemoteSchemaTree = forwardRef<
...
         if (searchableText.toLowerCase().includes(lowerSearchTerm)) {
-          return [...currentPath, itemId];
+          acc.push([...currentPath, itemId]);
```
robertkasza, line 75: This is quite complex logic and as far as I can tell the return value off **findAllItemPaths** is stored in the matches state variable in `RemoteSchemaPreview` component. And it looks like only the last item is used from that array in the `RemoteSchemaPreview` in the `navigateToMatch` function
```
const navigateToMatch = async (paths: string[][], index: number) => {
    if (!treeRef.current || paths.length === 0) {
      return;
    }

    const path = paths[index];
    await treeRef.current.expandToItem(path);
    const foundItemId = path[path.length - 1]; // only last item is needed.
    treeRef.current.selectItems([foundItemId]);
    treeRef.current.focusItem(foundItemId);
  };
...
// this is how it's used
const nextIndex = (matchIndex + step + matches.length) % matches.length;
setMatchIndex(nextIndex);
navigateToMatch(matches, nextIndex);
```

Do we need all the other **paths** from the `collectInItem` function? I do not seem to find where they are used other than just they are spread into the result of the collectInItem funtion.

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaTree.tsx
```diff
@@ -142,6 +142,37 @@ export const RemoteSchemaTree = forwardRef<
     return String(item.data);
   };
 
+  const highlightNode = (
```
robertkasza: This is a utility function should not be defined inside a component. Please find it a place(outside of the component) and move it there.

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaPreview.tsx
```diff
@@ -57,19 +72,70 @@ export default function RemoteSchemaPreview({
...
+      setSearchTerm('');
+      setMatches([]);
+      setMatchIndex(0);
+      setHasSearched(false);
     }
   };
 
+  useEffect(() => {
+    const onKeyDown = (e: KeyboardEvent) => {
+      const target = e.target as HTMLElement | null;
```
robertkasza: This cast in not necessary, `target` is only used once and we cast its type to node.

```
const { target } = e;
const inTree =  treeContainerRef.current?.contains(target as Node) ?? false;
```

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaPreview.tsx
dbm03: There could be errors when we're expanding the preview tree and it changes its structure for some reason (e.g. the introspection data network response changes). I think it's better to show on the UI, so I've replaced the `console.error` with an error toast in the interface. Thank you

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaPreview.tsx
dbm03: Fixed in the last commit, ty

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaTree.tsx
dbm03: I've added the following check to the `getText` function, and moved it inside `utils.ts`

```ts
if (children === undefined || children === null) {
      return '';
    }
```

Thanks

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaTree.tsx
dbm03: Only the last id is used for selection/focus, but `expandToItem` (called right before) needs the full path:
I've added the functionality to hit `'Enter'` and cycle through all the occurrences of your search. Therefore, I think we need to store the paths so we can expand all the items of each occurrence in the tree with `treeRef.current.expandToItem(path)`.

Is there another way of expanding the elements when you cycle through them, without having to store all the paths?
Maybe we could store only leaf ids, but we'd need a fast way to derive ancestor paths from an id, and I think that would a bit be more complicated.

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaPreview.tsx
dbm03: I fixed this in the last commit, thank you

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaTree.tsx
dbm03: I moved this to `utils.ts` in the last commit, thx

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaTree.tsx
robertkasza: It's okay. I am just blind I have not seen this: `await treeRef.current.expandToItem(path);`

dashboard/src/features/orgs/projects/remote-schemas/components/RemoteSchemaPreview/RemoteSchemaTree.tsx
robertkasza: Please use `isEmptyValue` instead. And Let's write some tests for this function

robertkasza (approved):

PR #3567, opened 2025-10-07
feat(dashboard): event triggers: redeliver events

dbm03:


dashboard/src/features/orgs/projects/events/event-triggers/components/EventTriggerInvocationLogsDataTable/InvocationLogActionsCell.tsx
```diff
@@ -21,8 +26,40 @@ export default function InvocationLogActionsCell({
...
 
+  const prevIdsRef = useRef<string[]>([]);
+  const [skeletonId, setSkeletonId] = useState<string | null>(null);
+  const loadingToastIdRef = useRef<string | null>(null);
+  const intervalRef = useRef<NodeJS.Timeout | null>(null);
+
+  const tableRows = table.getCoreRowModel().rows;
+
+  useEffect(() => {
+    const currentIds = tableRows.map((i) => i.original.id);
```
robertkasza: nit: tableRow or row instead of i

dashboard/src/features/orgs/projects/events/event-triggers/components/EventTriggerInvocationLogsDataTable/InvocationLogActionsCell.tsx
```diff
@@ -43,9 +123,13 @@ export default function InvocationLogActionsCell({
             size="sm"
             onClick={handleRedeliver}
             className="-ml-1 h-8 w-8 p-0"
-            disabled
+            disabled={isRedelivering || !!skeletonId}
           >
-            <CalendarSync className="h-4 w-4" />
+            {isRedelivering || skeletonId ? (
+              <Spinner size="small" />
+            ) : (
```
robertkasza: There is a button with loading indicator. 

https://github.com/nhost/nhost/blob/main/dashboard/src/components/ui/v3/button.tsx#L57

dashboard/src/features/orgs/projects/events/event-triggers/components/EventTriggerInvocationLogsDataTable/types.ts
```diff
@@ -4,4 +4,8 @@ import type { Dispatch, SetStateAction } from 'react';
 export interface EventTriggerInvocationLogsDataTableMeta {
   selectedLog: EventInvocationLogEntry | null;
   setSelectedLog: Dispatch<SetStateAction<EventInvocationLogEntry | null>>;
+  addPendingSkeleton?: () => string;
```
robertkasza: I believe these props can not be undefined.

dashboard/src/features/orgs/projects/events/event-triggers/hooks/useEventTriggerPagination/useEventTriggerPagination.ts
```diff
@@ -3,6 +3,8 @@ import { useCallback, useEffect, useMemo, useState } from 'react';
 type QueryResult<TData> = {
   data: TData | undefined;
   isLoading: boolean;
+  isInitialLoading: boolean;
+  refetch?: () => Promise<unknown> | void;
```
robertkasza: Refetch can be undefined?

dashboard/src/features/orgs/projects/events/event-triggers/hooks/useEventTriggerPagination/useEventTriggerPagination.ts
```diff
@@ -39,6 +41,8 @@ export interface UseEventTriggerPaginationResult<TData = unknown[]> {
   hasNoNextPage: boolean;
   data: TData | undefined;
   isLoading: boolean;
+  isInitialLoading: boolean;
+  refetch?: () => Promise<unknown> | void;
```
robertkasza: Same here?

dashboard/src/features/orgs/projects/events/event-triggers/components/EventTriggerInvocationLogsDataTable/InvocationLogActionsCell.tsx
```diff
@@ -32,7 +69,50 @@ export default function InvocationLogActionsCell({
     }
   };
 
-  const handleRedeliver = () => {};
+  const handleRedeliver = async () => {
```
robertkasza, line 66: How should the redeliver supposed to work? Can multiple events redelivered at the same time? In hasura it seems that you can redeliver only one event(because of the modal).

I am asking this because I am not fond of the skeleton + retry logic implementation and I want to understand what we are tying to achieve to see if there is a better way do this or not.

dbm03: After you redeliver an event through Hasura's API, you have to wait for a bit to get the invocation log associated to the redelivered event. It takes some time to send the request to the webhook URL.

In Hasura's console, this wait time is inside the modal. When the modal is opened, it's making SQL queries in the background, until it checks that there's a new item. It assumes that this new item is the invocation log associated with the event you just redelivered, and it shows you its request/response.

I though that the limitation of redeliver only one event is not necessary. You don't need to disrupt the user with a modal. Thus, you're able to redeliver more than one event in the dashboard. 

In our case for the dashboard, when you click on redeliver, a skeleton element is added, and we refetch the invocation logs of the redelivered event. Unlike Hasura, it doesn't use SQL queries, because the invocation logs refetching has the same necessary info. 

Overall, the algorithm that we follow is: Click on redeliver event -> Interval with setInterval is created, that refetches invocations every 5 seconds. In a useEffect, we detect whether the new invocation IDs that we just fetched include a new ID that wasn't there previously. If so, we can assume that it's the new invocation from our redelivered event; we remove the skeleton and replace it by the new invocation log information

dashboard/src/features/orgs/projects/events/event-triggers/components/EventTriggerInvocationLogsDataTable/InvocationLogActionsCell.tsx
dbm03: I've replaced it by `<ButtonWithLoading>`. Thank you

I'm leaving the check inside to not show both the loading spinner and the CalendarSync icon at the same time.

```tsx
<ButtonWithLoading
variant="ghost"
size="sm"
onClick={handleRedeliver}
className="-ml-1 h-8 w-8 p-0"
disabled={isRedeliverDisabled}
loading={isRedeliverDisabled}
>
    {!isRedeliverDisabled && <CalendarSync className="h-4 w-4" />}
</ButtonWithLoading>
```

dashboard/src/features/orgs/projects/events/event-triggers/components/EventTriggerInvocationLogsDataTable/EventTriggerInvocationLogsDataTable.tsx
```diff
@@ -137,13 +164,18 @@ export default function EventTriggerInvocationLogsDataTable({
...
-          {!isLoading && table.getRowModel().rows?.length === 0 && (
-            <TableRow>
-              <TableCell colSpan={columns.length} className="h-24 text-center">
-                No results.
-              </TableCell>
-            </TableRow>
-          )}
+          {!isLoading &&
+            table.getRowModel().rows?.length === 0 &&
+            !isRedeliverPending && (
```
robertkasza: Do we really need to check if there is a  pending redelivering? If there are no rows for the table we can not start an event redeliver, can we?

dashboard/src/features/orgs/projects/events/event-triggers/components/EventTriggerInvocationLogsDataTable/InvocationLogActionsCell.tsx
```diff
@@ -21,8 +27,33 @@ export default function InvocationLogActionsCell({
...
+  const loadingToastIdRef = useRef<string | null>(null);
+  const intervalRef = useRef<NodeJS.Timeout | null>(null);
+  const [isThisRedeliverActionLoading, setIsThisRedeliverActionLoading] =
+    useState(false);
+
+  const tableRows = table.getCoreRowModel().rows;
+
+  const isRedeliverDisabled = isRedelivering || meta.isRedeliverPending;
+
+  const cleanState = () => {
```
robertkasza: nit: clear instead of clean. I think we usually clear state instead of cleaning it. or rather we resetting it.

dbm03: I'm updating it to `resetState`, as we also use that same term in the codebase. Thx

dashboard/src/features/orgs/projects/events/event-triggers/components/EventTriggerInvocationLogsDataTable/EventTriggerInvocationLogsDataTable.tsx
dbm03: I removed the "is pending redelivering" check. It doesn't make sense to check it since there aren't any redeliver buttons when the table is empty

PR #3578, opened 2025-10-09
fix(dashboard): Run audit and lint in dashboard

robertkasza:
### Checklist

- [ ] No breaking changes
- [ ] Tests pass


dashboard/package.json
```diff
@@ -7,9 +7,11 @@
     "dev": "RECOIL_DUPLICATE_ATOM_KEY_CHECKING_ENABLED=false next dev",
     "build": "next build --no-lint",
     "analyze": "ANALYZE=true pnpm build --no-lint",
+    "audit:dashboard": "pnpm exec audit-ci --config ../audit-ci.jsonc",
     "start": "next start",
     "lint": "next lint --max-warnings 0",
-    "test": "vitest --run",
+    "test": "pnpm audit:dashboard && pnpm lint && pnpm test:vitest",
```
dbarrosop: the audit part is not really needed, `make check` runs the following:

https://github.com/nhost/nhost/blob/main/dashboard/project.nix#L93-L100

which is basically this:

https://github.com/nhost/nhost/blob/main/nixops/lib/js/js.nix#L81-L131

dashboard/src/features/orgs/projects/deployments/components/DeploymentServiceLogs/DeploymentServiceLogsHeader.test.tsx
```diff
@@ -20,6 +20,17 @@ const mockServices = [
   'job-backup',
 ];
 
+Object.defineProperty(HTMLElement.prototype, 'getBoundingClientRect', {
```
dbm03, line 23: I'm not sure if this is correct, but I believe that changing this global prototype property `getBoundingClientRect` could interfere with other test files as well, since it's the same worker that runs all the tests.

It could be better to use `spyOn` to mock the implementation of the method temporarily, and restore it afterwards, like in:

```ts
let rectSpy: ReturnType<typeof vi.spyOn>;

beforeEach(() => {
  rectSpy = vi
    .spyOn(HTMLElement.prototype, 'getBoundingClientRect')
    .mockReturnValue({ width: 100, height: 40, top: 0, left: 0, bottom: 40, right: 100 } as any);
});

afterEach(() => {
  rectSpy.mockRestore();
});
```

dashboard/src/tests/msw/mocks/graphql/permissionVariablesQuery.ts
```diff
@@ -1,11 +1,11 @@
+import { HttpResponse } from 'msw';
 import nhostGraphQLLink from './nhostGraphQLLink';
 
 const permissionVariablesQuery = nhostGraphQLLink.query(
   'GetRolesPermissions',
-  (_req, res, ctx) =>
-    res(
-      ctx.delay(250),
```
dbm03, line 7: Why was the `ctx.delay(250)` here in the first place? (I think removing the delay is correct of course. I'm just curious if you know it was added because of some specific test for loading states?)

dashboard/src/features/orgs/projects/deployments/components/DeploymentServiceLogs/DeploymentServiceLogsHeader.test.tsx
robertkasza: I do not think that they are interfere with each other. If they were it should be enough to add this only once, but we have this override in three test files to suppress the warnings. If you remove one of them the warnings reappear for that test file.

Also, I have not look looked into this, but if one test file could effect another that would be a fundamental flaw in how tests are executed/ran. Despite they are run by the same worker test files are isolated from each other.

dashboard/src/tests/msw/mocks/graphql/permissionVariablesQuery.ts
robertkasza: I think it was added to mimic how an API call would work in the real world. And it might have been added to test transition states, but not sure.

If you would like to test transition states it is better to "pause" the promise and only resolve it once you checked the transition state. It might not be such a bad way to use delays in vitest/jest tests to test this, but in wdio/playwright test it is not a good practice to use delays because it makes the test flaky.

https://github.com/nhost/nhost/blob/main/dashboard/src/tests/testUtils.tsx#L152-L176

This is a good way to test transition states. Here is an example:
https://github.com/nhost/nhost/blob/main/dashboard/src/features/orgs/layout/OrgLayout/OrganizationGuard.test.tsx

PR #3720, opened 2025-11-24
feat(dashboard): edit table relationships

dbm03:
### **PR Type**
Enhancement

---

### **Description**

This PR implements a comprehensive database relationship management system for the Nhost dashboard, enabling users to create, edit, and manage table relationships through an intuitive UI.

**Core Features:**
- **Create Relationships**: Support for object, array, and remote relationships via bulk operations
- **Relationship Suggestions**: AI-powered relationship discovery and creation workflows  
- **Visual Management**: Table-based interface showing relationship details with action buttons
- **Multi-Source Support**: Handle relationships between different databases and remote schemas
- **Form Validation**: Type-safe forms with zod schemas and real-time validation

**Technical Implementation:**
- Generated TypeScript schemas from OpenAPI specifications for type safety
- React Hook Form integration with custom FormCombobox component for better UX
- React Query for efficient data fetching and caching of relationship metadata
- Proper error handling and loading states throughout the relationship workflows
- Comprehensive utility functions for data transformation and validation

**API Operations Added:**
- `pg_create_object_relationship` - Create one-to-one relationships
- `pg_create_array_relationship` - Create one-to-many relationships  
- `pg_drop_relationship` - Remove existing relationships
- `suggest_relationships` - Get AI-powered relationship suggestions

---

### **Key Components**

- **EditRelationshipsForm**: Main interface for managing table relationships
- **BaseRelationshipDialog**: Reusable dialog for creating/editing relationships
- **FormCombobox**: Enhanced combobox component with search and validation
- **Relationship Hooks**: Custom hooks for CRUD operations on relationships
- **Type System**: Comprehensive TypeScript types generated from OpenAPI specs

---

### **Architecture Diagram**

```mermaid
flowchart LR
  A["OpenAPI Spec"] --> B["Generated TypeScript Schemas"]
  B --> C["Relationship Operations"]
  C --> D["Create Object/Array"]
  C --> E["Drop Relationship"] 
  C --> F["Suggest Relationships"]
  
  G["Dashboard UI"] --> H["EditRelationshipsForm"]
  H --> I["BaseRelationshipDialog"]
  I --> J["Form Components"]
  J --> K["API Mutations"]
  K --> C
```

---

### **Testing**

- **Unit Tests**: Core utilities and data transformation functions
- **Integration Tests**: API mutation hooks and form validation
- **Type Tests**: Comprehensive TypeScript coverage with generated schemas

---

### **Migration & Compatibility**

- **Non-Breaking**: All changes are additive and backward compatible
- **Database**: No schema changes required - uses existing Hasura metadata API
- **Dependencies**: Minimal new dependencies, leveraging existing tech stack

### DIFF
--- a/dashboard/src/utils/hasura-api/openapi.yaml
+++ b/dashboard/src/utils/hasura-api/openapi.yaml
@@ -19,39 +19,39 @@ paths:
           application/json:
             schema:
               oneOf:
-                - $ref: '#/components/schemas/ExportMetadata'
-                - $ref: '#/components/schemas/SetTableCustomizationBulkOperation'
-                - $ref: '#/components/schemas/SetTableIsEnumBulkOperation'
-                - $ref: '#/components/schemas/SuggestRelationshipsOperation'
-                - $ref: '#/components/schemas/DropRelationshipBulkOperation'
-                - $ref: '#/components/schemas/DeleteRemoteRelationshipBulkOperation'
-                - $ref: '#/components/schemas/CreateObjectRelationshipBulkOperation'
-                - $ref: '#/components/schemas/CreateArrayRelationshipBulkOperation'
-                - $ref: '#/components/schemas/CreateRemoteRelationshipBulkOperation'
-                - $ref: '#/components/schemas/RenameRelationshipBulkOperation'
-                - $ref: '#/components/schemas/AddRemoteSchemaOperation'
-                - $ref: '#/components/schemas/UpdateRemoteSchemaBulkOperation'
-                - $ref: '#/components/schemas/RemoveRemoteSchemaOperation'
-                - $ref: '#/components/schemas/ReloadRemoteSchemaOperation'
-                - $ref: '#/components/schemas/IntrospectRemoteSchemaOperation'
-                - $ref: '#/components/schemas/RemoteSchemaRemoteRelationshipOperation'
-                - $ref: '#/components/schemas/DeleteRemoteSchemaRemoteRelationshipOperation'
-                - $ref: '#/components/schemas/RemoteSchemaPermissionsBulkOperation'
-                - $ref: '#/components/schemas/GetEventLogsOperation'
-                - $ref: '#/components/schemas/GetEventAndInvocationLogsByIdOperation'
-                - $ref: '#/components/schemas/RedeliverEventOperation'
-                - $ref: '#/components/schemas/CreateEventTriggerBulkOperation'
-                - $ref: '#/components/schemas/DeleteEventTriggerBulkOperation'
-                - $ref: '#/components/schemas/TestWebhookTransformOperation'
-                - $ref: '#/components/schemas/InvokeEventTriggerOperation'
-                - $ref: '#/components/schemas/GetCronTriggersOperation'
-                - $ref: '#/components/schemas/CreateCronTriggerBulkOperation'
-                - $ref: '#/components/schemas/DeleteCronTriggerOperation'
-                - $ref: '#/components/schemas/GetScheduledEventsOperation'
-                - $ref: '#/components/schemas/GetScheduledEventInvocationsOperation'
-                - $ref: '#/components/schemas/DeleteScheduledEventOperation'
+                - $ref: "#/components/schemas/ExportMetadata"
+                - $ref: "#/components/schemas/SetTableCustomizationBulkOperation"
+                - $ref: "#/components/schemas/SetTableIsEnumBulkOperation"
+                - $ref: "#/components/schemas/SuggestRelationshipsOperation"
+                - $ref: "#/components/schemas/DropRelationshipBulkOperation"
+                - $ref: "#/components/schemas/DeleteRemoteRelationshipBulkOperation"
+                - $ref: "#/components/schemas/CreateObjectRelationshipBulkOperation"
+                - $ref: "#/components/schemas/CreateArrayRelationshipBulkOperation"
+                - $ref: "#/components/schemas/CreateRemoteRelationshipBulkOperation"
+                - $ref: "#/components/schemas/RenameRelationshipBulkOperation"
+                - $ref: "#/components/schemas/AddRemoteSchemaOperation"
+                - $ref: "#/components/schemas/UpdateRemoteSchemaBulkOperation"
+                - $ref: "#/components/schemas/RemoveRemoteSchemaOperation"
+                - $ref: "#/components/schemas/ReloadRemoteSchemaOperation"
+                - $ref: "#/components/schemas/IntrospectRemoteSchemaOperation"
+                - $ref: "#/components/schemas/RemoteSchemaRemoteRelationshipOperation"
+                - $ref: "#/components/schemas/DeleteRemoteSchemaRemoteRelationshipOperation"
+                - $ref: "#/components/schemas/RemoteSchemaPermissionsBulkOperation"
+                - $ref: "#/components/schemas/GetEventLogsOperation"
+                - $ref: "#/components/schemas/GetEventAndInvocationLogsByIdOperation"
+                - $ref: "#/components/schemas/RedeliverEventOperation"
+                - $ref: "#/components/schemas/CreateEventTriggerBulkOperation"
+                - $ref: "#/components/schemas/DeleteEventTriggerBulkOperation"
+                - $ref: "#/components/schemas/TestWebhookTransformOperation"
+                - $ref: "#/components/schemas/InvokeEventTriggerOperation"
+                - $ref: "#/components/schemas/GetCronTriggersOperation"
+                - $ref: "#/components/schemas/CreateCronTriggerBulkOperation"
+                - $ref: "#/components/schemas/DeleteCronTriggerOperation"
+                - $ref: "#/components/schemas/GetScheduledEventsOperation"
+                - $ref: "#/components/schemas/GetScheduledEventInvocationsOperation"
+                - $ref: "#/components/schemas/DeleteScheduledEventOperation"
       responses:
-        '200':
+        "200":
           description: Operation executed successfully
           content:
             application/json:
@@ -62,16 +62,16 @@ paths:
                       message:
                         type: string
                         example: success
-                  - $ref: '#/components/schemas/ExportMetadataResponse'
-                  - $ref: '#/components/schemas/SuggestRelationshipsResponse'
-                  - $ref: '#/components/schemas/IntrospectRemoteSchemaResponse'
-                  - $ref: '#/components/schemas/GetEventLogsResponse'
-                  - $ref: '#/components/schemas/GetEventAndInvocationLogsByIdResponse'
-                  - $ref: '#/components/schemas/TestWebhookTransformResponse'
-                  - $ref: '#/components/schemas/InvokeEventTriggerResponse'
-                  - $ref: '#/components/schemas/GetCronTriggersResponse'
-                  - $ref: '#/components/schemas/GetScheduledEventsResponse'
-                  - $ref: '#/components/schemas/GetScheduledEventInvocationsResponse'
+                  - $ref: "#/components/schemas/ExportMetadataResponse"
+                  - $ref: "#/components/schemas/SuggestRelationshipsResponse"
+                  - $ref: "#/components/schemas/IntrospectRemoteSchemaResponse"
+                  - $ref: "#/components/schemas/GetEventLogsResponse"
+                  - $ref: "#/components/schemas/GetEventAndInvocationLogsByIdResponse"
+                  - $ref: "#/components/schemas/TestWebhookTransformResponse"
+                  - $ref: "#/components/schemas/InvokeEventTriggerResponse"
+                  - $ref: "#/components/schemas/GetCronTriggersResponse"
+                  - $ref: "#/components/schemas/GetScheduledEventsResponse"
+                  - $ref: "#/components/schemas/GetScheduledEventInvocationsResponse"
                   - type: array
                     description: List of success messages for bulk operations
                     items:
@@ -81,13 +81,13 @@ paths:
                           type: string
                           example: success
 
-        '400':
+        "400":
           description: Bad request
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/ErrorResponse'
-        '401':
+                $ref: "#/components/schemas/ErrorResponse"
+        "401":
           description: Unauthorized
           content:
             application/json:
@@ -96,12 +96,12 @@ paths:
                 properties:
                   error:
                     type: string
-        '500':
+        "500":
           description: Internal server error
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/ErrorResponse'
+                $ref: "#/components/schemas/ErrorResponse"
   /apis/migrate:
     post:
       summary: Execute a migration
@@ -112,20 +112,20 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/MigrationRequest'
+              $ref: "#/components/schemas/MigrationRequest"
       responses:
-        '200':
+        "200":
           description: Migration executed successfully
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/SuccessResponse'
-        '500':
+                $ref: "#/components/schemas/SuccessResponse"
+        "500":
           description: Internal server error
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/ErrorResponse'
+                $ref: "#/components/schemas/ErrorResponse"
               example:
                 code: internal_error
                 message: "{\n  \"error\": \"query execution failed\",\n  \"path\": \"$[0]\",\n  \"code\": \"postgres-error\",\n  \"internal\": {\n    \"arguments\": [],\n    \"error\": {\n      \"description\": null,\n      \"exec_status\": \"FatalError\",\n      \"hint\": null,\n      \"message\": \"syntax error at or near \\\":\\\"\",\n      \"status_code\": \"42601\"\n    },\n    \"prepared\": false,\n    \"statement\": \"CREATE TABLE asd ( id TEXT ):;\"\n  }\n}"
@@ -161,7 +161,7 @@ components:
         args:
           type: array
           items:
-            $ref: '#/components/schemas/SetTableCustomizationStep'
+            $ref: "#/components/schemas/SetTableCustomizationStep"
 
     SetTableCustomizationStep:
       type: object
@@ -173,7 +173,7 @@ components:
           type: string
           enum: [pg_set_table_customization]
         args:
-          $ref: '#/components/schemas/SetTableCustomizationArgs'
+          $ref: "#/components/schemas/SetTableCustomizationArgs"
 
     SetTableCustomizationArgs:
       type: object
@@ -185,10 +185,10 @@ components:
           description: Name of the source database of the table
           default: default
         table:
-          $ref: '#/components/schemas/QualifiedTable'
+          $ref: "#/components/schemas/QualifiedTable"
         configuration:
           description: Configuration for the table/view
-          $ref: '#/components/schemas/TableConfig'
+          $ref: "#/components/schemas/TableConfig"
 
     TableConfig:
       type: object
@@ -198,12 +198,12 @@ components:
           description: Customize the <table-name> with the provided custom name value. The GraphQL nodes for the table will be generated according to the custom name.
           nullable: true
         custom_root_fields:
-          $ref: '#/components/schemas/CustomRootFields'
+          $ref: "#/components/schemas/CustomRootFields"
         column_config:
-          description: 'A { [column_name]: ColumnConfigValue } object.'
+          description: "A { [column_name]: ColumnConfigValue } object."
           type: object
           additionalProperties:
-            $ref: '#/components/schemas/ColumnConfigValue'
+            $ref: "#/components/schemas/ColumnConfigValue"
         custom_column_names:
           type: object
           description: Customize the column fields (deprecated in favour of custom_name on column_config)
@@ -233,57 +233,57 @@ components:
           description: Customize the <table-name> root field. Using a String customizes the field name.
           oneOf:
             - type: string
-            - $ref: '#/components/schemas/CustomRootField'
+            - $ref: "#/components/schemas/CustomRootField"
         select_by_pk:
           description: Customize the <table-name>_by_pk root field. Using a String customizes the field name.
           oneOf:
             - type: string
-            - $ref: '#/components/schemas/CustomRootField'
+            - $ref: "#/components/schemas/CustomRootField"
         select_aggregate:
           description: Customize the <table-name>_aggregate root field. Using a String customizes the field name.
           oneOf:
             - type: string
-            - $ref: '#/components/schemas/CustomRootField'
+            - $ref: "#/components/schemas/CustomRootField"
         select_stream:
           description: Customize the <table-name>_stream root field. Using a String customizes the field name.
           oneOf:
             - type: string
-            - $ref: '#/components/schemas/CustomRootField'
+            - $ref: "#/components/schemas/CustomRootField"
         insert:
           description: Customize the insert_<table-name> root field. Using a String customizes the field name.
           oneOf:
             - type: string
-            - $ref: '#/components/schemas/CustomRootField'
+            - $ref: "#/components/schemas/CustomRootField"
         insert_one:
           description: Customize the insert_<table-name>_one root field. Using a String customizes the field name.
           oneOf:
             - type: string
-            - $ref: '#/components/schemas/CustomRootField'
+            - $ref: "#/components/schemas/CustomRootField"
         update:
           description: Customize the update_<table-name> root field. Using a String customizes the field name.
           oneOf:
             - type: string
-            - $ref: '#/components/schemas/CustomRootField'
+            - $ref: "#/components/schemas/CustomRootField"
         update_many:
           description: Customize the update_many_<table-name> root field. Using a String customizes the field name.
           oneOf:
             - type: string
-            - $ref: '#/components/schemas/CustomRootField'
+            - $ref: "#/components/schemas/CustomRootField"
         update_by_pk:
           description: Customize the update_<table-name>_by_pk root field. Using a String customizes the field name.
           oneOf:
             - type: string
-            - $ref: '#/components/schemas/CustomRootField'
+            - $ref: "#/components/schemas/CustomRootField"
         delete:
           description: Customize the delete_<table-name> root field. Using a String customizes the field name.
           oneOf:
             - type: string
-            - $ref: '#/components/schemas/CustomRootField'
+            - $ref: "#/components/schemas/CustomRootField"
         delete_by_pk:
           description: Customize the delete_<table-name>_by_pk root field. Using a String customizes the field name.
           oneOf:
             - type: string
-            - $ref: '#/components/schemas/CustomRootField'
+            - $ref: "#/components/schemas/CustomRootField"
 
     CustomRootField:
       type: object
@@ -319,7 +319,7 @@ components:
         args:
           type: array
           items:
-            $ref: '#/components/schemas/SetTableIsEnumStep'
+            $ref: "#/components/schemas/SetTableIsEnumStep"
 
     SetTableIsEnumStep:
       type: object
@@ -331,7 +331,7 @@ components:
           type: string
           enum: [pg_set_table_is_enum]
         args:
-          $ref: '#/components/schemas/SetTableIsEnumArgs'
+          $ref: "#/components/schemas/SetTableIsEnumArgs"
 
     SetTableIsEnumArgs:
       type: object
@@ -340,7 +340,7 @@ components:
         - is_enum
       properties:
         table:
-          $ref: '#/components/schemas/QualifiedTable'
+          $ref: "#/components/schemas/QualifiedTable"
         is_enum:
           type: boolean
           description: Whether or not the table should be used as an enum table <enum table>.
@@ -360,7 +360,7 @@ components:
           type: string
           enum: [pg_suggest_relationships]
         args:
-          $ref: '#/components/schemas/SuggestRelationshipsArgs'
+          $ref: "#/components/schemas/SuggestRelationshipsArgs"
 
     SuggestRelationshipsArgs:
       type: object
@@ -375,7 +375,7 @@ components:
           items:
             oneOf:
               - type: string
-              - $ref: '#/components/schemas/QualifiedTable'
+              - $ref: "#/components/schemas/QualifiedTable"
             default: All Tables
         omit_tracked:
           type: boolean
@@ -389,14 +389,14 @@ components:
           type: array
           items:
             oneOf:
-              - $ref: '#/components/schemas/SuggestedObjectRelationship'
-              - $ref: '#/components/schemas/SuggestedArrayRelationship'
+              - $ref: "#/components/schemas/SuggestedObjectRelationship"
+              - $ref: "#/components/schemas/SuggestedArrayRelationship"
 
     SuggestedRelationshipElement:
       type: object
       properties:
         table:
-          $ref: '#/components/schemas/QualifiedTable'
+          $ref: "#/components/schemas/QualifiedTable"
         columns:
           type: array
           items:
@@ -412,13 +412,13 @@ components:
           enum: [object]
         from:
           allOf:
-            - $ref: '#/components/schemas/SuggestedRelationshipElement'
+            - $ref: "#/components/schemas/SuggestedRelationshipElement"
             - type: object
               properties:
                 constraint_name:
                   type: string
         to:
-          $ref: '#/components/schemas/SuggestedRelationshipElement'
+          $ref: "#/components/schemas/SuggestedRelationshipElement"
 
     SuggestedArrayRelationship:
       type: object
@@ -427,10 +427,10 @@ components:
           type: string
           enum: [array]
         from:
-          $ref: '#/components/schemas/SuggestedRelationshipElement'
+          $ref: "#/components/schemas/SuggestedRelationshipElement"
         to:
           allOf:
-            - $ref: '#/components/schemas/SuggestedRelationshipElement'
+            - $ref: "#/components/schemas/SuggestedRelationshipElement"
             - type: object
               properties:
                 constraint_name:
@@ -454,7 +454,7 @@ components:
         args:
           type: array
           items:
-            $ref: '#/components/schemas/DropRelationshipStep'
+            $ref: "#/components/schemas/DropRelationshipStep"
 
     DropRelationshipStep:
       type: object
@@ -472,7 +472,7 @@ components:
             - relationship
           properties:
             table:
-              $ref: '#/components/schemas/QualifiedTable'
+              $ref: "#/components/schemas/QualifiedTable"
             relationship:
               type: string
               description: Name of the relationship that needs to be dropped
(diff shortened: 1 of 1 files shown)


dashboard/src/components/form/FormCombobox/FormCombobox.tsx
dbm03, line 1: I'm adding this component because I think it'd be useful in other forms in the future as well. What do you think?

dashboard/src/utils/hasura-api/generated/default/default.ts
dbm03, line 1: Most changes in generated/ files are because of the orval version update signature

dashboard/src/features/orgs/projects/database/dataGrid/components/RenameRelationshipDialog/RenameRelationshipDialog.tsx
```diff
@@ -0,0 +1,249 @@
...
+          hideCloseButton
+          disableOutsideClick={isSubmitting}
+        >
+          <DialogHeader>
+            <DialogTitle className="text-foreground">
+              Rename Relationship
+            </DialogTitle>
+            <DialogDescription>
+              Provide a new name for the{' '}
+              <TextWithTooltip
```
robertkasza, line 177: Image

This looks strange. But first, could you tell me why a tooltip is needed to display the relationship's name?

dashboard/e2e/utils.ts
```diff
@@ -396,3 +396,150 @@ export async function cleanupRemoteSchemaTestIfNeeded() {
...
+  page: Page;
+  relationshipName: string;
+  type: 'object' | 'array';
+  referenceTable: string;
+  sourceColumn: string;
+  referenceColumn: string;
+  referenceSource?: string;
+  referenceSchema?: string;
+}) {
+  // Fill relationship name
```
robertkasza: Could you remove the comments where it's just describes what the next few line do, I mean it is just unnecessary in my opinion. In the other test file too.

dashboard/src/components/form/FormCombobox/FormCombobox.tsx
robertkasza: It's a good idea, but I do not like the implementation.
Why are we passing the options as children? I think we should pass them as an array of objects. Or is there something that I am not seeing?

The other thing that I do not understand is why we need to find the selected label. Won't just findSelecetedLabel  return the value? To me it seems like that fieldValue and selectedLabel will be the same. Same thing I might missing something here.

Could you explain it to me?

dashboard/src/components/form/FormCombobox/FormCombobox.tsx
```diff
@@ -0,0 +1,338 @@
...
+          ref: fieldRef,
+          onChange,
+          value,
+        } = isNotEmptyValue(transform)
+          ? getTransformedFieldProps(field, transform)
+          : field;
+
+        let fieldValue = '';
+        if (isNotEmptyValue(value)) {
+          fieldValue = typeof value === 'string' ? value : String(value);
```
robertkasza: Can this be anything other than string?

dashboard/src/features/orgs/projects/database/dataGrid/components/AddSuggestedRelationshipDialog/AddSuggestedRelationshipDialog.tsx
```diff
@@ -0,0 +1,305 @@
...
+   * Table where the new relationship will be created.
+   */
+  tableName: string;
+  /**
+   * Suggested relationship to be transformed into a tracked relationship.
+   */
+  suggestion: SuggestedObjectRelationship | SuggestedArrayRelationship;
+}
+
+function getDefaultRelationshipName(
```
robertkasza: I think we should use the same naming convention that we are using when creating/adding foreign keys to a table.

Singular for object relationships and plural for array relationships. These relationship names are used in a graphql queries. Let's look at this example: 

We have an author table and a books table. When I want to get the author's books I would write(or would make sense to me) the following:

```
query {
  author {
    name
    books {
      title
    }
  }
}


query {
  book {
    title
    author {
      name
    }
  }
}
```

This makes much more sense to me than book_array or author_object.

dashboard/src/features/orgs/projects/database/dataGrid/components/AddSuggestedRelationshipDialog/AddSuggestedRelationshipDialog.tsx
```diff
@@ -0,0 +1,305 @@
...
+      const createRelationship =
+        suggestion.type === 'array'
+          ? createArrayRelationship
+          : createObjectRelationship;
+
+      promise = createRelationship({
+        resourceVersion,
+        args,
+      });
+    } catch (error) {
```
robertkasza, line 170: Move create relationship to execPromiseWithErrorToast

dashboard/src/features/orgs/projects/database/dataGrid/components/AddSuggestedRelationshipDialog/AddSuggestedRelationshipDialog.tsx
```diff
@@ -0,0 +1,305 @@
...
+  tableName,
+  suggestion,
+}: AddSuggestedRelationshipDialogProps) {
+  const [open, setOpen] = useState(false);
+  const { data: resourceVersion } = useGetMetadataResourceVersion();
+
+  const {
+    mutateAsync: createObjectRelationship,
+    isLoading: isCreatingObjectRelationship,
+  } = useCreateObjectRelationshipMutation();
```
robertkasza: useCreateRelationshipMutation with types from hasura (pg_create_object_relationship,  pg_create_array_relationship) pass the type to the createRelationship mutation

dashboard/src/features/orgs/projects/database/dataGrid/components/BaseRelationshipDialog/sections/RemoteSchemaRelationshipDetails.tsx
```diff
@@ -0,0 +1,301 @@
...
+
+  const { watch, formState } = form;
+  const { isSubmitting: disabled } = formState;
+
+  const remoteSchemaFormValue = watch('remoteSchema');
+  const remoteSchema = remoteSchemaFormValue.name;
+  const initialRemoteField = remoteSchemaFormValue?.remoteField;
+  const selectedFromSource = watch('fromSource');
+
+  const { data: fromTableData } = useTableQuery(
```
robertkasza, line 37: This query also fetches the rows, do we need them here? If not there is an option to prevent row fetching(`preventRowFetching`)

dashboard/src/features/orgs/projects/database/dataGrid/components/BaseRelationshipDialog/sections/RemoteSchemaRelationshipDetails.tsx
```diff
@@ -0,0 +1,301 @@
...
+export interface RemoteSchemaRelationshipDetailsProps {
+  onChange: (value: RemoteSchemaRelationshipDetailsValue) => void;
+}
+
+export default function RemoteSchemaRelationshipDetails({
+  onChange,
+}: RemoteSchemaRelationshipDetailsProps) {
+  const onChangeRef = useRef(onChange);
+  useEffect(() => {
+    onChangeRef.current = onChange;
```
robertkasza: Move the onchange logic inside this component

dashboard/src/features/orgs/projects/database/dataGrid/components/BaseRelationshipDialog/sections/TableRelationshipDetails.tsx
```diff
@@ -0,0 +1,195 @@
...
+  const {
+    fields: fieldMappingFields,
+    append: appendFieldMapping,
+    remove: removeFieldMapping,
+  } = useFieldArray({
+    control,
+    name: 'fieldMapping',
+  });
+
+  const { data: fromTableData } = useTableQuery(
```
robertkasza, line 28: Same here, this query fetches the rows as well, I do not think we need them.

dashboard/src/features/orgs/projects/database/dataGrid/components/BaseRelationshipDialog/sections/TableRelationshipDetails.tsx
```diff
@@ -0,0 +1,195 @@
...
+              </FormSelect>
+
+              <div className="col-span-2 flex justify-center">
+                <ArrowRight className="h-4 w-4 text-muted-foreground" />
+              </div>
+
+              <FormSelect
+                control={form.control}
+                name={`fieldMapping.${index}.referenceColumn`}
+                label=""
```
robertkasza: I think we can change the the label prop to be optional

dashboard/src/features/orgs/projects/database/dataGrid/components/BaseRelationshipDialog/BaseRelationshipDialog.tsx
```diff
@@ -0,0 +1,199 @@
...
+  submitButtonText?: string;
+  initialValues?: BaseRelationshipFormValues;
+  onSubmit: (values: BaseRelationshipFormValues) => Promise<void>;
+  isEditing?: boolean;
+}
+
+export default function BaseRelationshipDialog({
+  open,
+  setOpen,
+  dialogTitle = 'Create Relationship',
```
robertkasza: please do not set a default value like this for a base component. 
My problem with this is that the name of the component is telling me that it is a "generic/base" component, but the default values suggesting me that it is a component for creating relationships.

dashboard/src/features/orgs/projects/database/dataGrid/components/BaseRelationshipDialog/BaseRelationshipFormTypes.tsx
```diff
@@ -0,0 +1,162 @@
+import { z } from 'zod';
```
robertkasza, line 1: is this really a types file? I am not sure that this counts as types. What do you think?

dashboard/src/features/orgs/projects/database/dataGrid/components/BaseRelationshipDialog/SourceAndReferenceSelector.tsx
```diff
@@ -0,0 +1,389 @@
...
+  }, [remoteSchemas, remoteSchemasStatus]);
+
+  return (
+    <div className="flex flex-row gap-4">
+      <div className="flex flex-1 flex-col gap-4 rounded-md border p-4">
+        <h3 className="font-semibold text-foreground text-sm">From Source</h3>
+
+        <div className="space-y-4">
+          <div className="grid gap-4 sm:grid-cols-2">
+            <FormSelect
```
robertkasza: The source is always disabled I  think that we do not need to use Forminput and select for readonly values. What do you think?

dashboard/src/features/orgs/projects/database/dataGrid/components/BaseRelationshipDialog/SourceAndReferenceSelector.tsx
```diff
@@ -0,0 +1,389 @@
...
+      );
+    }
+
+    if (!remoteSchemas?.length) {
+      return null;
+    }
+    return (
+      <>
+        <SelectSeparator />
+        <SelectItem disabled value="__remote-schemas-label">
```
robertkasza: Why aren't we using a SelectGroup instead?

dashboard/src/components/form/FormCombobox/FormCombobox.tsx
```diff
@@ -0,0 +1,338 @@
...
+  label: ReactNode;
+  placeholder?: string;
+  className?: string;
+  containerClassName?: string;
+  inline?: boolean;
+  helperText?: string | null;
+  disabled?: boolean;
+  searchPlaceholder?: string;
+  emptyText?: string;
+  showSelectedIcon?: boolean;
```
robertkasza: we do not need this

dashboard/src/components/form/FormCombobox/FormCombobox.tsx
```diff
@@ -0,0 +1,338 @@
...
+  transform?: Transformer;
+  'data-testid'?: string;
```
robertkasza, line 54: opitons prop for displaying the command items({ label:string, value: string }[])

dashboard/src/features/orgs/projects/database/dataGrid/components/BaseRelationshipDialog/sections/RemoteSchemaRelationshipDetails.tsx
```diff
@@ -0,0 +1,301 @@
...
+                </div>
+              </div>
+            );
+          })()}
+        </div>
+
+        {selectedRootFieldPath ? (
+          <div className="space-y-2 rounded-md border border-border p-3">
+            <div className="font-medium text-sm">Selected field tree</div>
+            {(() => {
```
robertkasza: create component instead of an iife

dashboard/src/features/orgs/projects/database/dataGrid/components/BaseRelationshipDialog/sections/TableRelationshipDetails.tsx
```diff
@@ -0,0 +1,195 @@
...
+        control={form.control}
+        name="relationshipType"
+        label="Relationship Type"
+        placeholder="Select relationship type"
+      >
+        <SelectItem value="object">Object Relationship</SelectItem>
+        <SelectItem value="array">Array Relationship</SelectItem>
+      </FormSelect>
+
+      <div className="space-y-2 rounded-md border p-4">
```
robertkasza, line 98: I can create a relationship without mappings, shouldn't mappings be required?

dashboard/src/features/orgs/projects/database/dataGrid/components/DataBrowserSidebar/DataBrowserSidebar.tsx
```diff
@@ -450,6 +451,22 @@ function DataBrowserSidebarContent({
                               false,
                             );
                           }}
+                          onEditRelationships={() => {
+                            openDrawer({
+                              title: 'Edit Relationships',
+                              component: (
+                                <EditRelationshipsForm
```
robertkasza: It looks like I cannot add a relationship to untracked tables. I think we should disable or hide the option if the table is not tracked.

dashboard/src/features/orgs/projects/database/dataGrid/components/DeleteRelationshipDialog/DeleteRelationshipDialog.tsx
```diff
@@ -0,0 +1,159 @@
...
+  } = useDeleteRemoteRelationshipMutation();
+
+  const isDeletingRelationship = isRemoteRelationship
+    ? isDeletingRemoteRelationship
+    : isDeletingLocalRelationship;
+
+  const { data: resourceVersion } = useGetMetadataResourceVersion();
+
+  const handleDeleteDialogClick = async () => {
+    if (!resourceVersion) {
```
robertkasza: Shouldn't we disable or/and show a loading indicator while we do not have the resourceVersion. Or maybe fetch it before the actual mutation, so that we have the latest. I am not sure where but I did get a couple of resource version mismatch error. But I am not sure, it is just a thought.

dashboard/src/features/orgs/projects/database/dataGrid/components/DeleteRelationshipDialog/DeleteRelationshipDialog.tsx
```diff
@@ -0,0 +1,159 @@
...
+      triggerToast('Metadata is not ready yet. Please try again in a moment.');
+      return;
+    }
+
+    let deleteRelationshipPromise: Promise<MetadataOperation200>;
+    if (isRemoteRelationship) {
+      deleteRelationshipPromise = deleteRemoteRelationship({
```
robertkasza: I think it would be better to just select the mutation based on a condition, because it looks like both mutation accepts the same arguments

dashboard/src/features/orgs/projects/database/dataGrid/components/BaseRelationshipDialog/BaseRelationshipFormTypes.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+  static createTypeRemoteSchemaFromName(name: string): ReferenceSource {
+    return new ReferenceSource(
+      `${ToReferenceSourceTypePrefix.REMOTE_SCHEMA}${name}`,
+    );
+  }
+}
+
+const baseRelationshipFormSchema = z.object({
+  name: z.string().min(1, { message: 'Name is required' }),
```
robertkasza: There is no other validation for the name in the BaseRelationshipDialog component, but there is a message saying you cannot use special characters when renaming a relationship. Can we add validation for only using letters, numbers, and underscores, because i get an error if I submit the form with a special character in its name.

dashboard/src/features/orgs/projects/database/dataGrid/components/RenameRelationshipDialog/RenameRelationshipDialog.tsx
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    if (trimmedName === relationshipToRename.trim()) {
+      showRelationshipNameError(
+        'New relationship name must be different from the current name.',
+      );
+      return;
+    }
+
+    const promise = renameRelationship({
```
robertkasza: we can move this inside the execPromiseWihtErrorToast call.

dashboard/src/features/orgs/projects/database/dataGrid/EditRelationshipsForm/EditRelationshipsForm.tsx
```diff
@@ -0,0 +1,231 @@
...
+    relationships: existingRelationshipsViewModel,
+    isLoading: isRelationshipsLoading,
+  } = useGetRelationships({
+    dataSource,
+    schema,
+    tableName: originalTable.table_name,
+  });
+  const tableQueryKey = [`${dataSource}.${schema}.${originalTable.table_name}`];
+
+  const { status: tableStatus, error: tableError } = useTableQuery(
```
robertkasza, line 63: preventRowFetching

dashboard/src/features/orgs/projects/database/dataGrid/hooks/useGetSuggestedRelationships/useGetSuggestedRelationships.ts
```diff
@@ -0,0 +1,70 @@
...
+
+interface UseGetSuggestedRelationshipsOptions {
+  dataSource: string;
+  schema: string;
+  tableName: string;
+}
+
+const filterNotNullRelationshipSuggestionViewModel = (
+  item: RelationshipSuggestionViewModel | null,
+): item is RelationshipSuggestionViewModel => item !== null;
```
robertkasza: I think you can use isNotEmptyValue here it will preserve the correct type. But i think this check is not the best because all i can say for sure about item after the check that it is not null. But I don't know probably TS would warn us about it somewhere else.

dashboard/src/features/orgs/projects/database/dataGrid/utils/buildRemoteFieldFromSelection/buildRemoteFieldFromSelection.ts
```diff
@@ -0,0 +1,93 @@
+import type { RemoteFieldArgumentMappingsByPath } from '@/features/orgs/projects/database/dataGrid/types/relationships/relationships';
```
robertkasza, line 1: Why are we adding new stuff to the remote-schemas feature? I think I know, but these are only used in the new edit relationship feature. But I am on the fence...

dashboard/src/features/orgs/projects/database/dataGrid/utils/parseRemoteFieldToSelection/parseRemoteFieldToSelection.ts
```diff
@@ -0,0 +1,98 @@
...
+  const [rootFieldName] = Object.keys(remoteField);
+  if (!rootFieldName) {
+    return {
+      rootFieldPath: '',
+      selectedFieldPaths,
+      argumentMappingsByPath,
+    };
+  }
+
+  const walk = (
```
robertkasza, line 43: Just out of curiosity, where does this function walk?

dashboard/src/features/orgs/projects/remote-schemas/utils/parseRemoteFieldToSelection.ts
```diff
@@ -0,0 +1,98 @@
...
+      const argEntries = Object.entries(args);
+      if (argEntries.length > 0) {
+        argumentMappingsByPath[currentPath] = argEntries.reduce<
+          Record<string, RemoteFieldArgumentMapping>
+        >((accumulator, [argumentName, argumentValue]) => {
+          const isColumnReference =
+            typeof argumentValue === 'string' && argumentValue.startsWith('$');
+
+          return {
+            ...accumulator,
```
robertkasza: I do not like this naming. It might as well be anything. CAn you give it a better(more descriptive) name?

dashboard/e2e/utils.ts
dbm03: I've just removed all of them in the last commit

dashboard/src/features/orgs/projects/database/dataGrid/components/AddSuggestedRelationshipDialog/AddSuggestedRelationshipDialog.tsx
dbm03: You're right, I've updated this so that it uses the same naming convention as the foreign keys. Thx

dashboard/src/features/orgs/projects/database/dataGrid/components/AddSuggestedRelationshipDialog/AddSuggestedRelationshipDialog.tsx
dbm03: I've just refactored all this so that there's a single `useCreateRelationshipMutation` hook for object and array relationships.

When you query suggestions Hasura returns either `object` or `array`, not the literal operation type that we have to send to create the relationship (e.g. `pg_create_array_relationship`). To derive the type I'm using this ternary expression. IIRC you said to use a Map, but I think it's simple enough that I can leave it like this (with the ternary expression):
```ts
const type =
  suggestion.type === 'array'
    ? 'pg_create_array_relationship'
    : 'pg_create_object_relationship';

// ...

await execPromiseWithErrorToast(
  async () => {
    await createRelationship({
      resourceVersion,
      args,
      type,
    });
  },
// ...
```
What do you think?

Regarding the create relationship by hand (without suggestions), I changed the values in the `Relationship Type` select to have the operation type (`pg_create_array_relationship` or `pg_create_object_relationship`)

dashboard/src/features/orgs/projects/database/dataGrid/components/AddSuggestedRelationshipDialog/AddSuggestedRelationshipDialog.tsx
dbm03: Done, thank you

dashboard/src/features/orgs/projects/database/dataGrid/components/RenameRelationshipDialog/RenameRelationshipDialog.tsx
dbm03: I've solved the styling issue by adding `inline-flex`

dashboard/src/features/orgs/projects/database/dataGrid/components/RenameRelationshipDialog/RenameRelationshipDialog.tsx
dbm03: Done thx

dashboard/src/features/orgs/projects/database/dataGrid/EditRelationshipsForm/EditRelationshipsForm.tsx
dbm03: I've just added this thx

dashboard/src/features/orgs/projects/database/dataGrid/components/BaseRelationshipDialog/SourceAndReferenceSelector.tsx
dbm03: I've added these two components `<ReadOnlySelect>` and `<ReadOnlyCombobox>`:

https://github.com/nhost/nhost/blob/feat/table-relationships/dashboard/src/components/presentational/ReadOnlySelect/ReadOnlySelect.tsx

https://github.com/nhost/nhost/blob/feat/table-relationships/dashboard/src/components/presentational/ReadOnlyCombobox/ReadOnlyCombobox.tsx

dashboard/src/features/orgs/projects/database/dataGrid/components/BaseRelationshipDialog/SourceAndReferenceSelector.tsx
dbm03: I'm using a SelectGroup now (only for styling purposes, the computed prefix to distinguish between remote schemas and databases names is still there). Thx

dashboard/src/features/orgs/projects/database/dataGrid/components/BaseRelationshipDialog/sections/TableRelationshipDetails.tsx
dbm03: I'm adding `preventRowFetching: true` here and in the `toTableData` table query as well. Thx

dashboard/src/features/orgs/projects/database/dataGrid/components/BaseRelationshipDialog/sections/RemoteSchemaRelationshipDetails.tsx
dbm03: I've added `preventRowFetching: true` here. Thx

dashboard/src/features/orgs/projects/database/dataGrid/components/BaseRelationshipDialog/sections/TableRelationshipDetails.tsx
dbm03: I've made it optional `label?:`, thx

dashboard/src/features/orgs/projects/database/dataGrid/components/BaseRelationshipDialog/BaseRelationshipDialog.tsx
dbm03: I've removed the default values from here. I understand the confusion, thank you

dashboard/src/features/orgs/projects/database/dataGrid/components/BaseRelationshipDialog/BaseRelationshipFormTypes.tsx
dbm03: I'll think more next time about where to place the form schemas. It is true that they're related to the types that you construct, but I see it's not ideal to mix functionality in a file that is called `-Types`. Also there's the `ReferenceSource` class that's also related with some types.

If you're not fine with this right now I can look. Otherwise, I'll just be careful about this inconvenience next time

dashboard/src/components/form/FormCombobox/FormCombobox.tsx
dbm03: I've removed this

dashboard/src/components/form/FormCombobox/FormCombobox.tsx
dbm03: I've added the options prop you suggest here. It's probably the best approach for now, it made the component a lot simpler. It's no longer needed to have a "enhanceChildren" function to add the `onChange` props to the command items, FormCombobox adds the props directly in the JSX.
Thanks.

dashboard/src/features/orgs/projects/database/dataGrid/hooks/useGetSuggestedRelationships/useGetSuggestedRelationships.ts
dbm03: I'm deleting `filterNotNullRelationshipSuggestionViewModel` and replacing it with `.filter(isNotEmptyValue);` directly. TS's fine with this. Thank you

dashboard/src/features/orgs/projects/database/dataGrid/components/DeleteRelationshipDialog/DeleteRelationshipDialog.tsx
dbm03: In a similar way to the `useCreateRelationshipMutation`, I've created a `useDeleteRelationshipMutation` file that handles deleting both local and remote relationships.

I'm using it like this:
```ts
    const type = isRemoteRelationship ? 'remote' : 'local';

    await execPromiseWithErrorToast(
      async () => {
        await deleteRelationship({
          resourceVersion: resourceVersion!,
          type,
          args: {
            relationshipName: relationshipToDelete,
            table: {
              schema,
              name: tableName,
            },
            source,
          },
        });
      },
...
```
I think it's simpler this way, thank you

dashboard/src/features/orgs/projects/database/dataGrid/components/BaseRelationshipDialog/BaseRelationshipFormTypes.tsx
dbm03: I'm adding the same validation that we have for table names (since it has the same rules):
```ts
  name: z
    .string()
    .min(1, { message: 'Name is required' })
    .regex(/^([A-Za-z]|_)+/i, {
      message: 'Name must start with a letter or underscore.',
    })
    .regex(/^\w+$/i, {
      message: 'Name must contain only letters, numbers, or underscores.',
    }),
```

Thank you

dashboard/src/features/orgs/projects/database/dataGrid/components/DeleteRelationshipDialog/DeleteRelationshipDialog.tsx
dbm03: Did you get the error `Metadata is not ready yet. Please try again in a moment.`? 
The version mismatch is a bit odd, since every mutation invalidates the `'export-metadata'` queryKey afterwards. The resourceVersion should be updated after that.

I've changed this specific part of the code so that it refetches the resource version before deleting the relationship:
```ts
  const handleDeleteDialogClick = async () => {
    setLoadingDelete(true);
    const { data: latestResourceVersion } = await refetchResourceVersion();
    const type = isRemoteRelationship ? 'remote' : 'local';

    await execPromiseWithErrorToast(
      async () => {
        await deleteRelationship({
          resourceVersion: latestResourceVersion!,
...
```
If this version mismatch error happens in other places I think I'll just add the refetch before every action. The bad thing is that this adds ~600ms to every mutation (the time it takes to fetch the metadata with the resourceVersion just before).

dashboard/src/features/orgs/projects/database/dataGrid/components/BaseRelationshipDialog/sections/TableRelationshipDetails.tsx
dbm03: Hasura doesn't require the mappings for object relationships (but does require them for array relationships). I don't think there's a use case where you'd want that, so I'm adding validation for both. 
Thanks

dashboard/src/features/orgs/projects/database/dataGrid/components/DataBrowserSidebar/DataBrowserSidebar.tsx
dbm03: I've just made untracked tables not show the `Edit Relationships` option in the dropdown. Thank you

By the way, I noticed that we also make the same mistake with the Edit Permissions and Edit settings for untracked tables, here's an issue I just created for this: https://github.com/nhost/nhost/issues/3860

dashboard/src/features/orgs/projects/database/dataGrid/components/BaseRelationshipDialog/sections/RemoteSchemaRelationshipDetails.tsx
dbm03: Just done and it works as well, thanks

dashboard/src/features/orgs/projects/database/dataGrid/components/BaseRelationshipDialog/sections/TableRelationshipDetails.tsx
dbm03: I'm adding preventRowFetching: true here and in the toTableData table query as well. Thx

dashboard/src/components/form/FormCombobox/FormCombobox.tsx
dbm03: It could if we added a CommandItem with value={1}, or something else. 
However, since this will get validated in the `options` prop as a string, it's not really necessary. 
The actual type of `value` is `PathValue<TFieldValues, TName>`, but I'm now using it directly (without doing the fieldValue type checking) and it works exactly the same way as a string, and TypeScript doesn't complain.

dashboard/src/features/orgs/projects/database/dataGrid/utils/buildRemoteFieldFromSelection/buildRemoteFieldFromSelection.ts
dbm03: Yeah this is only used it in the edit relationship feature. I placed it here because I thought it's more related to remote schemas utility functions (it uses a utility function from the remote schemas folder).

Anyway, I've moved this inside the database folders, since the purpose of this PR is to edit only that.

dashboard/src/features/orgs/projects/remote-schemas/utils/parseRemoteFieldToSelection.ts
dbm03: I'm changing the name to `argumentMappings` instead of accumulator

dashboard/src/features/orgs/projects/database/dataGrid/components/BaseRelationshipDialog/sections/RemoteSchemaRelationshipDetails.tsx
dbm03: I've extracted this into components, thank you

dashboard/src/features/orgs/projects/database/dataGrid/utils/parseRemoteFieldToSelection/parseRemoteFieldToSelection.ts
dbm03: When we want to edit an already existing relationship to a remote schema, we fetch a fieldmap from Hasura. That fieldmap has a depth that we don't know. There can be many nested elements, so this `walk` function traverses the JSON entirely and gets all the paths for every checkbox in the UI that is pre-selected, e.g. ("user", "user.profile", "user.profile.avatar"...).

We need it to traverse the list of checkboxes that need to be checked by default

dashboard/src/features/orgs/projects/database/dataGrid/components/BaseRelationshipDialog/sections/TableRelationshipDetails.tsx
robertkasza: That's good, then we can remove the prop here. or do we need it for some reason?

dbm03: I've just removed it, sorry about that

PR #3906, opened 2026-02-11
fix(dashboard): use functions instead of static only classes

robertkasza:
## Summary

Refactors static-only classes to use modern functional exports in data grid storage utilities, addressing Biome lint warnings and improving code maintainability.

## Key Changes

- **PersistentDataGridFilterStorage.ts**: Converted static-only class to named function exports
  - `getDataGridFilters()` and `saveDataGridFilters()` functions
- **PersistentDataTableConfigurationStorage.ts**: Converted static-only class to named function exports  
  - `getColumnVisibility()`, `saveColumnVisibility()`, `toggleColumnVisibility()`
  - `getColumnOrder()`, `saveColumnOrder()`
  - `getDataTableViewConfiguration()`, `saveRowDensity()`
- **File rename**: Fixed typo in filename `PersistenDataTableConfigurationStorage` → `PersistentDataTableConfigurationStorage`
- **Updated imports**: All 9 consuming components updated to use new named imports
- **Test updates**: Test file renamed and updated to test new function exports

## Technical Details

- Eliminates `noStaticOnlyClass` Biome lint warnings
- Maintains identical API contracts and TypeScript signatures
- No breaking changes to functionality - pure refactoring
- Improves tree-shaking potential with named exports
- All existing localStorage patterns and data validation preserved

### Checklist

- [x] No breaking changes (API remains identical)
- [ ] Tests pass (pending missing import fixes)
- [x] New features have new tests (N/A - refactoring only)
- [x] Documentation is updated (N/A - internal refactoring)

**Note**: Missing imports identified in review comments need to be addressed before merge.


dashboard/src/features/orgs/projects/storage/dataGrid/utils/PersistentDataTableConfigurationStorage.ts
```diff
@@ -0,0 +1,115 @@
...
+  localStorage.setItem(
+    COLUMN_CONFIGURATION_STORAGE_KEY,
+    JSON.stringify(updatedAllStoredData),
+  );
+}
+
+export function toggleColumnVisibility(tablePath: string, columnId: string) {
+  const columnVisibility = getColumnVisibility(tablePath);
+
+  const isVisible = !!columnVisibility[columnId];
```
dbm03: I think this should be instead:
```ts
const isVisible = columnVisibility[columnId] !== false;
```

Otherwise, the first toggle of a never-touched column doesn't persist false (hidden) to localStorage, it saves `true` instead. The first time you open a table it has a `undefined` value, so the `!!undefined` returns true, when it should be `false` instead.

To reproduce this:
1. Navigate to a table that you haven't opened before
2. Open the column customizer popover (the column visibility config).
3. Click a column's visibility checkbox once to hide a column that has never been toggled before (it's visible by default and has no entry in localStorage).
4. The column disappears from the table (fine)
5. Refresh the page.
6. The column reappears as visible, so the hidden column did not persist.

robertkasza: Good catch! But the issue was something else. I've pushed the changes. Please check.

Btw !!undefined is false because undefined is falsy and if we double negating it, it turns into false because the first ! turns undefined into true and the second turns it into false.

When you see  the !! pattern it is usually used to turn falsy/truthy values into boolean values. falsy -> false and truthy -> true.

dbm03: It looks good now, thanks for fixing it!

> Btw !!undefined is false because undefined is falsy and if we double negating it, it turns into false because the first ! turns undefined into true and the second turns it into false.

Do you have any style preference between `Boolean(undefined)` or `!!undefined`? just curious

robertkasza: It's more of a habit for me to use `!!`, but I think `Boolean(someVariable)` might be more readable. Need to think about it, but I might start using Boolean instead.

PR #3923, opened 2026-02-17
feat(dashboard): track/untrack tables in database page

dbm03:
### Checklist

- [x] No breaking changes
- [x] Tests pass  

### DIFF
--- a/dashboard/e2e/database/create-table.test.ts
+++ b/dashboard/e2e/database/create-table.test.ts
@@ -346,7 +346,6 @@ test('should be able to untrack and re-track a table', async ({
 
   await page.getByRole('button', { name: /back/i }).click();
 
-  await expect(page.getByText('Not tracked in GraphQL')).toBeVisible();
   await expect(page.getByRole('button', { name: /track now/i })).toBeVisible();
 
   await page.getByRole('button', { name: /track now/i }).click();


dashboard/src/features/orgs/projects/database/dataGrid/components/TrackTableButton/TrackTableButton.tsx
```diff
@@ -0,0 +1,72 @@
+import { useRouter } from 'next/router';
+import { ButtonWithLoading as Button } from '@/components/ui/v3/button';
+import { useGetMetadataResourceVersion } from '@/features/orgs/projects/common/hooks/useGetMetadataResourceVersion';
+import { useIsTrackedTable } from '@/features/orgs/projects/database/dataGrid/hooks/useIsTrackedTable';
+import { useSetTableTrackingMutation } from '@/features/orgs/projects/database/dataGrid/hooks/useSetTableTrackingMutation';
+import { execPromiseWithErrorToast } from '@/features/orgs/utils/execPromiseWithErrorToast';
+
+export default function TrackTableButton() {
+  const { query } = useRouter();
+  const { dataSourceSlug, tableSlug, schemaSlug } = query;
```
dbm03, line 10: question: is it fine to pull dataSourceSlug, schemaSlug, and tableSlug from useRouter() directly? Or should we pass it as props to TrackTableButton.

My reasoning is that this way we don't add useless prop that we can compute inside the component (since it's only meant to be used inside the data browser header)

dashboard/src/features/orgs/projects/common/hooks/useGetTrackedTablesNames/useGetTrackedTablesNames.ts
dbm03, line 1: note: deleted `useGetTrackedTablesNames.ts`, added `useGetTrackedTablesSet.ts` instead

dashboard/src/features/orgs/projects/database/dataGrid/hooks/useSetTableTrackingMutation/useSetTableTrackingMutation.ts
```diff
@@ -0,0 +1,64 @@
...
+    unknown,
+    UseSetTableTrackingMutationVariables
+  >({
+    mutationFn: ({ tracked, resourceVersion, args }) => {
+      const appUrl = generateAppServiceUrl(
+        project!.subdomain,
+        project!.region,
+        'hasura',
+      );
+      const base = {
```
robertkasza: `Tracked` and `args` passed to both setTable function, shouldn't they be included here too? And rename the variable to something else, maybe?

dashboard/src/features/orgs/projects/database/dataGrid/components/TrackTableButton/TrackTableButton.tsx
robertkasza: Yes, it's okay to use the query.

dashboard/src/features/orgs/projects/database/dataGrid/components/EditGraphQLSettingsForm/EditGraphQLSettingsForm.tsx
```diff
@@ -22,19 +23,21 @@ export interface EditTableSettingsFormProps {
...
-}: EditTableSettingsFormProps) {
-  const handleCancel = () => {
-    onCancel?.();
-  };
-
+}: EditGraphQLSettingsFormProps) {
   return (
     <div className="flex h-full min-h-0 flex-col overflow-hidden">
       <div className="flex min-h-0 flex-1 flex-col gap-4 overflow-y-auto pb-4">
+        <TrackUntrackSection
```
robertkasza, line 79: What should happen if the user toggles the tracking here? The form should be disabled, right? I mean saving will not succeed anyway. What do you think?

dashboard/src/features/orgs/projects/database/dataGrid/components/EditGraphQLSettingsForm/EditGraphQLSettingsForm.tsx
```diff
@@ -22,19 +23,21 @@ export interface EditTableSettingsFormProps {
   disabled?: boolean;
 }
 
-export default function EditTableSettingsForm({
+export default function EditGraphQLSettingsForm({
```
robertkasza, line 31: We are hiding relationships when  the table is not tracked. And we renamed this form to graphqlSettingsForm.

This is just an idea, can't we merge somehow the relationship and graphql setting forms? No need to implement it just want to know your opinion.

dashboard/src/features/orgs/projects/database/dataGrid/components/EditGraphQLSettingsForm/EditGraphQLSettingsForm.tsx
robertkasza: We really should disable the form while the tracking request is in progress.

dashboard/src/features/orgs/projects/database/dataGrid/components/EditGraphQLSettingsForm/EditGraphQLSettingsForm.tsx
dbm03: I renamed this to "GraphQL Settings" because "Edit Settings" was really undescriptive, it could mean anything. In my opinion it makes more sense to have relationships inside this GraphQL settings form. 

Permissions are also related to GraphQL settings, but they have their own dropdown item. However, it's a way older feature so I would leave it as is. Most users are probably used to where it's currently at, with a separated dropdown item.

I can ask the others and we'll decide what to do with the Relationships. I agree with you

dashboard/src/features/orgs/projects/database/dataGrid/hooks/useSetTableTrackingMutation/useSetTableTrackingMutation.ts
dbm03: I've renamed this to `commonParams` and included `tracked` and `args`

dashboard/src/features/orgs/projects/database/dataGrid/components/EditGraphQLSettingsForm/EditGraphQLSettingsForm.tsx
dbm03: I've disabled the other sections when it's untracked, or tracking/untracking (pending mutation).

Also, I've added a tooltip so you know that you can't edit/submit the form because the table is untracked

Screenshot 2026-02-25 at 16 03 37

dashboard/src/features/orgs/projects/database/dataGrid/components/EditGraphQLSettingsForm/EditGraphQLSettingsForm.tsx
dbm03: I'm resolving this comment. We'll keep them in a different dropdown

dbm03:
> I know this is a "straightforward" feature, but I really would like to see some tests written.
> 
> Let's be as thorough as possible, but do not go over board. I just want to make sure that we have some mechanism to catch if we break something in the future.

I've added an e2e test that creates a table, untracks it from the Edit GraphQL settings drawer, and then clicks on the "Track Now" button in the data grid. I think that should cover most of the functionality.

Also, I've added tests for the setTableTrackingMigration and the TrackUntrackSection component.

PR #3924, opened 2026-02-17
fix(dashboard): escape with encodeURIcomponent

nunopato:
## Summary

Fixes URL encoding issue in service configuration sharing by properly escaping base64-encoded config data in query parameters.

## Changes

- Updated ServiceForm.tsx:234 to wrap base64Config with encodeURIComponent() when constructing the one-click install URL
- Prevents potential URL parsing issues when the base64-encoded service configuration contains URL-unsafe characters (like +, /, or =)

## Technical Details

The service configuration is JSON-stringified, base64-encoded, and included as a query parameter in the one-click install URL. Without proper URI encoding, special characters in the base64 string could cause URL parsing issues or data corruption when the URL is shared or processed by different systems.

**Before:** config=${base64Config}
**After:** config=${encodeURIComponent(base64Config)}


dbarrosop:
Should we use base64 URL encoding instead? Also, we should make this backwards compatible to not break existing links

nunopato:
@dbarrosop `encodeURIComponent` seems to be the standard way to put random values into URL query params, as the browser/Next.js automatically decodes it when reading router.query.

There should be no backwards compatibility issue because this only fixes the encoding for already broken/unsupported base64 strings with weird chars (e.g. + -). For existing working links nothing changes as the decoding part is done automatically.

dbarrosop:
I am not sure what next.js does but the standard way is using base64url (standard as in there is an RFC for it and it's what's used in other protocols where data needs to be encoded and passed in a url). If we want to stick to next.js semantics that's fine but the standard way regardless of framework is using base64url encoding which has the added benefit of being nicer and shorter than using regular URI encoding on top of base64 (i.e. no %XX% for the padding or other html-specific characters).

PR #3954, opened 2026-02-24
fix(dashboard): clear query cache on project switch

### LINKED ISSUE #3891: dashboard: query cache doesn't revalidate when navigating to a different project
**Describe the bug**
If you navigate from one project to another in your navigation tree, sometimes you can still see the data from the previous project instead of the project you've just selected. 
This happens to the sidebar with table names in database view, storage view and logs inside events' view.

**To Reproduce**
Example (database view)
1. Go to Database in a project A
2. Click on a project A's table in the sidebar
3. Now Go to Database in project B
4. You see project A's tables in the sidebar, even though you're on project B's database page (refreshing the page solves it)

**To Solve**
Make sure we add `project.subdomain` to queryKey, so that when we switch to a different project the information in the query client's cache is invalidated and we trigger network requests to refresh the data accordingly.

**Expected behavior**
Query client's cache should be invalidated when switching projects

dbm03:
Resolves #3891

### Checklist

- [X] No breaking changes
- [X] Tests pass

### **PR Type**
Bug fix

___

### **Description**
- Clear the TanStack Query cache when the user navigates between projects so stale data from the previous project does not leak into the new project view.
- Use the existing `usePreviousData` hook to detect when `appSubdomain` changes and call `queryClient.clear()` only on actual project switches — not on initial render.
- Add two new test cases verifying cache-clearing behavior: one asserting the cache is **not** cleared on initial render, and another asserting it **is** cleared when switching projects.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["Router (appSubdomain change)"] --> B["usePreviousData hook"]
  B --> C{"previousAppSubdomain !== appSubdomain\n&& previousAppSubdomain !== undefined?"}
  C -- Yes --> D["queryClient.clear()"]
  C -- No --> E["No-op"]
```

<tr><td><strong>Core logic</strong></td><td><details><summary>1 file</summary><table>
<tr>
  <td><strong>ProjectViewWithState.tsx</strong><dd><code>Add useEffect that clears query cache when appSubdomain changes between projects</code></dd></td>
  <td><a href="https://github.com/nhost/nhost/pull/3954/files#diff-552483dc05a5828b39b1f92f1c4b72977d7916055f6e185c88b029ea7240ff7c">+15/-1</a></td>
</tr>
</table></details></td></tr>
<tr><td><strong>Tests</strong></td><td><details><summary>1 file</summary><table>
<tr>
  <td><strong>ProjectViewWithState.test.tsx</strong><dd><code>Add tests for cache-clearing on project switch and no-op on initial render; parameterize router helper</code></dd></td>
  <td><a href="https://github.com/nhost/nhost/pull/3954/files#diff-2fe1035846f048d8d1063c2b894047f7b596fe7e164990c477f0ab94522a78ca">+43/-6</a></td>
</tr>
</table></details></td></tr>
</tbody></table>

</details>

___

### DIFF
--- a/dashboard/src/features/orgs/layout/OrgLayout/ProjectLayoutContent.test.tsx
+++ b/dashboard/src/features/orgs/layout/OrgLayout/ProjectLayoutContent.test.tsx
@@ -13,7 +13,6 @@ import {
   screen,
   waitFor,
 } from '@/tests/testUtils';
-import { ApplicationStatus } from '@/types/application';
 import ProjectLayoutContent from './ProjectLayoutContent';
 
 const mocks = vi.hoisted(() => ({
@@ -35,23 +34,19 @@ function TestComponent() {
 
 const server = setupServer(tokenQuery, getProjectStateQuery());
 
-const getUseRouterObject = ({
-  route = '/orgs/[orgSlug]/projects/[appSubdomain]',
-  appSubdomain = 'test-project',
-}: {
-  route?: string;
-  appSubdomain?: string;
-} = {}) => ({
+const getUseRouterObject = (
+  route: string = '/orgs/[orgSlug]/projects/[appSubdomain]',
+) => ({
   basePath: '',
-  pathname: `/orgs/xyz/projects/${appSubdomain}`,
+  pathname: '/orgs/xyz/projects/test-project',
   route,
-  asPath: `/orgs/xyz/projects/${appSubdomain}`,
+  asPath: '/orgs/xyz/projects/test-project',
   isLocaleDomain: false,
   isReady: true,
   isPreview: false,
   query: {
     orgSlug: 'xyz',
-    appSubdomain,
+    appSubdomain: 'test-project',
   },
   push: mocks.push,
   replace: vi.fn(),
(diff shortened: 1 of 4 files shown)


dbm03:
> Not sure if this is the best solution. Keys for caching should be unique to the project. But should be okay for now.

Thank you. I thought about it and yeah, the best would be to prepend `'<appSubdomain>'` to every query. However, that would only save requests if we switch projects back and forth on the same tab, and it's also a more tedious solution.

I thought about adding a wrapper of useQuery, called something like useProjectQuery: this would handle the prepend of appSubdomain.

But yeah, this PR with a small change fixes this for now

robertkasza:
> However, that would only save requests if we switch projects back and forth on the same tab

I am not sure I understand what you mean by this.

dbm03:
> I am not sure I understand what you mean by this.

If we added the subdomain to the queryKey, we wouldn't have to fetch requests again when we switch projects. We could re-use the query cache that we had.
E.g.:
1. You're on project A and you've loaded `"<projectASubdomain>.default.public.users"`. 
2. You navigate to project B and you load `"<projectBSubdomain>.default.public.users"`. 
3. Then, when you navigate back to project A, you haven't lost the cache from project A and you don't have to re-fetch the data again (as long as it's not over the stale time)

Thus, that'd only save network requests when you switch projects. If you stay on the same project you wouldn't notice a difference

robertkasza (approved):

PR #3962, opened 2026-02-27
feat(lazyreview): add syntax highlighting to diff viewer

dbm03:
Use chroma/v2 to tokenize diff lines per-file with language detection from file paths. Syntax-colored foreground tokens are overlaid on subtle dark backgrounds for added/removed lines, similar to GitHub's dark-mode diff view. Results are cached per file path so View() stays fast.

TODO: check how `chroma/regexp2` was added to `go.mod`


dbm03:
Feel free to merge this PR @dbarrosop

PR #4085, opened 2026-03-29
fix(auth): select primary verified email from GitHub OAuth

JeffreyUrban:
## Summary

The GitHub OAuth provider calls `/user/emails` to get the user's email addresses, but blindly uses `emails[0]` — the first entry in the array. GitHub's API does not guarantee ordering, so users can end up with a non-primary (or even unverified) email stored in their Nhost account.

This PR:
- Adds the `Primary` field to the `gitHubEmail` struct (previously not parsed from the API response)
- Selects the primary + verified email first
- Falls back to first verified email, then first email overall

## Context

This is a follow-up to [hasura-auth#658](https://github.com/nhost/hasura-auth/pull/658), which added the `/user/emails` API call but took `emails[0]` without filtering. The automated reviewer [flagged this](https://github.com/nhost/hasura-auth/pull/658) at the time, but it was merged without addressing the comment.

GitHub API reference: https://docs.github.com/en/rest/users/emails#list-email-addresses-for-the-authenticated-user

The response includes `email`, `verified`, `primary`, and `visibility` fields per entry. Only `primary` was missing from the struct.

## Test plan

- [ ] Sign in with a GitHub account that has multiple email addresses where the primary email is not first in the API response
- [ ] Verify the primary verified email is stored in `auth.users.email`
- [ ] Sign in with a GitHub account that has no primary email set — verify fallback to first verified email

### **PR Type**
Bug fix

___

### **Description**
- Fix GitHub OAuth email selection to prefer the user's primary verified email instead of blindly using the first email in the API response
- Add `Primary` field to `gitHubEmail` struct to parse the `primary` flag from GitHub's `/user/emails` endpoint
- Implement a three-tier fallback: primary+verified → first verified → first email

___

<tr><td><strong>Core logic</strong></td><td><details><summary>1 file</summary><table>
<tr>
  <td><strong>github.go</strong><dd><code>Add Primary field to gitHubEmail struct and implement email selection logic with fallback chain</code></dd></td>
  <td><a href="https://github.com/nhost/nhost/pull/4085/files#diff-5f724740af1a1095b598582f9abc277f0531ec3c644b46c490fb04eddac6b1f7">+21/-2</a></td>
</tr>
</table></details></td></tr>
</tbody></table>

</details>

___

### DIFF
--- a/services/auth/go/providers/github.go
+++ b/services/auth/go/providers/github.go
@@ -47,10 +47,39 @@ type gitHubUser struct {
 	AvatarURL string `json:"avatar_url"`
 }
 
-type gitHubEmail []struct {
+type gitHubEmailEntry struct {
 	Email    string `json:"email"`
 	Verified bool   `json:"verified"`
 	Primary  bool   `json:"primary"`
+}
+
+type gitHubEmail []gitHubEmailEntry
+
+// selectEmail picks the best email from a GitHub user's email list.
+// Priority: primary+verified > first verified > first email.
+func selectEmail(emails gitHubEmail) gitHubEmailEntry {
+	selected := emails[0]
+	primarySeen := false
+
+	for _, e := range emails {
+		if e.Primary && e.Verified {
+			return e
+		}
+
+		if e.Primary {
+			primarySeen = true
+		}
+
+		if e.Verified && !selected.Verified {
+			selected = e
+		}
+
+		if selected.Verified && primarySeen {
+			break
+		}
+	}
+
+	return selected
 }
 
 func (g *Github) GetProfile(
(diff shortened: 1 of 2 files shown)


dbarrosop (changes requested):
Thanks, pushed a fix to a bug and some unit tests.

services/auth/go/providers/github.go
```diff
@@ -83,10 +84,28 @@ func (g *Github) GetProfile(
 		return oidc.Profile{}, errors.New("GitHub user has no email addresses") //nolint:err113
 	}
 
+	// Select the primary verified email. Fall back to first verified,
+	// then first email if none are verified.
+	selected := emails[0]
+	for _, e := range emails {
```
dbarrosop: I think there is a bug here, if the primary isn't verified and no other email is verified either we will select emails[0] instead of the primary. We can fix that bug and avoid two loops by doing the following instead:

```go
	selected := emails[0]
	primarySeen := false

	for _, e := range emails {
		if e.Primary && e.Verified {
			selected = e

			break
		}

		if e.Primary {
			primarySeen = true
		}

		if e.Verified && !selected.Verified {
			selected = e
		}

		if selected.Verified && primarySeen {
			break
		}
	}
```

dbarrosop (approved):

PR #4160, opened 2026-04-17
fix(cli): Use `$$` when `$` is in an env var for Docker Compose

CaspianChaharom:
From the [docker compose documentation](https://docs.docker.com/reference/compose-file/interpolation/#:~:text=double-dollar%20sign) "You can use a $$ (double-dollar sign) when your configuration needs a literal dollar sign."

This is an issue for commands like `nhost up cloud` because the Hasura admin secret (used in a few places as an env var `HASURA_GRAPHQL_ADMIN_SECRET` and `NHOST_ADMIN_SECRET` for `nhost/graphql-engine`, and `NEXT_PUBLIC_NHOST_ADMIN_SECRET` for `nhost/dashboard`) may contain a dollar sign, which is directly written to the environment variable in `.nhost/docker-compose.yaml` where it is not interpreted correctly, causing `nhost up cloud` to fail.

### Checklist

- [X] No breaking changes
- [x] Tests pass

### **PR Type**
Bug fix

___

### **Description**
- Introduce a dedicated `Environment` map type for Docker Compose service environment variables that escapes literal `$` to `$$` during YAML marshalling, preventing Docker Compose from interpreting `$` in values (e.g., database passwords) as variable substitution.
- Switch the `Service.Environment` field from `map[string]string` to the new `Environment` type so escaping happens transparently at marshal time without mutating the source map.
- Add a unit test verifying that `$` characters in environment values are escaped to `$$` in the produced YAML and that plain values pass through unchanged.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["Service struct<br/>(Environment field)"] --> B["yaml.Marshal"]
  B --> C["Environment.MarshalYAML"]
  C --> D["Replace $ → $$<br/>(copy, no mutation)"]
  D --> E["docker-compose.yaml"]
  E --> F["Docker Compose<br/>passes literal $ to container"]
```

<tr><td><strong>Core logic</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>compose.go</strong><dd><code>Add Environment type with MarshalYAML that escapes $ to $$; change Service.Environment to this new type.</code></dd></td>
  <td><a href="https://github.com/nhost/nhost/pull/XXX/files#diff-cfe5848f846b5d5bcfeb5bf52e76d2c23c08e4da760f2b85c1dd3ed01d0a635b">+15/-1</a></td>
</tr>
</table></details></td></tr>
<tr>
  <td><strong>compose_test.go</strong><dd><code>Add regression test asserting that $ in environment values is escaped to $$ in marshalled YAML.</code></dd></td>
  <td><a href="https://github.com/nhost/nhost/pull/XXX/files#diff-a69e2bbe1008aa647eb6a88bf6614e4b3dcc5988db54012ad2bca6dc85b87f7f">+50/-0</a></td>
</tr>
</table></details></td></tr>
</tbody></table>

</details>

___


dbarrosop:
Thanks for the PR. I think mutating the variable in-place isn't entirely safe. I would suggest the following instead:

1. Create a type like:

```go
type EnvValue string

func (e EnvValue) MarshalYAML() (any, error) {
    return strings.ReplaceAll(string(e), "$", "$$"), nil
}
```

2. We use it in [compose.go](https://github.com/nhost/nhost/blob/main/cli/dockercompose/compose.go#L67) we use it like:

```go
Environment map[string]EnvValue `yaml:"environment,omitempty"`
```

That way we do the escaping when marshaling into yaml without mutating anything.

CaspianChaharom:
It looks like that exactly isn't allowed in Go (I'm not a Go programmer though) unless every assignment is wrapped in `EnvValue` throughout the codebase because
```cannot use env (variable of type map[string]string) as map[string]EnvValue value in struct literal```

But declaring the full type `type Environment map[string]string` seems to allow the same thing without needing to wrap everything.

Would you rather have the type be `map[string]EnvValue` and wrap all assignments, or have the type be `Environment` (which itself is a map) and not need to wrap all the assignments?

**Comment by dbarrosop** [comment]
Ok, feel free to give it a go with `type Environment map[string]string`, thanks!

**Comment by CaspianChaharom** [comment]
I implemented the change and tested locally, and it works.

**Comment by dbarrosop** [comment]
perfect, could you add a test to avoid regressions in the future. For instance, create a file `compose_test.go` with:

```go
package dockercompose_test

import (
	"testing"

	"github.com/google/go-cmp/cmp"
	"github.com/nhost/nhost/cli/dockercompose"
	"gopkg.in/yaml.v3"
)

func TestServiceEnvironmentEscapesDollar(t *testing.T) {
	t.Parallel()

	svc := &dockercompose.Service{
		Image: "nhost/example:1.0",
		Environment: dockercompose.Environment{
			"PLAIN":               "no-dollar",
			"HASURA_ADMIN_SECRET": "secret$with$dollars",
		},
		ExtraHosts: []string{"host.docker.internal:host-gateway"},
		Restart:    "always",
	}

	got, err := yaml.Marshal(svc)
	if err != nil {
		t.Fatalf("marshal failed: %v", err)
	}

	expected := `image: nhost/example:1.0
environment:
    HASURA_ADMIN_SECRET: secret$$with$$dollars
    PLAIN: no-dollar
extra_hosts:
    - host.docker.internal:host-gateway
restart: always
`

	if diff := cmp.Diff(expected, string(got)); diff != "" {
		t.Error(diff)
	}
}
```

or similar. Thanks!

CaspianChaharom:
Added it. Let me know if that is good.

dbarrosop (approved):

PR #4228, opened 2026-05-01
fix(dashboard): Display boolean values in CustomCheckEditor

robertkasza:
### Checklist

- [x] No breaking changes
- [x] Tests pass

### **PR Type**
Bug fix

___

### **Description**
- Fixes the variable-combobox trigger label in `CustomCheckEditor` so non-string `comboboxValue`s (e.g. boolean `true`/`false`) render as readable text instead of falling back to the placeholder or breaking the trigger.
- Introduces a small `getDisplayValue` helper that returns strings as-is, treats `null`/`undefined` as empty so the `'Select variable...'` placeholder still wins, and `String()`-coerces other primitives.
- Adds a parametrized vitest case covering boolean, string, `null`, and `undefined` values to lock in the trigger label rendering.

___

<tr><td><strong>Core logic</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>ConditionValue.tsx</strong><dd><code>Add getDisplayValue helper to safely stringify non-string combobox values for the trigger label.</code></dd></td>
  <td><a href="https://github.com/nhost/nhost/pull/4228/files#diff-810657158af1493e2b841f61a1dfa02c99077838c527e860607fd0c5ca311f08">+13/-1</a></td>
</tr>
</table></details></td></tr>
<tr>
  <td><strong>ConditionValue.test.tsx</strong><dd><code>Add parametrized trigger-label test covering boolean, string, null, and undefined values.</code></dd></td>
  <td><a href="https://github.com/nhost/nhost/pull/4228/files#diff-19fa3f88d3784d5a287e92b6b48ca0e9542916589235e0a00372d75f79abd5a4">+25/-0</a></td>
</tr>
</table></details></td></tr>
</tbody></table>

</details>

___

### DIFF
--- a/dashboard/src/features/orgs/projects/database/dataGrid/components/CustomCheckEditor/ConditionValue.tsx
+++ b/dashboard/src/features/orgs/projects/database/dataGrid/components/CustomCheckEditor/ConditionValue.tsx
@@ -233,11 +233,23 @@ function ConditionValue({
     );
   }
 
+  function getDisplayValue(value: unknown): string {
+    if (typeof value === 'string') {
+      return value;
+    }
+    if (value === null || value === undefined) {
+      return '';
+    }
+    return String(value);
+  }
+
   const selectedVariable = availableHasuraPermissionVariables.find(
     (variable) => variable.value === comboboxValue,
   );
   const comboboxLabel =
-    selectedVariable?.label || comboboxValue || 'Select variable...';
+    selectedVariable?.label ||
+    getDisplayValue(comboboxValue) ||
+    'Select variable...';
 
   return (
     <Popover open={open} onOpenChange={setOpen}>
--- a/dashboard/src/features/orgs/projects/database/dataGrid/components/CustomCheckEditor/ConditionValue.test.tsx
+++ b/dashboard/src/features/orgs/projects/database/dataGrid/components/CustomCheckEditor/ConditionValue.test.tsx
@@ -229,4 +229,32 @@ describe('ConditionValue', () => {
       },
     );
   });
+
+  describe('combobox trigger label', () => {
+    beforeEach(() => {
+      mocks.useGetRolesPermissionsQuery.mockImplementation(() => ({
+        data: mockPermissionsData,
+        loading: false,
+      }));
+    });
+
+    it.each([
+      { value: true, expected: 'true' },
+      { value: false, expected: 'false' },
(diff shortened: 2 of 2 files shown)


dashboard/src/features/orgs/projects/database/dataGrid/components/CustomCheckEditor/ConditionValue.tsx
```diff
@@ -236,8 +278,14 @@ function ConditionValue({
   const selectedVariable = availableHasuraPermissionVariables.find(
     (variable) => variable.value === comboboxValue,
   );
+  const displayValue =
+    typeof comboboxValue === 'string'
+      ? comboboxValue
+      : comboboxValue == null
+        ? ''
+        : String(comboboxValue);
```
dbm03: nit: nested ternaries are usually advised against

dashboard/src/features/orgs/projects/database/dataGrid/components/CustomCheckEditor/ConditionValue.tsx
```diff
@@ -236,8 +278,14 @@ function ConditionValue({
...
   const comboboxLabel =
-    selectedVariable?.label || comboboxValue || 'Select variable...';
+    selectedVariable?.label || displayValue || 'Select variable...';
```
dbm03: Before this PR's changes, we used to show in the UI whether the selected value was something different to `true|false`, e.g. "X-Hasura-User-Id".

```json
{
  "booleano": {
    "_eq": "X-Hasura-User-Id"
  }
}
```

Before:
Image

After (this PR's branch):
Image
The value seems empty but it's actually "X-Hasura-User-Id" if you go to the JSON tab

I believe it'd be better to show the actual selected value (even if it's not selectable in the `<Select>` component)

robertkasza: I think you are right, but that check would never works. But we need to support env variables.

robertkasza: Good catch by the way. I will revert the changes and fix the issue that i noticed that true is not displayed.
image

dbm03 (approved):

PR #4246, opened 2026-05-05
fix(ci): use the environment NHOST_PAT directly

meh:
### Checklist

- [X] No breaking changes
- [X] Tests pass

### Description

This avoids dirtying the tree when running `make check` for `cli`, the derivation was already impure since the `Makefile` was editing the script generating the access token.

### DIFF
--- a/cli/Makefile
+++ b/cli/Makefile
@@ -1,10 +1,6 @@
 ROOT_DIR?=$(abspath ../)
+NIX_CHECK_EXTRA_ARGS=--impure
 include $(ROOT_DIR)/build/makefiles/general.makefile
-
-
-.PHONY: _check-pre
-_check-pre:
-	@sed -i 's/$$NHOST_PAT/$(NHOST_PAT)/' get_access_token.sh
 
 
 .PHONY: _dev-env-up
--- a/build/makefiles/general.makefile
+++ b/build/makefiles/general.makefile
@@ -63,10 +63,13 @@ develop:  ## Start a nix develop shell
 _check-pre:  ## Pre-checks before running nix flake check
 
 
+NIX_CHECK_EXTRA_ARGS?=
+
 .PHONY: check
 check: _check-pre ## Run nix flake check
 	nix build \
 		--print-build-logs \
+		$(NIX_CHECK_EXTRA_ARGS) \
 		.\#checks.$(ARCH)-$(OS).$(NAME)
 
 
@@ -75,6 +78,7 @@ check-dry-run:  ## Returns the derivation of the check
 	@nix build \
 		--dry-run \
 		--json \
+		$(NIX_CHECK_EXTRA_ARGS) \
 		.\#checks.$(ARCH)-$(OS).$(NAME) | jq -r '.[].outputs.out'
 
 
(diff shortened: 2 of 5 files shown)


cli/Makefile
```diff
@@ -2,9 +2,21 @@ ROOT_DIR?=$(abspath ../)
 include $(ROOT_DIR)/build/makefiles/general.makefile
 
 
-.PHONY: _check-pre
-_check-pre:
-	@sed -i 's/$$NHOST_PAT/$(NHOST_PAT)/' get_access_token.sh
+.PHONY: check
```
dbarrosop: I am not a fan of this change as this service now deviates from the rest and global fixes won't automatically apply

cli/Makefile
```diff
@@ -2,9 +2,21 @@ ROOT_DIR?=$(abspath ../)
...
+check:  ## Run nix flake check (--impure for NHOST_PAT)
+	nix build \
+		--print-build-logs \
+		--impure \
```
dbarrosop: This has the potential of leading to other issues as other environment variables leak into it. It also disables the caching of the derivation if I am not mistaken

meh: As far as I know the environment variables only leak through to the nix derivation itself (so one can use `builtins.getEnv`) and then only `NHOST_PAT` is passed through to the build. As far as caching goes, since a file in the source was modified before building, the derivation was only cached per NHOST_PAT (which still happens since the environment variable becomes part of the source) so in effect nothing should have changed behaviorally.

cli/Makefile
meh: Would setting an `IMPURE` variable in check-pre and then adding `--impure` from `general.makefile` if that is set be a better option?

dbarrosop: yeah, I was actually thinking in something like `NIX_CHECK_EXTRA_ARGS=--impure` to be a bit more generic and allow for potential future use

cli/Makefile
dbarrosop: yeah, this doesn't work exactly the way I thought, let me do a couple of more tests and read a bit more on it to understand better the implications

cli/Makefile
dbarrosop: hold onto this though, because if we go down this path we may want to fix the `make get-version VER=xxxx` hack with this mechanism too, in which case `--impure` globally might make sense

meh: Ideally `--impure` is only used where it's actually needed, since it's an indicator the builds are depending on external stuff (time, environment, whatnot) and won't be reproducible, the version stuff is still dependent on the source itself so it shouldn't need to be impure methinks, while the NHOST_PAT stuff is completely external.

cli/Makefile
dbarrosop: ok, we can disregard this issue

cli/Makefile
dbarrosop: ok, let's go with the `NIX_CHECK_EXTRA_ARGS`, the get-version thing is for the build anyway so it doesn't apply here.

dbarrosop (approved):

PR #4260, opened 2026-05-07
fix(dashboard): Improve logs page

### LINKED ISSUE #3353: dashboard: improve logs page virtualization
Virtual rows in the logs page aren't properly positioned, causing all rows to render instead of only visible ones. This impacts performance with large log datasets.

Fix needed: Look into absolute positioning with transforms to virtual rows.

robertkasza:
### Checklist

- [ ] No breaking changes
- [ ] Tests pass

### **PR Type**
Enhancement

___

### **Description**
- Refactors `LogsBody` into smaller modules (`LogsTable`, `LogsSearchBar`, `LogsDetailSheet`, `columns`, `types`, `LogsBodyCustomMessage`) and introduces a dedicated in-table search/filter experience with match navigation and "filter to matches" mode.
- Adds in-table search via the new `useLogSearch` hook plus a `LogsSearchBar` component, with prev/next navigation, match counter, Esc-to-clear, and a global Cmd/Ctrl+F shortcut handled by `useGlobalSearchShortcut`.
- Adds a side-sheet detail view (`LogsDetailSheet`) opened by an actions-column button on each row; the log body is JSON-pretty-printed when possible and copyable.
- Introduces severity detection (`severity.ts`) producing a new "Level" column, and service-styled chips (`serviceStyle.ts`) used both in the table and the detail sheet.
- Adds a syntax-highlighter (`highlightLog`) for UUIDs, trace/span IDs, JSON/logfmt keys, durations, HTTP methods, and status codes, with search-range overlays driving scroll-to-current behavior.
- Adds a new `filters` prop on `LogsBody` that is used as a React `key` (via `JSON.stringify`) to reset internal search/selection state when the caller's filters change; updated all three call sites (logs page, deployments, serverless functions).
- Renames the regex filter placeholder from "Filter logs ..." to "Search logs ..." and updates the affected tests.


dashboard/src/features/orgs/projects/logs/components/LogsSearchBar/LogsSearchBar.tsx
```diff
@@ -0,0 +1,135 @@
...
+          size="icon"
+          aria-pressed={filterMode}
+          className={cn(
+            'h-8 w-8',
+            filterMode && 'bg-primary/10 text-primary',
+          )}
+          onClick={onToggleFilter}
+          aria-label={filterMode ? 'Show all rows' : 'Filter to matching rows'}
+        >
+          <ListFilter className="h-4 w-4" />
```
dbm03, line 97: I believe it'd be nice to add a tooltip when we hover over this button. The first time I saw it I didn't know what it would do. Maybe it could say "Toggle row filter"

Image

dashboard/src/features/orgs/projects/logs/components/LogsDetailSheet/LogsDetailSheet.tsx
```diff
@@ -0,0 +1,96 @@
...
+        if (!next) {
+          onClose();
+        }
+      }}
+    >
+      <SheetContent
+        side="right"
+        className="flex w-[600px] flex-col gap-4 text-base text-foreground sm:max-w-[90vw]"
+        onPointerDownOutside={(event) => event.preventDefault()}
+        onInteractOutside={(event) => event.preventDefault()}
```
dbm03, line 53: This appears in console when you open the log details sheet
```
Warning: Missing `Description` or `aria-describedby={undefined}` for {DialogContent}.
```
If you don't want to add a `<SheetDescription>` I think passing `aria-describedby={undefined}` here should be fine with Radix

dashboard/src/features/orgs/projects/logs/components/LogsBody/LogsBody.tsx
```diff
@@ -43,84 +41,107 @@ export interface LogsBodyProps {
...
+    tableData,
+    matches,
+    rangesByRow,
+    totalMatches,
+    currentMatch,
+    setQuery,
+    setCurrentMatchIndex,
+    toggleFilter,
+    clear: clearSearch,
+  } = useLogSearch(data);
```
dbm03, line 85: I find it odd that when you type something in the filter logs input, this doesn't automatically scroll to the result, until you press Enter (so you go to next element).

I'm used to the browser's <kbd>CTRL + F</kbd> that directly scrolls to the first finding (without having to press <kbd>ENTER</kbd> first)

If you add the following code here you can see how it scrolls directly after the first match with your search input string is found:
```ts
useEffect(() => {
  if (matches.length === 0) {
    return;
  }
  tableRef.current?.scrollToRow(matches[0].rowIndex);
}, [matches]);
```

Adding useEffect is not ideal for this either, but it's an example if you want to try what I mean.

This is a bit nitpicky, so if you don't have a preference feel free to leave it as is. No big deal

dbm03: It's fine now, thx!

dbm03 (approved):

PR #4280, opened 2026-05-13
fix(dashboard): route useGetRolesPermissionsQuery through localMimirClient in ColumnPresetsSection

### LINKED ISSUE #4279: dashboard: local dashboard, Column preset doesn't show custom session variables
### Bug

In the table permissions editor, the **Column presets** value combobox does not show custom session variables when the dashboard is running locally (`!isPlatform`).

### Steps to reproduce

1. Run the dashboard locally against a local Nhost stack.
2. Go to **Settings → Roles and Permissions** and add a custom session variable (e.g. `X-Hasura-Org-Id`).
3. Go to **Database → any table → Permissions**, pick a role with action INSERT, expand **Column presets**, and open the value combobox.

### Expected

The custom variable appears alongside `X-Hasura-User-Id`, like it does in the Row permissions custom check editor on the same screen.

### Actual

Only the system variable `X-Hasura-User-Id` is listed.

### Cause

`useGetRolesPermissionsQuery` in `dashboard/src/features/orgs/projects/database/dataGrid/components/EditPermissionsForm/sections/ColumnPresetsSection.tsx` does not route through `localMimirClient` when `!isPlatform`, so it queries the wrong config source locally.

HirenGajjar:
## Problem
In the Column presets section of the table permissions editor, the value combobox was not showing custom session variables (e.g. `X-Hasura-Org-Id`) when the dashboard runs locally (`!isPlatform`). Only the default `X-Hasura-User-Id` was listed.

## Cause
`useGetRolesPermissionsQuery` in `ColumnPresetsSection.tsx` was not routing through `localMimirClient` when `!isPlatform`, so it queried the wrong config source locally.

## Fix
Added `useIsPlatform` and `useLocalMimirClient` hooks and passed `client: localMimirClient` when `!isPlatform` — the same pattern already used in `CustomCheckEditor/ConditionValue.tsx`.

Fixes #4279

### DIFF
--- a/dashboard/src/features/orgs/projects/database/dataGrid/components/EditPermissionsForm/sections/ColumnPresetsSection.tsx
+++ b/dashboard/src/features/orgs/projects/database/dataGrid/components/EditPermissionsForm/sections/ColumnPresetsSection.tsx
@@ -15,12 +15,15 @@ import {
 } from '@/components/ui/v3/select';
 import { useTableSchemaQuery } from '@/features/orgs/projects/database/common/hooks/useTableSchemaQuery';
 import type { RolePermissionEditorFormValues } from '@/features/orgs/projects/database/dataGrid/components/EditPermissionsForm/RolePermissionEditorForm';
+import { useIsPlatform } from '@/features/orgs/projects/common/hooks/useIsPlatform';
+import { useLocalMimirClient } from '@/features/orgs/projects/hooks/useLocalMimirClient';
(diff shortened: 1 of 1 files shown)


dbm03 (approved):
Thanks @HirenGajjar for the contribution! I verified this with the local dashboard: the column presets section now picks up permission variables correctly. 

I pushed a small lint cleanup on top so CI is green.

LGTM, merging shortly.

PR #4297, opened 2026-05-16
fix(dashboard): enable hover tooltip for storage data cells

### LINKED ISSUE #3677: fix: Datacell tooltip
https://github.com/nhost/nhost/blob/main/dashboard/src/features/orgs/projects/storage/dataGrid/components/DataGridCell/DataGridCell.tsx#L355-L370

Fix the tooltip for data cell. It does not seem to work.

singhvishalkr:
The tooltip for storage data cells wasn't showing on hover. Looking at the code, the Radix tooltip was in controlled mode (`open={tooltipOpen}`) but `tooltipOpen` only became true when opening an error message like "X is non-nullable".

Changed the `open` prop to `open={tooltipOpen || undefined}`. When `tooltipOpen` is false, passing `undefined` switches the tooltip to uncontrolled mode, enabling normal hover behavior. When `tooltipOpen` is true (error message showing), it stays in controlled mode.

Also updated the tooltip content to show the cell value preview on hover, falling back to 'Empty' for null values. Error messages still take priority when shown.

Closes #3677


dashboard/src/features/orgs/projects/storage/dataGrid/components/DataGridCell/DataGridCell.tsx
```diff
@@ -365,20 +365,28 @@ function DataGridCellContent<
         )}
     </div>
   );
-  // TODO: https://github.com/nhost/nhost/issues/3677
   if (isEditable) {
+    const cellValuePreview =
```
robertkasza: This PR will fundamentally change how the tooltip works, which we discussed with the team and we like how the tooltip would show the content of the cell on hover.

So, this tooltip was used to display a warning message if you were about to delete a value in a non-nullable column. Since we do not need that anymore we should clean that up too.

dashboard/src/features/orgs/projects/storage/dataGrid/components/DataGridCell/DataGridCell.tsx
```diff
@@ -365,20 +365,28 @@ function DataGridCellContent<
...
+        ? typeof optimisticValue === 'object'
+          ? JSON.stringify(optimisticValue)
+          : String(optimisticValue)
+        : null;
+
     return (
       <Tooltip
         delayDuration={100}
-        open={tooltipOpen}
+        open={tooltipOpen || undefined}
```
robertkasza: The v3 tooltip component will show the tooltip on hover if you do not pass the open prop

dashboard/src/features/orgs/projects/storage/dataGrid/components/DataGridCell/DataGridCell.tsx
```diff
@@ -365,20 +365,28 @@ function DataGridCellContent<
...
         onOpenChange={(newState) => {
```
robertkasza: we do not need this prop anymore.

dashboard/src/features/orgs/projects/storage/dataGrid/components/DataGridCell/DataGridCell.tsx
```diff
@@ -365,20 +365,28 @@ function DataGridCellContent<
...
           if (!newState) {
             resetTooltipTitle();
           }
         }}
       >
         <TooltipTrigger asChild>{content}</TooltipTrigger>
-        <TooltipContent>{tooltipTitle}</TooltipContent>
+        <TooltipContent>
```
robertkasza: could you position the tooltip to bottom center?
Something like this:

```
      <Tooltip delayDuration={100}>
        <TooltipTrigger asChild>{content}</TooltipTrigger>
        <TooltipContent side="bottom" align="center">
          Hello There
        </TooltipContent>
      </Tooltip>
```

dashboard/src/features/orgs/projects/storage/dataGrid/components/DataGridCell/DataGridCell.tsx
```diff
@@ -365,20 +365,28 @@ function DataGridCellContent<
         )}
     </div>
   );
-  // TODO: https://github.com/nhost/nhost/issues/3677
   if (isEditable) {
```
robertkasza: Since we are showing the content of the cell, we do not care about if the cell is editable or not.

robertkasza (commented):
Hi @singhvishalkr !
Thanks for the contribution! I had a few comments. Please take a look.

robertkasza (approved):

PR #4298, opened 2026-05-16
fix(storage): include file details in batch upload error responses

### LINKED ISSUE #3516: Batch upload error response should include specific file error details
### Issue summary

When uploading files in a batch to Hasura Storage, if one file fails, the API does return a `processedFiles` array for the successfully uploaded files, which is great. However, the `error` field in the response only contains a generic message (e.g., "an internal server error occurred") rather than something specific about which file failed and why (e.g., file name, reason like file too large, etc).

This makes it difficult for client applications to give users actionable feedback about which files failed and why.

### Example

**Actual response when a file fails due to size:**
```json
{
    "processedFiles": [
        { "id": "...", "name": "goodfile.pdf", ... },
        { "id": "...", "name": "anothergood.jpg", ... }
    ],
    "error": {
        "message": "an internal server error occurred"
    }
}
```

**Expected response:**
```json
{
    "processedFiles": [ ... ],
    "error": {
        "message": "File 'bigfile.pdf' is too large (10485760 bytes)",
        "data": {
            "file": "bigfile.pdf",
            "reason": "file too large"
        }
    }
}
```

### Steps to reproduce
1. Batch upload several files, including at least one file that exceeds the configured size limit.
2. Observe the response from the storage API.
3. The error message is generic and does not mention the failing file or the reason.

### Impact
- Client applications can't give users clear feedback about which file failed and why.
- Troubleshooting is harder for both users and developers.

### Suggested fix
- Update error handling in the batch upload endpoint to include the file name and the specific reason for failure in the `error` field of the response.
- Avoid wrapping all file errors in a generic internal server error message.

### Environment
- Version: (latest as of 2025-06-02)
- How deployed: Docker

### Additional context
Happy to provide logs or more details if needed. Thanks for the great project!

singhvishalkr:
﻿When a batch file upload fails due to individual file constraints (like exceeding size limits), the API response shows a generic "internal server error" instead of the specific error that occurred. This makes it difficult for callers to know why the upload failed and which file caused it.

The issue is in `processFile`: when `checkFileSize` returns a specific error like `FileTooBigError`, we were wrapping it in `InternalServerError` and losing all the useful details (filename, actual size, max size).

I fixed this by:
1. Returning the specific error from `checkFileSize` directly instead of wrapping it
2. Adding the failing file's name to the error data in the upload loop so callers know exactly which file failed

Now the response will include the proper error type, message, and file details instead of just "internal server error".

Fixes #3516


dbarrosop:
thanks, would you mind fixing the tests?

singhvishalkr:
@dbarrosop CI was failing `UploadFiles/unauthorized`: the upload loop called `SetData("file", ...)` on every `processFile` error, including Forbidden paths from metadata init. Those responses are supposed to keep `error.data` empty, matching the shape other handlers return.

Sizing errors still expose the offending file through `FileTooBigError` / `FileTooSmallError` data (`filename` plus limits). Virus detection still merges `file` through `scanAndReportVirus`.

I dropped the unconditional extra key and pushed https://github.com/singhvishalkr/nhost/commit/cb6c61a484ec4881d151582fedd65a080ec43f0f on this branch.

dbarrosop:
Thanks again, I think you should keep the previous commit and update the tests, adding a test that covers the `FileTooBigError` (or `FileTooSmallError`) would be ideal too to avoid regressions.

singhvishalkr:
@dbarrosop Restore the batch loop `file` annotation, align the unauthorized fixture on `error.data`, and add controller coverage for an oversized file in a multipart batch. Pushed https://github.com/singhvishalkr/nhost/commit/bef7dccc2f14828f2d3ac15bd8a04043ff286170

dbarrosop:
Thanks, pushed a couple of minor fixes but we should be good to merge

dbarrosop (approved):

PR #4343, opened 2026-05-24
feat(dashboard): add pagination to SQL editor results

### LINKED ISSUE #3716: dashboard: paginate large select queries in SQL editor
### Summary

The Raw SQL editor in the dashboard renders the full result set of a `SELECT` query directly into the DOM with no virtualization, or pagination. As a result, running a query that returns more than a few thousand rows freezes the main thread for several seconds and can effectively hang the dashboard.

### Impact

- A user running a `SELECT *` against a moderately sized table (~10k rows and up) will lock up their browser.

### Reproduction

1. Open the SQL Editor (`/orgs/<slug>/projects/<sub>/database/browser/<ds>/editor`).
2. Run any query that returns a large number of rows, e.g. `SELECT * FROM auth.users CROSS JOIN auth.users LIMIT 50000;` or a `generate_series` query.
For instance:
```sql
SELECT
  i AS id,
  'user' || i || '@example.com' AS email,
  'First' || i AS first_name,
  'Last' || i AS last_name,
  (now() - (i || ' seconds')::interval) AS created_at,
  repeat('lorem ipsum ', 8) AS description
FROM generate_series(1, 10000) AS s(i);
```
3. Observe the dashboard browser tab slow down and have higher memory usage.

### Root cause

All in `dashboard/src/features/orgs/projects/database/dataGrid/`:

- **`components/SQLEditor/SQLEditor.tsx:246-257`** — rows are rendered with a flat `rows.map(...)` over MUI `<TableRow>`/`<TableCell>`. No windowing.
- **`components/SQLEditor/SQLEditor.tsx:222`** — the table uses `tableLayout: 'auto'` combined with `whitespace-nowrap` on every cell (line 251), so the browser has to measure every cell to compute column widths. This compounds the cost beyond just "many DOM nodes".

### Proposed approach

`@tanstack/react-virtual` is already a dependency in the dashboard (used elsewhere), so adding it here doesn't introduce a new package.

1. Client-side virtualization of the result table: wrap the rendering in `useVirtualizer` from `@tanstack/react-virtual` (or migrate the result view to a virtualized data grid).
2. Or, add pagination to the result table, with a button that lets you go to the next/previous page.

### Acceptance criteria

- A `SELECT` returning ≥ 20k rows in the local dashboard renders without freezing the main thread for more than ~1s.
- Existing behavior for `CommandOk` results, errors, and small result sets is preserved.
- No regression to `Track this`, `Cascade metadata`, `Read only`, or migration-mode workflows.

### Files

- `dashboard/src/features/orgs/projects/database/dataGrid/components/SQLEditor/SQLEditor.tsx`

TaichKarna:
Fixes #3716

### Checklist

- [x] No breaking changes
- [x] Tests pass

PR Type : Enhancement

Large SELECT queries were rendering all rows into the DOM at once, freezing the browser tab.
Added client-side pagination to the SQL Editor result table. Users can now choose how many rows to show per page (25, 50, 100, 250, 500) and navigate between pages.


dbm03:
Hi @TaichKarna, thanks for your contribution! I've reviewed this and it works fine. I added some commits to fix the following:
- Running a query always lands on page 1 now, the page no longer persists across runs.
- Only show the pagination footer when there's a table result
- Biome linter check is fine now (if you run `pnpm lint` locally you can try it)

I'll be merging this soon, thanks again!

dbm03 (approved):

TaichKarna:
@dbm03 Thanks for the changes and the review!!

PR #4333, opened 2026-05-22
feat(dashboard): resolve sizing, color, and delay API gaps in v3/spinner

Dakshx07:
### Checklist

- [x] No breaking changes
- [x] Tests pass
- [x] New features have new tests (Not applicable; component-level upgrades verified via full existing dashboard tests)
- [x] Documentation is updated (if applicable)
- [x] Title of the PR is in the correct format

### Description

This PR resolves three critical API gaps in the modern Tailwind-based `Spinner` (v3) component to prepare it for a complete, regression-free migration from the legacy Material-UI-based `ActivityIndicator` (v2). 

1. **Sizing (`xs` preset)**: Added an `xs` preset (`size-3` / 12px) to match the fixed size of legacy inline loaders and prevent layout misalignment.
2. **Suppress Flicker (`delay` prop)**: Added support for an optional `delay` prop (`delay?: number` defaulting to 0) utilizing a lightweight React hook to defer spinner rendering and avoid flash animations on fast requests.
3. **Styling Defaults**: Keeps default light/dark stroke colors exactly as before to avoid visual breaks on existing spinner instances.

All existing dashboard unit tests, Typecheck compilers, and Biome lint validations pass cleanly with zero issues.

Addresses part of #4301

PR #4391, opened 2026-05-29
feat(dashboard): migrate onboarding pages to v3/spinner

Dakshx07:
This PR migrates legacy v2 `ActivityIndicator` loaders inside the onboarding pages to the modern `v3/spinner` component introduced in #4333.

### Changes
- Replaced `ActivityIndicator` imports and page-level loading screens in `onboarding/index.tsx` and `onboarding/project.tsx` with the new `Spinner`.
- Replaced button-level loaders during form submission with `<Spinner className="mr-2" size="xs" />`, utilizing the new `xs` size preset.
- Verified that all changes pass local TypeScript checks cleanly with zero errors.

### Checklist
- [x] No breaking changes
- [x] Tests pass

Hey @robertkasza, now that the base spinner changes from #4333 are merged, this is the follow-up PR migrating the onboarding flow loaders. Ready for your review!

### DIFF
--- a/dashboard/src/pages/onboarding/index.tsx
+++ b/dashboard/src/pages/onboarding/index.tsx
@@ -7,7 +7,7 @@ import { useForm } from 'react-hook-form';
 import { z } from 'zod';
 import { AuthenticatedLayout } from '@/components/layout/AuthenticatedLayout';
 import { Container } from '@/components/layout/Container';
-import { ActivityIndicator } from '@/components/ui/v2/ActivityIndicator';
+import { Spinner } from '@/components/ui/v3/spinner';
 import { Alert } from '@/components/ui/v2/Alert';
 import { Box } from '@/components/ui/v2/Box';
 import { Text } from '@/components/ui/v2/Text';
@@ -166,7 +166,7 @@ export default function OnboardingPage() {
   if (loadingOrgs || loadingPlans || loadingInvites) {
     return (
       <div className="flex h-screen items-center justify-center">
-        <ActivityIndicator />
+        <Spinner />
       </div>
     );
   }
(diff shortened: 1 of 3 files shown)


dashboard/src/pages/onboarding/index.tsx
```diff
@@ -418,7 +418,7 @@ export default function OnboardingPage() {
                 >
                   {form.formState.isSubmitting ? (
                     <>
-                      <ActivityIndicator className="mr-2 h-4 w-4" />
+                      <Spinner className="mr-2" size="xs" />
```
robertkasza: What do you think about using the ButtonWithLoading from [the v3 button component](https://github.com/nhost/nhost/blob/main/dashboard/src/components/ui/v3/button.tsx) instead?

dashboard/src/pages/onboarding/project.tsx
```diff
@@ -310,7 +310,7 @@ export default function OnboardingProjectPage() {
                   >
                     {form.formState.isSubmitting ? (
                       <>
-                        <ActivityIndicator className="mr-2 h-4 w-4" />
+                        <Spinner className="mr-2" size="xs" />
```
robertkasza: I think we could use the ButtonWithLoading from [the v3 button component](https://github.com/nhost/nhost/blob/main/dashboard/src/components/ui/v3/button.tsx) too, what do you think?

Dakshx07: Absolutely agree! That is much cleaner and fully aligns with our v3 components structure. 

I've updated the PR to use `ButtonWithLoading` on both onboarding pages, which simplifies the submit button markup and removes the manual spinner logic. Thanks for the catch!

robertkasza (approved):

PR #4402, opened 2026-05-31
fix(docs): solve overflow issue on chromium due to hero

TaichKarna:
### Checklist

- [x] No breaking changes
- [x] Tests pass

screenshot-1780247842

On chromium the 100vw is calculated from the viewport width, which historically includes the area occupied by classic scrollbars. When a vertical scrollbar is present, 100vw can therefore be slightly wider than 100%, producing horizontal overflow just like here, this happened for welcome, product overview and platform pages.


dbarrosop:
thanks for the contribution!

dbarrosop (approved):

TaichKarna:
@dbarrosop Thanks for the review

PR #4431, opened 2026-06-02
fix(auth): prevent JSON injection in turnstile siteverify request

skoveit:
### **What this PR solves**
  The Cloudflare Turnstile verification request was assembled by string-concatenating the attacker-controlled `x-cf-turnstile-response`
  header into a JSON body. A crafted header could close the `response` string and append a second `secret`/`response` pair; Cloudflare's
  last-wins parsing then used the attacker's own sitekey, returned `success:true`, and bypassed the abuse gate on signup, passwordless,
  OTP, and password-reset endpoints. This builds the request with `json.Marshal`, so the header value is escaped and can no longer override
  the configured secret.

  ___

  ### **PR Type**
  Bug fix

  ___
  - Eliminates the JSON-injection (CWE-116) that smuggled a duplicate `secret` key and defeated the Turnstile abuse gate.
  - No API or contract change: valid tokens produce a semantically identical request; only malicious payloads are now correctly rejected
  (verified live against Cloudflare siteverify with their test secrets).

  ___

  ### Diagram Walkthrough

  ```mermaid
  flowchart LR
    A["Request with<br/>x-cf-turnstile-response header"] --> B["Turnstile middleware"]
    B --> C["makeTurnstileRequest<br/>json.Marshal(secret, response)"]
    C --> D["Cloudflare siteverify"]
    D --> E{"success?"}
    E -->|true| F["Protected handler"]
    E -->|false| G["403 Forbidden"]

```

### DIFF
--- a/services/auth/go/middleware/turnstile.go
+++ b/services/auth/go/middleware/turnstile.go
@@ -32,11 +32,19 @@ func makeTurnstileRequest(
 	secret string,
 	tokenResponse string,
 ) (*TurnstileResponse, error) {
+	body, err := json.Marshal(map[string]string{
+		"secret":   secret,
+		"response": tokenResponse,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal turnstile request: %w", err)
+	}
+
 	request, err := http.NewRequestWithContext(
 		ctx,
 		http.MethodPost,
 		tunrstileURL,
-		bytes.NewBufferString(`{"secret": "`+secret+`","response":"`+tokenResponse+`"}`),
+		bytes.NewReader(body),
 	)
 	if err != nil {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: services/auth/go/middleware/turnstile.go [hunk 1]
```diff
@@ -32,11 +32,19 @@ func makeTurnstileRequest(
 	secret string,
 	tokenResponse string,
 ) (*TurnstileResponse, error) {
+	body, err := json.Marshal(map[string]string{
```
dbarrosop, line 35: can we add tests to avoid regressions (and to demonstrate the fix)?

skoveit: done, added `TestMakeTurnstileRequestBuildsSafeBody`, it fails against the old concatenated body

dbarrosop (approved):

PR #4432, opened 2026-06-02
fix(auth): validate PKCE verifier before consuming authorization code

skoveit:
### **What this PR solves**
  The PKCE token-exchange endpoint deleted the authorization code (a `DELETE … RETURNING`) *before* validating the code verifier, so a wrong/forged verifier still burned the code — letting an attacker who intercepts a `?code=` DoS the legitimate user's flow. This reorders the logic to validate first and only consume the code after the verifier passes.

  ___

  ### **PR Type**
  Bug fix

  ___

  ### **Description**
  - Replaced the destructive `ConsumePKCEAuthorizationCode` (`DELETE … RETURNING`) with a non-destructive `GetPKCEAuthorizationCode` SELECT
  plus a separate `DeletePKCEAuthorizationCode` DELETE.
  - Reordered `TokenExchange` to SELECT → validate verifier → DELETE, so the code is consumed only after successful PKCE validation.
  - Added a regression assertion: the "wrong code verifier" test sets up the SELECT but no DELETE, so gomock fails if the code is consumed
  on validation failure.
  - Regenerated sqlc query code and the DBClient mock to match the new interface.

  ___

  ### Diagram Walkthrough

  ```mermaid
  flowchart LR
    A["POST /v1/token/exchange"] --> B["GetPKCEAuthorizationCode (SELECT)"]
    B --> C["ValidateS256 verifier"]
    C -->|valid| D["DeletePKCEAuthorizationCode (DELETE)"]
    C -->|invalid| E["400 — code preserved"]
    D --> F["Issue session"]
    
    ```

### DIFF
--- a/services/auth/go/sql/query.sql
+++ b/services/auth/go/sql/query.sql
@@ -558,7 +558,7 @@ RETURNING *;
 
 -- name: ConsumePKCEAuthorizationCode :one
 DELETE FROM auth.pkce_authorization_codes
-WHERE code_hash = $1 AND expires_at > now()
+WHERE code_hash = $1 AND code_challenge = $2 AND expires_at > now()
 RETURNING *;
 
 -- name: DeleteExpiredPKCEAuthorizationCodes :exec
--- a/services/auth/go/controller/controller.go
+++ b/services/auth/go/controller/controller.go
@@ -135,7 +135,7 @@ type DBClient interface { //nolint:interfacebloat
 		ctx context.Context, arg sql.InsertPKCEAuthorizationCodeParams,
 	) (sql.AuthPkceAuthorizationCode, error)
 	ConsumePKCEAuthorizationCode(
-		ctx context.Context, codeHash string,
+		ctx context.Context, arg sql.ConsumePKCEAuthorizationCodeParams,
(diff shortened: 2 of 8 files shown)


services/auth/go/sql/query.sql
```diff
@@ -556,10 +556,13 @@ INSERT INTO auth.pkce_authorization_codes (
 ) VALUES ($1, $2, $3, $4, $5)
 RETURNING *;
 
--- name: ConsumePKCEAuthorizationCode :one
```
dbarrosop, line 559: If the issue is "wrong/forged verifier still burned the code", can we instead do:

```
-- name: ConsumePKCEAuthorizationCode :one
DELETE FROM auth.pkce_authorization_codes
WHERE code_hash = $1 AND code_challenge = $2 AND expires_at > now()
RETURNING *;
```

Otherwise we have a race condition with the changes here as it is possible to have two requests run `GetPKCEAuthorizationCode`, pass the check and then both call `DeletePKCEAuthorizationCode`.

skoveit: definitely cleaner, going with your approach instead

dbarrosop (approved):

PR #4445, opened 2026-06-03
fix(storage): bound transform dimensions and blur

skoveit:
### Checklist

- [x] No breaking changes
- [x] Tests pass
- [x] Documentation is updated (`--help`)

Storage now enforces configurable maximum image transform dimensions and blur sigma before sending transform parameters into libvips. This prevents oversized width, height, or blur requests from exhausting storage-service resources while documenting the limits for API and SDK consumers.

___

### **Change Type**
Bug fix

___

### **Description**
- Adds storage service configuration for maximum transform dimension and maximum blur sigma, then passes those limits into image transformation paths.
- Rejects over-limit `w`, `h`, and `b` query parameters for file download, HEAD metadata, and presigned URL content flows.
- Updates OpenAPI/schema, generated SDK docs, and storage image transformation docs to describe the default limits and self-hosted configuration knobs.
- Expands controller and image package tests to cover oversized dimensions, derived dimensions, blur limits, and custom limit settings.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["Storage config/env"] --> B["controller.NewFile"]
  B --> C["image.NewTransformer(max dimension, max blur)"]
  C --> D["GET /files + HEAD metadata"]
  C --> E["presigned URL contents"]
  D --> F["validate transform query params"]
  E --> F
  F --> G["libvips transform"]
```

<tr><td><strong>Storage service</strong></td><td><details><summary>8 files</summary><table>
<tr><td><strong>services/storage/cmd/serve.go</strong><dd><code>Adds image transformer max-dimension and max-blur config flags/env wiring.</code></dd></td><td>+22/-1</td></tr>
<tr><td><strong>services/storage/controller/get_file.go</strong><dd><code>Propagates transform limit errors through file download responses.</code></dd></td><td>+26/-1</td></tr>
<tr><td><strong>services/storage/controller/get_file_metadata_headers.go</strong><dd><code>Returns transform metadata header errors when HEAD transformations exceed limits.</code></dd></td><td>+6/-1</td></tr>
<tr><td><strong>services/storage/controller/get_file_metadata_headers_test.go</strong><dd><code>Updates test transformer construction for the new limit-aware signature.</code></dd></td><td>+1/-1</td></tr>
<tr><td><strong>services/storage/controller/get_file_test.go</strong><dd><code>Adds controller regression tests for default and custom transform limit failures.</code></dd></td><td>+193/-18</td></tr>
<tr><td><strong>services/storage/controller/get_file_with_presigned_url.go</strong><dd><code>Propagates transform errors from presigned URL content requests.</code></dd></td><td>+4/-1</td></tr>
<tr><td><strong>services/storage/image/image.go</strong><dd><code>Adds limit-aware transformer validation for dimensions, derived dimensions, and blur sigma.</code></dd></td><td>+144/-21</td></tr>
<tr><td><strong>services/storage/image/image_test.go</strong><dd><code>Adds image transformer tests for default/custom limits and limit validation.</code></dd></td><td>+156/-2</td></tr>
</table></details></td></tr>
<tr><td><strong>API and clients</strong></td><td><details><summary>3 files</summary><table>
<tr><td><strong>services/storage/controller/openapi.yaml</strong><dd><code>Documents maximums for transform width, height, and blur query parameters.</code></dd></td><td>+9/-9</td></tr>
<tr><td><strong>docs/src/schemas/storage.yaml</strong><dd><code>Mirrors storage schema transform limit descriptions for docs consumers.</code></dd></td><td>+9/-9</td></tr>
<tr><td><strong>packages/nhost-js/src/storage/client.ts</strong><dd><code>Refreshes generated storage client parameter comments with limit descriptions.</code></dd></td><td>+12/-12</td></tr>
</table></details></td></tr>
<tr><td><strong>Documentation</strong></td><td><details><summary>2 files</summary><table>
<tr><td><strong>docs/src/content/docs/products/storage/image-transformation.mdx</strong><dd><code>Updates image transformation ranges and explains configurable server-side bounds.</code></dd></td><td>+8/-4</td></tr>
<tr><td><strong>docs/src/content/docs/reference/javascript/nhost-js/storage.md</strong><dd><code>Refreshes SDK storage transform parameter docs with default limit behavior.</code></dd></td><td>+6/-6</td></tr>
</table></details></td></tr>
</tbody></table>

</details>

___


services/storage/image/image.go
```diff
@@ -23,6 +23,17 @@ const (
 	ImageTypeHEIC
 )
 
+const (
+	// maxImageDimension bounds the width and height an image may be resized
+	// to. libvips allocates the full output buffer up front, so an unbounded
+	// dimension lets a single (potentially unauthenticated) request exhaust
+	// process memory.
+	maxImageDimension = 8000
```
dbarrosop: what is the memory impact of these values? what is the worst case scenario? how did we reach to this number? would be nice to understand the issue first better. We already set some workers based on the number of CPUs precisely to bound cpu/memory usage so understanding how everything ties together would be nice and help us decide whether this number should be dynamic or configurable or if some large number suffices (I suspect 8000 may be too small but maybe 50000 is perfectly fine)

skoveit: the numbers come from the [security report](https://github.com/nhost/projects/blob/security-review/security_review/storage/STORAGE-01-image-transform-dos.md) which has empirical measurements ([P2](https://github.com/nhost/projects/issues/180))

50000x50000 was confirmed to consume ~7.5 GB on a live instance (OOM-kills instantly). 8000 was chosen conservatively from there.

memory scales as width x height x ~4 bytes, so 8000x8000 is ~256 MB per request. 

Note that the cap only applies to the resize target, not the original file, images larger than 8000px can still be served at their original size with no transform params. to address the limitation of a hardcoded value, I'll make it configurable following the same pattern as `flagImageTransformerWorkers`

dbarrosop:
A few comments:

1. We need to allow configuring this in the backend (I can help)
2. We need to update the openapi spec
3. We need to update the docs (https://docs.nhost.io/products/storage/image-transformation/#_top)
4. Maybe add a few regression tests that trigger the DoS protection

Other than that LGTM so we can merge once we have the other stuff ready.

docs/src/content/docs/products/storage/image-transformation.mdx
```diff
@@ -15,16 +15,20 @@ Images can be transformed on-the-fly by adding query parameters to file URLs. Tr
...
 | `f` | Output format | `auto`, `same`, `jpeg`, `webp`, `png`, `avif` |
-| `b` | Blur sigma | `0` or higher |
+| `b` | Blur sigma | `0` -- `250` |
 
 **Format options:**
 - `auto` -- uses content negotiation based on the `Accept` header (serves WebP/AVIF to browsers that support it)
 - `same` -- keeps the original format (default)
 
+:::note
+`w`, `h`, and `b` are bounded to keep a single transformation from exhausting server memory. Values above the maximum are clamped rather than rejected. The values shown are the defaults; self-hosted instances can tune them with the `IMAGE_TRANSFORMER_MAX_DIMENSION` and `IMAGE_TRANSFORMER_MAX_BLUR_SIGMA` environment variables on the storage service.
```
dbarrosop: I don't think this is true anymore, the openapi file says it is rejected

services/storage/controller/get_file_test.go
```diff
@@ -297,3 +297,76 @@ func TestGetFile(t *testing.T) { //nolint:maintidx
 		})
 	}
 }
+
+func TestGetFileRejectsOversizedImageManipulation(t *testing.T) {
```
dbarrosop: is this test needed or can the test cases be added to the test above?

services/storage/image/image.go
```diff
@@ -107,13 +133,36 @@ func NewTransformer(maxWorkers int) *Transformer {
...
+		maxBlurSigma: maxBlurSigma,
 	}
 }
 
 func (t *Transformer) Shutdown() {
 	vips.Shutdown()
 }
 
+// MaxDimension reports the configured upper bound for output width and height.
+// A nil Transformer reports the package default so callers can validate request
```
dbarrosop: this is actually inaccurate and a bit misleading. Returning the package default isn't good to validate request parameters against the limit. If I configure the server to have a t.maxDimension of 1000 but you compare against the default 8000 what you are comparing is incorrect. I think we should return either 0 or just remove the nil guard as I am not sure we should support a nil Transformer

services/storage/controller/get_file.go
```diff
@@ -116,10 +117,54 @@ type ImageManipulationOptionsGetter interface {
...
 }
 
+// checkImageManipulationLimits rejects a request whose explicitly requested
+// width, height or blur exceeds the server-configured maximum. We return a 400
+// here rather than silently clamping so the caller learns the limit instead of
+// receiving an image at an unexpected size. The maxima are operator-configurable
+// (IMAGE_TRANSFORMER_MAX_DIMENSION / IMAGE_TRANSFORMER_MAX_BLUR_SIGMA), so the
+// bound is enforced against the live values rather than as a static OpenAPI
+// constraint, which could not track the configured value.
+func checkImageManipulationLimits(
```
dbarrosop: this is a bit confusing because we are doing this check here but then on [image.go](https://github.com/nhost/nhost/pull/4445/changes#diff-c24b8cd55fb444467526ed20802d32fa874ec1561e8487e95c8e776af5529f4cR209-R210). I think we should do two things:

1. Decide whether we want to clamp down the value or return an error
2. Do it on the image transformer or leave the image transformer dumb and do it  here, either would be fine but doing it in both places seems overkill and will lead to errors like the one we have now

services/storage/controller/get_file_test.go
skoveit: Folded in, standalone test removed

services/storage/controller/get_file.go
skoveit: all limit logic is now in the image transformer, controller just delegates and maps the error to a 400. The two checks (explicit pre-download, derived post-load) cover values that become known at different times so neither can be dropped, but no value is checked twice

services/storage/controller/get_file_test.go
```diff
@@ -22,6 +24,8 @@ func TestGetFile(t *testing.T) { //nolint:maintidx
 		name     string
 		request  api.GetFileRequestObject
 		expected api.GetFileResponseObject
+		wantCode int
```
dbarrosop: I don't think we need wantCode or wantMsg. Maybe we need to change expected to be any so we can then set it to either api.GetFileResponseObject or controller.APIError but if I am not mistaken cmp.Diff should handle this gracefully

dbarrosop: Correction, keeping wantCode is fine, but I think wantMsg shouldn't be needed and instead we should use expected. I would not rely on empty wantCode, I would set it to the correct status code we always expect (200, 201, etc)

services/storage/controller/get_file_test.go
```diff
@@ -250,15 +286,15 @@ func TestGetFile(t *testing.T) { //nolint:maintidx
 					ExtraHeaders:  make(http.Header),
 				},
 				nil,
-			)
+			).Times(downloadCalls)
```
dbarrosop: if we go with `expected any` you can gate GetFile based on the type assertion. For instance:

```go
if _, ok := tc.expected.(controller.APIError); ok {
			contentStorage.EXPECT().GetFile(
				gomock.Any(),
				"55af1e60-0f28-454e-885e-ea6aab2bb288",
				gomock.Any(),
			).Return(
				&controller.File{
					StatusCode:    200,
					Etag:          `"55af1e60-0f28-454e-885e-ea6aab2bb288"`,
					Body:          io.NopCloser(strings.NewReader("Hello, world!")),
					ContentLength: 64,
					ExtraHeaders:  make(http.Header),
				},
				nil,
			)
}
```

services/storage/controller/get_file_test.go
```diff
@@ -297,3 +339,111 @@ func TestGetFile(t *testing.T) { //nolint:maintidx
...
+	}
+}
+
+// TestGetFileRejectsOversizedDerivedDimension exercises the limit a request can
+// only hit once the source image is loaded: nhost.jpg is 678x258 (landscape),
+// so asking for only the height at the cap derives a width ~2.6x larger. The
+// controller validates the explicit height fine (it is at the cap), so the
+// oversized derived width is caught by the transformer, which must surface it
+// as a 400 rather than a 500.
+func TestGetFileRejectsOversizedDerivedDimension(t *testing.T) {
```
dbarrosop, line 387: any reason why this couldn't be folded into TestGetFile

services/storage/controller/get_file_test.go
```diff
@@ -297,3 +339,111 @@ func TestGetFile(t *testing.T) { //nolint:maintidx
 		})
 	}
 }
+
+func assertAPIError(t *testing.T, resp any, wantCode int, wantMsg string) {
```
dbarrosop: let's use cmp.Diff instead

services/storage/controller/get_file_test.go
dbarrosop: ignore this one

services/storage/controller/get_file_test.go
dbarrosop: ok, look a bit more into this. I would suggest the following:

1. Refactor TestGetFile table to avoid wantCode == 0
     - Use one expected any field.
     - Normal cases keep api.GetFile...Response.
     - Error cases use a small test-only struct, not raw *controller.APIError.

 Example shape:

 ```go
   type expectedAPIError struct {
      StatusCode int
      Message   string
   }
 ```

 Then cases become:

 ```go
   {
      name: "width over max is rejected",
      request: api.GetFileRequestObject{
         Id:     fileID,
         Params: api.GetFileParams{W: new(9000)},
      },
      expected: expectedAPIError{
         StatusCode: http.StatusBadRequest,
         Message:   "image manipulation parameters out of range: width 9000 exceeds the maximum of 8000",
      },
      wantDownload: false,
   }
 ```

 2. Add explicit wantDownload bool
     - Don’t infer download behavior from error-ness.
     - Explicit option errors should not download.
     - Derived dimension errors must download first.

 3. Use cmp.Diff, but against public values
     - Don’t compare *controller.APIError directly because its fields are private.
     - Convert it to expectedAPIError:

 ```go
   got := expectedAPIError{
      StatusCode: apiErr.StatusCode(),
      Message:   apiErr.PublicMessage(),
   }

dbarrosop (approved):

PR #4503, opened 2026-06-10
fix(storage): add security response headers to file responses

skoveit:
thanks @hunt-with-4bh1

### Checklist

- [x] No breaking changes
- [x] Tests pass

### **What this PR solves**
Storage serves user-uploaded bytes with `Content-Disposition: inline` and the MIME type detected at upload, but set no browser security headers on those responses — leaving stored HTML/SVG as a potential XSS vector. This adds a `securityheaders` gin middleware (mirroring the existing `cdncachecontrol` writer pattern) that sets `X-Content-Type-Options: nosniff` on every response and a locked-down `Content-Security-Policy: default-src 'none'; sandbox` on active-content types only. It is defense-in-depth against stored XSS and content sniffing, covers the unauthenticated presigned-URL path, and is non-breaking: non-active files (PDF, images) still render and embed.

### **PR Type**
Bug fix

___

### **Description**
- Add a `securityheaders` gin middleware: `X-Content-Type-Options: nosniff` on every response, and `Content-Security-Policy: default-src 'none'; sandbox` only when the response `Content-Type` is active content.
- Scope active content by rule, not by a hardcoded list: `text/html`, `application/xml`, `text/xml`, and — per RFC 6839 — any `+xml` type (SVG, XHTML, RSS, Atom, …). This closes the gap where detector-emitted feed/vocabulary types (`application/rss+xml`, etc.) would otherwise slip past an enumerated list.
- Fail closed on a malformed, non-empty `Content-Type`: if `mime.ParseMediaType` errors, apply the sandbox CSP rather than risk serving active content uncovered (a browser may still read `text/html` from a malformed header). Empty `Content-Type` is left alone — nothing to execute, and `nosniff` still applies.
- Decide the CSP at write time via a deferred `ResponseWriter` wrapper, because the handler sets the `Content-Type` the policy scopes on while it runs.
- Register the middleware unconditionally in `configureMiddleware`, first in the chain, so the writer wraps before any handler writes — covering direct download, range requests, `HEAD` metadata, and the presigned-URL contents endpoint.
- Add a table-driven middleware unit test (text/html/svg/xml/rss/gpx/json/pdf/html-error/malformed-html) and a `client` integration test asserting headers on real file responses, including the unauthenticated presigned-URL path and the image-is-not-sandboxed case.
- Extend `IgnoreResponseHeaders()` in the `client` golden tests to ignore the two cross-cutting headers, consistent with how `Date`/`Surrogate-Key`/`Last-Modified`/`X-B3-*` are handled.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["File request<br/>(download / presigned)"] --> B["securityheaders middleware"]
  B --> C["Set nosniff up-front; wrap writer"]
  C --> D["Handler sets Content-Type + writes bytes"]
  D --> E{"Parse Content-Type"}
  E -->|"parse error, non-empty"| F["Sandbox (fail closed)"]
  E -->|"html / xml / +xml"| F
  E -->|"pdf / image / other / empty"| G["No CSP (still renders/embeds)"]
```

<tr><td><strong>Security headers middleware</strong></td><td><details><summary>2 files</summary><table>
<tr>
  <td><strong>securityheaders.go</strong><dd><code>New middleware: nosniff always; CSP scoped to active content via +xml rule; fail-closed on bad type</code></dd></td>
  <td><a href="https://github.com/nhost/nhost/pull/4503/files#diff-4094b28a640a92546287a2a4c0582c95fd6eb7983673b64e359ed5928249c781">+150/-0</a></td>
</tr>
<tr>
  <td><strong>serve.go</strong><dd><code>Register the middleware unconditionally, first in the chain</code></dd></td>
  <td><a href="https://github.com/nhost/nhost/pull/4503/files#diff-574dc29b7b8f295e128c48091a8de992243bc59c1fdd080a71bb7edcb102dbaa">+4/-0</a></td>
</tr>
</table></details></td></tr>
<tr><td><strong>Tests</strong></td><td><details><summary>3 files</summary><table>
<tr>
  <td><strong>get_file_security_headers_test.go</strong><dd><code>Integration test on real file responses incl. presigned URL + image-not-sandboxed</code></dd></td>
  <td><a href="https://github.com/nhost/nhost/pull/4503/files#diff-5e7d3a5550727f4a73ed1c379a7d4ebd514ad1c8b44f751f0517ed4da3cdd90a">+196/-0</a></td>
</tr>
<tr>
  <td><strong>securityheaders_test.go</strong><dd><code>Table-driven unit test: nosniff always, CSP on active content + +xml + fail-closed</code></dd></td>
  <td><a href="https://github.com/nhost/nhost/pull/4503/files#diff-5943294277b36b17c900797f431349ca35ee7cb627a8ec108972a7f0c5876887">+115/-0</a></td>
</tr>
<tr>
  <td><strong>client_test.go</strong><dd><code>Ignore the cross-cutting headers in golden comparisons</code></dd></td>
  <td><a href="https://github.com/nhost/nhost/pull/4503/files#diff-6faf9c9c1b7afdb262d0e931ef175258402e2dc3893638bd362011c4d11ec956">+6/-0</a></td>
</tr>
</table></details></td></tr>
</tbody></table>

</details>

___


services/storage/middleware/securityheaders/securityheaders.go
```diff
@@ -0,0 +1,43 @@
...
+	valueNoSniff = "nosniff"
+
+	// valueCSP constrains how the browser may interpret a served response: the
+	// sandbox directive (without allow-scripts) prevents active content from
+	// running, default-src 'none' blocks every subresource and external load,
+	// and frame-ancestors 'none' blocks framing. Static files still render.
+	valueCSP = "default-src 'none'; sandbox; frame-ancestors 'none'"
+
+	// valueXFrameOptions backs up frame-ancestors for legacy browsers.
+	valueXFrameOptions = "DENY"
```
meh: Wouldn't the frame deny / `frame-ancestors 'none'` possibly break some frontend uses and embedding in third-party websites uses?

services/storage/middleware/securityheaders/securityheaders.go
```diff
@@ -0,0 +1,150 @@
...
+// HTML is the only non-XML executable document type. The whole XML family is
+// covered: text/xml and application/xml, plus — per RFC 6839 — any type with a
+// "+xml" structured-syntax suffix (image/svg+xml, application/xhtml+xml,
+// application/rss+xml, application/atom+xml, ...). A browser parses those as XML
+// and honours an <?xml-stylesheet?> PI, so an XSLT can emit script-bearing HTML
+// in this origin; matching the suffix avoids enumerating (and missing) each
+// vocabulary.
+func isActiveContentType(mediaType string) bool {
+	switch mediaType {
+	case "text/html", "application/xml", "text/xml":
```
dbarrosop, line 69: I think this may fail if a contenttype specified something like the charset `text/html; charset=unicode;`. The stdlib in go has a method to compare content types IIRC

dbarrosop: Nevermind, I should have scrolled a bit further down

dbarrosop (approved):

PR #4531, opened 2026-06-13
fix(functions): replace ctx.watch() with mtime polling for Docker bind mounts

### LINKED ISSUE #4233: local dev not hot reloading on file changes in /functions
copied from discord conversation

> **Note:** Bug reports that are clearly AI-generated will not be accepted and will be closed immediately. Please write your bug report in your own words.

**Describe the bug**
usually when I changed any file in /functions folder nhost automatically picked up the change and redeployed the functions automatically, I was able to see this in the logs. This is not working anymore since I updated the CLI, nhost version shows 1.45.0. 
I had quite a few package updates but I doubt that those are the problem. Is this a know issue? 
Right now I have to nhost down && nhost up to see the changes I made, which takes around 4 minutes, so thats not usable

**Additional context**
reply from claude

```
Root Cause

 Two sequential CLI upgrades introduced the regression:

 v1.44.0 (April 20, 2026): Updated the functions Docker image from v1.5.0 → v2.0.0-rc2. The v2 image is a
  major rewrite (moved from Google Cloud Run to AWS Lambda architecture, with per-request esbuild
 rebuilds of ~200ms). The v2 image still officially documents "hot reload of package.json, lock files,
 and the functions directory."

 v1.45.0 (April 30, 2026) — PR #4211 "Reuse esbuild context to lower memory usage": Refactored the
 CLI-side esbuild to use a single shared context across all functions instead of one context per
 function. This introduced a "rebuildable watch context" with new lifecycle methods
 (prepareFunctionEntry, disposeFunctionEntry, rebuildContext). The shared-context refactor appears to
 have broken the file-change detection that previously triggered automatic rebuilds.

 Current state: .nhost/docker-compose.yaml uses image nhost/functions:22-2.1.0 with the project root
 bind-mounted at /opt/project, so the files ARE reaching the container — the issue is in the
 watch/rebuild triggering logic.
```

lakshitsoni26:
## What

Resolves #4233 — Functions fail to hot-reload on content changes during local development.

## Why

Currently, `esbuild`'s `ctx.watch()` relies on `inotify` (Linux) or `FSEvents` (macOS) to detect filesystem events. These events do not propagate reliably through Docker bind-mount volumes, which is the exact mechanism `nhost up` uses to mount the local workspace into the functions container. 

The server already had a 1-second polling loop in place to handle structural changes (new or deleted files) for this exact reason, but content changes to existing files were still falling back to the broken `ctx.watch()` implementation.

## How

- **Removed `ctx.watch()` entirely:** It provides no value inside the Docker volume constraints.
- **Unified Change Detection:** Extended the existing 1-second `setInterval` loop to track and compare `mtimeMs` for all known source files.
- **Zero-Overhead Rebuilds:** When an mtime diff is detected, the poller triggers `scheduleIncrementalRebuild()`. This explicitly re-uses the existing esbuild `buildContext`, avoiding the heavy memory/CPU overhead of disposing and recreating the AST/context.
- **Safe Baselines:** Context recreates (which happen on entry-set changes) now safely re-snapshot all `mtimes` to ensure the next poll starts with a clean baseline.

## Testing

*(Testing via local Node.js environment replacing Docker volume injection)*

**Before:**
Saving an existing function file resulted in no build logs, and subsequent requests returned stale responses.

**After:**
Function changes are detected within ~1s and correctly trigger an incremental esbuild reload.

Screenshot 2026-06-13 at 1 43 16 PM
Screenshot 2026-06-13 at 1 43 42 PM

### Checklist

- [x] No breaking changes
- [x] Tests pass

### DIFF
--- a/services/functions/server.js
+++ b/services/functions/server.js
@@ -67,6 +67,11 @@ const functionEntries = new Map();
 // Track discovered functions for metadata endpoint
 const functionMeta = new Map();
 
+// Track source file mtimes for polling-based change detection.
+// inotify/FSEvents do not fire reliably over Docker bind-mount volumes,
+// so ctx.watch() silently misses edits to existing function files.
+const functionMtimes = new Map();
+
 // Single esbuild context covering every function's wrapper as an entry point.
 // Per-function contexts retained the parsed dep graph (express, aws-sdk, ...)
 // once per function, multiplying memory by N. A single context shares one
@@ -83,6 +88,19 @@ function scheduleRebuild(functionsPath) {
   rebuildInFlight = rebuildInFlight
     .catch(() => {})
     .then(() => rebuildContext(functionsPath));
+  return rebuildInFlight;
+}
+
+// Incremental rebuild reuses the existing context (no dispose/recreate), used
+// when only file contents changed (not the entry-point set).
+function scheduleIncrementalRebuild() {
+  rebuildInFlight = rebuildInFlight
+    .catch(() => {})
+    .then(async () => {
+      if (buildContext) {
+        await buildContext.rebuild();
+      }
+    });
   return rebuildInFlight;
 }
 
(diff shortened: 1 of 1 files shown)


dbarrosop:
Thanks for the PR, unfortunately this issue is already addressed in #4231

PR #4537, opened 2026-06-15
feat(examples): add OAuth account linking to react demo

skoveit:
Adds an OAuth account-linking demo to the react example.

- New "Connected Accounts" panel on the Profile page to link a provider (GitHub) to the current account and show its linked state
- New `connect/callback` page that renders the link success/error result
- Grant the `user` role self-scoped select on `auth_user_providers` so the demo can read which providers are linked
Screenshot_20260614_155047


PVUL (approved):
The code itself looks fine. I think if this is a noteworthy change, it should be added to the README.md of the react-demo directory, so other developers can use it as a reference.

PR #4565, opened 2026-06-18
fix(dashboard): prevent stale 1:2-ratio validation error when switching compute resource presets

### LINKED ISSUE #4562: dashboard: stale 1:2 ratio validation error after switching presets in compute resources
### Summary
In the Compute Resources settings form, switching presets in a certain order leaves a per-service validation error: "vCPU and Memory for this service must follow a 1:2 ratio when more than one replica is selected or when the autoscaler is activated.", on services that are actually valid (1 replica, autoscaler off).

### Steps to reproduce
1. Open **Settings → Compute Resources** (with the feature enabled).
2. On the **Overview** tab, select the **Performance** preset.
3. Then select the **Standard** preset.
4. Open the **Advanced** tab.

➡️ The **Auth** and **Storage** sections show the red "...must follow a 1:2 ratio..." error,
even though both show **1 replica** and **Autoscaler off**.

### Expected
No validation error: with 1 replica and the autoscaler off, the 1:2 ratio rule does not apply.

### Actual
A stale ratio error is displayed on every service whose target preset ratio isn't 1:2 (Auth/Storage for Standard; Hasura instance is unaffected because its Standard ratio is already 1:2). 
The error is misleading only, Save is not blocked.

### Root cause
`applyPresetToForm` (`dashboard/.../resources/settings/utils/presets/useApplyPreset.ts`)
applies each field with `{ shouldValidate: true }` in the order `vcpu → memory → replicas → autoscale → …`. When moving from an autoscaling preset to one whose ratio isn't 1:2, the `memory` field is validated while `autoscale` is still `true` (it's flipped to `false` two writes later), so the `is-matching-ratio` test in `resourceSettingsValidationSchema.ts` fails and writes a field-scoped error to `{service}.memory`. 
The subsequent `autoscale = false` write only re-validates the `autoscale` field, so the stale `memory` error is never cleared, and the trailing `trigger()` does not reliably clear it (it clears only when the per-service fields are mounted, e.g. the Advanced tab is already open). 
The leftover error surfaces when the Advanced tab mounts the affected `ServiceRow`.

### Suggested fix
In `applyPresetToForm`, avoid validating transient intermediate states — apply all fields with `shouldValidate: false` and call a single `trigger()` once the values are consistent (and/or set `autoscale`/`replicas` before `vcpu`/`memory`).

Antoniodeoliveirasegura:
Fixes #4562.

The issue was happening when switching compute-resource presets, like Performance to Standard. Auth/Storage would still show a fake `1:2 ratio` error because `applyPresetToForm` was validating every field as it was written.

So memory was getting checked while autoscale still had the old preset value, and when autoscale later got set to `false`, it didn't clear the error.

Fixed it by applying all the preset fields first without validating on each write, then letting the existing trailing `trigger()` run validation once at the end.

Also added a regression test for this.

---

Switching compute-resource presets in the dashboard could leave a stale "1:2 ratio" validation error on services because each `setValue` call validated mid-update against partially-applied form state. The fix defers validation until all preset fields are written, so a single final `trigger()` validates the coherent end state.

___

### **Change Type**
Bug fix

___

### **Description**
- Changed `applyPresetToForm` to write all preset fields with `shouldValidate: false`, then validate once via the existing trailing `trigger()` call.
- Added a test verifying no per-write validation occurs, that validation runs exactly once, and that no stale `auth.memory` / `storage.memory` ratio errors remain after applying a preset.

___

<tr><td><strong>Bug fix</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>useApplyPreset.ts</strong><dd><code>Set shouldValidate:false on per-field writes; rely on trailing trigger() for one final validation</code></dd></td>
  <td>+2/-1</td>
</tr>
</table></details></td></tr>
<tr>
  <td><strong>useApplyPreset.test.ts</strong><dd><code>New test asserting deferred validation and no stale ratio errors after preset switch</code></dd></td>
  <td>+85/-0</td>
</tr>
</table></details></td></tr>
</tbody></table>

</details>

___


dbm03 (approved):

dbm03:
Hi @Antoniodeoliveirasegura! Thank you so much for your contribution. I've reviewed it and it looks good to me.

I've merged the last changes from the `main` branch into your branch, and fixed something that the linter was complaining about. I'll be merging this PR soon

Thanks again!

Issue #3605, opened 2025-10-13
Add Phone Number to User Account


rsnider-wisecode:
**Is your feature request related to a problem? Please describe.**
After a user has authenticated with another method, we want to be able to add a verified phone number to their account as well.

**Describe the solution you'd like**
a new `/user/phone/change` endpoint that sends a OTP and then is verified with a new `/user/phone/change/otp` endpoint

**Describe alternatives you've considered**
I have tried to call `/signin/passwordless/sms` as an authenticated user, but it created a new user (as i suspected)

**Additional context**

Issue #2210, opened 2023-08-30
Allow users to update their phone number


nuurcodes:
Hi, currently there is no way to update a users phone number without changing permission on the auth.users table.

It would be nice if we could update the users phone number similar to the users email 
In react this could look like `useUpdatePhoneNumber()`

I have considered changing permission on the auth.users table to allow updates to phoneNumber, but this means I also need to manage the phoneNumberVerified column and set it to false every time they change the number. It would be nice if this logic was encapsulated in the useUpdatePhoneNumber hook that I am proposing.

This would also require a method to verify a users phone number.

PR #4576, opened 2026-06-22
feat(auth): add SMS deanonymization and phone number changes

### LINKED ISSUE #3604: De-anonymize User with SMS Passwordless
We are adding SMS auth to our app and discovered that there isn't a way to deanonymize a user that way.

update the `/user/deanonymize` endpoint so that:
* email is optional
* add an optional string phoneNumber

then a OTP is sent to the user, and `/signin/passwordless/sms/otp` is used to verify it

dbarrosop:
Fixes #3604 #3605 and #2210

Adds first-class SMS flows for deanonymizing anonymous users and changing verified phone numbers, while separating staged/unverified phone numbers from claimed verified numbers. It also updates the Go auth service, OpenAPI/SDK/docs, database schema, and route tests to support those flows end to end.

___

### **Change Type**
Enhancement

___

### **Description**
- Add `new_phone_number` staging, phone uniqueness checks, and controller/workflow support for SMS deanonymization and phone-number changes.
- Update OpenAPI schemas, JavaScript SDK auth methods, and docs for the new user SMS endpoints.
- Rework dev SMS delivery/test helpers and add route/controller coverage for phone verification, squat, and deanonymization behavior.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  SDK["JS SDK / API clients"] --> API["Auth HTTP routes"]
  API --> Controllers["Go controllers"]
  Controllers --> Workflows["Auth workflows"]
  Workflows --> DB[("auth.users new_phone_number")]
  Controllers --> SMS["SMS provider / dev output"]
  Tests["Route and controller tests"] --> API
```

<tr><td><strong>Documentation</strong></td><td><details><summary>2 files</summary><table><tr><td><strong>docs/src/content/docs/products/auth/otp/sms.mdx</strong><dd><code>Documents SMS OTP/deanonymization and phone-number-change flows</code></dd></td><td>+53/-1</td></tr>
<tr><td><strong>docs/src/content/docs/reference/javascript/nhost-js/auth.md</strong><dd><code>Adds JS auth reference entries for SMS deanonymization and phone changes</code></dd></td><td>+143/-0</td></tr></table></details></td></tr>
<tr><td><strong>API and SDK surface</strong></td><td><details><summary>3 files</summary><table><tr><td><strong>docs/src/schemas/auth.yaml</strong><dd><code>Adds OpenAPI schema/routes for SMS deanonymization and phone changes</code></dd></td><td>+142/-0</td></tr>
<tr><td><strong>services/auth/docs/openapi.yaml</strong><dd><code>Adds auth service OpenAPI routes for SMS deanonymization and phone changes</code></dd></td><td>+142/-0</td></tr>
<tr><td><strong>packages/nhost-js/src/auth/client.ts</strong><dd><code>Exposes SDK helpers/types for new SMS auth operations</code></dd></td><td>+196/-0</td></tr></table></details></td></tr>
<tr><td><strong>Auth service configuration</strong></td><td><details><summary>4 files</summary><table><tr><td><strong>services/auth/go/cmd/email.go</strong><dd><code>Adjusts SMS/email command wiring for dev SMS output handling</code></dd></td><td>+4/-5</td></tr>
<tr><td><strong>services/auth/go/cmd/serve.go</strong><dd><code>Adds server configuration for dev SMS output handling</code></dd></td><td>+7/-0</td></tr>
<tr><td><strong>services/auth/go/middleware/ratelimit/rate_limit.go</strong><dd><code>Includes new SMS-sending endpoints in SMS rate limiting</code></dd></td><td>+2/-0</td></tr>
<tr><td><strong>services/auth/go/providers/provider.go</strong><dd><code>Updates provider behavior around phone-based auth fields</code></dd></td><td>+2/-2</td></tr></table></details></td></tr>
<tr><td><strong>Go controllers and workflows</strong></td><td><details><summary>22 files</summary><table><tr><td><strong>services/auth/go/controller/change_user_phone_number.go</strong><dd><code>Adds change-phone-number request and confirmation handlers</code></dd></td><td>+79/-0</td></tr>
<tr><td><strong>services/auth/go/controller/change_user_phone_number_test.go</strong><dd><code>Covers change-phone-number handler success and error cases</code></dd></td><td>+375/-0</td></tr>
<tr><td><strong>services/auth/go/controller/controller.go</strong><dd><code>Registers new user SMS routes on the controller</code></dd></td><td>+11/-0</td></tr>
<tr><td><strong>services/auth/go/controller/deanonymize_user_sms.go</strong><dd><code>Adds SMS deanonymization handler flow</code></dd></td><td>+97/-0</td></tr>
<tr><td><strong>services/auth/go/controller/deanonymize_user_sms_test.go</strong><dd><code>Covers SMS deanonymization handler success and error cases</code></dd></td><td>+361/-0</td></tr>
<tr><td><strong>services/auth/go/controller/errors.go</strong><dd><code>Adds API errors for phone-number-change and deanonymization cases</code></dd></td><td>+14/-0</td></tr>
<tr><td><strong>services/auth/go/controller/get_user_test.go</strong><dd><code>Updates user assertions for staged phone-number state</code></dd></td><td>+1/-0</td></tr>
<tr><td><strong>services/auth/go/controller/link_id_token_test.go</strong><dd><code>Updates token/link tests for new user phone fields</code></dd></td><td>+4/-0</td></tr>
<tr><td><strong>services/auth/go/controller/sign_in_otp_email_test.go</strong><dd><code>Updates test fixtures for staged phone-number fields</code></dd></td><td>+5/-2</td></tr>
<tr><td><strong>services/auth/go/controller/sign_in_passwordless_email_test.go</strong><dd><code>Updates passwordless email tests for new phone-field expectations</code></dd></td><td>+7/-3</td></tr>
<tr><td><strong>services/auth/go/controller/sign_in_passwordless_sms_test.go</strong><dd><code>Updates passwordless SMS tests for new phone-field expectations</code></dd></td><td>+9/-3</td></tr>
<tr><td><strong>services/auth/go/controller/sign_in_provider_callback_get_test.go</strong><dd><code>Updates provider callback tests for new phone-field expectations</code></dd></td><td>+4/-2</td></tr>
<tr><td><strong>services/auth/go/controller/sign_up_email_password.go</strong><dd><code>Initializes staged phone-number fields during signup</code></dd></td><td>+1/-0</td></tr>
<tr><td><strong>services/auth/go/controller/sign_up_email_password_test.go</strong><dd><code>Updates signup tests for new phone-field expectations</code></dd></td><td>+4/-0</td></tr>
<tr><td><strong>services/auth/go/controller/sign_up_otp_email_test.go</strong><dd><code>Updates OTP signup tests for new phone-field expectations</code></dd></td><td>+1/-0</td></tr>
<tr><td><strong>services/auth/go/controller/sign_up_passwordless_email.go</strong><dd><code>Initializes staged phone-number fields for passwordless email signup</code></dd></td><td>+1/-0</td></tr>
<tr><td><strong>services/auth/go/controller/sign_up_passwordless_email_test.go</strong><dd><code>Updates passwordless email signup tests for new phone-field expectations</code></dd></td><td>+2/-0</td></tr>
<tr><td><strong>services/auth/go/controller/sign_up_passwordless_sms.go</strong><dd><code>Adjusts passwordless SMS signup around staged phone-number state</code></dd></td><td>+2/-1</td></tr>
<tr><td><strong>services/auth/go/controller/sign_up_passwordless_sms_test.go</strong><dd><code>Updates passwordless SMS signup tests for staged phone-number behavior</code></dd></td><td>+2/-1</td></tr>
<tr><td><strong>services/auth/go/controller/verify_sign_in_passwordless_sms.go</strong><dd><code>Completes staged SMS verification and anonymous SMS deanonymization</code></dd></td><td>+12/-0</td></tr>
<tr><td><strong>services/auth/go/controller/verify_sign_in_passwordless_sms_test.go</strong><dd><code>Covers SMS verification and deanonymization completion behavior</code></dd></td><td>+112/-0</td></tr>
<tr><td><strong>services/auth/go/controller/workflows.go</strong><dd><code>Adds workflow helpers for phone uniqueness, changes, and SMS deanonymization</code></dd></td><td>+178/-0</td></tr></table></details></td></tr>
<tr><td><strong>Database and SQL</strong></td><td><details><summary>4 files</summary><table><tr><td><strong>services/auth/go/migrations/postgres/00023_user_new_phone_number.up.sql</strong><dd><code>Adds new_phone_number and migrates unverified phone numbers into it</code></dd></td><td>+8/-0</td></tr>
<tr><td><strong>services/auth/go/migrations/postgres/00023_user_new_phone_number.down.sql</strong><dd><code>Drops staged phone-number column on rollback</code></dd></td><td>+7/-0</td></tr>
<tr><td><strong>services/auth/go/sql/auth_schema_dump.sql</strong><dd><code>Updates schema dump with staged phone-number column</code></dd></td><td>+1/-0</td></tr>
<tr><td><strong>services/auth/go/sql/query.sql</strong><dd><code>Adds SQL queries for staged phone-number flows and verified-phone uniqueness</code></dd></td><td>+84/-7</td></tr></table></details></td></tr>
<tr><td><strong>SMS notifications</strong></td><td><details><summary>5 files</summary><table><tr><td><strong>services/auth/go/notifications/sms/dev.go</strong><dd><code>Writes dev SMS messages to optional per-phone output files</code></dd></td><td>+50/-3</td></tr>
<tr><td><strong>services/auth/go/notifications/sms/dev_test.go</strong><dd><code>Covers dev SMS output-file behavior and sanitization</code></dd></td><td>+69/-0</td></tr>
<tr><td><strong>services/auth/go/notifications/sms/errors.go</strong><dd><code>Removes obsolete SMS verification error definitions</code></dd></td><td>+0/-5</td></tr>
<tr><td><strong>services/auth/go/notifications/sms/sms.go</strong><dd><code>Simplifies SMS interface away from provider-side verification checks</code></dd></td><td>+1/-13</td></tr>
<tr><td><strong>services/auth/go/notifications/sms/twilio_verification.go</strong><dd><code>Removes Twilio Verify implementation from SMS provider path</code></dd></td><td>+0/-83</td></tr></table></details></td></tr>
<tr><td><strong>Route tests and utilities</strong></td><td><details><summary>4 files</summary><table><tr><td><strong>services/auth/test/routes/user/deanonymize-sms.test.ts</strong><dd><code>Adds end-to-end route tests for SMS deanonymization</code></dd></td><td>+253/-0</td></tr>
<tr><td><strong>services/auth/test/routes/user/phone-number-change.test.ts</strong><dd><code>Adds end-to-end route tests for changing user phone numbers</code></dd></td><td>+222/-0</td></tr>
<tr><td><strong>services/auth/test/routes/user/phone-squat.test.ts</strong><dd><code>Adds route tests for phone-number squat and verified-claim behavior</code></dd></td><td>+295/-0</td></tr>
<tr><td><strong>services/auth/test/utils.ts</strong><dd><code>Adds SMS OTP helper utilities and dev output support for tests</code></dd></td><td>+32/-0</td></tr></table></details></td></tr>
</tbody></table>

</details>

___


services/auth/go/controller/workflows.go
```diff
@@ -1148,6 +1253,79 @@ func (wf *Workflows) DeanonymizeUser(
...
+		if err != nil {
+			logger.ErrorContext(ctx, "error marshalling metadata", logError(err))
+			return ErrInternalServerError
+		}
+	}
+
+	if err := wf.db.UpdateUserDeanonymizeSMS(
+		ctx,
+		sql.UpdateUserDeanonymizeSMSParams{
+			Roles:            *options.AllowedRoles,
```
skoveit, line 1349: Could we defer applying Roles until after the OTP is verified? The anonymous refresh token remains valid here, so POST /token can reload these new roles and issue an access token with non-anonymous permissions before phone ownership is proven

skoveit: The preceding DeleteUserRoles also means abandoning the flow permanently removes the anonymous role

services/auth/go/sql/query.sql
```diff
@@ -95,12 +97,17 @@ SELECT (
 
 -- name: GetUserByPhoneNumberAndOTP :one
 UPDATE auth.users
-SET otp_hash_expires_at = now(), phone_number_verified = true
+SET
+    phone_number = COALESCE(new_phone_number, phone_number),
```
skoveit: This assignment can promote a stale, unverified new_phone_number. If Y remains staged after an abandoned change and the user later signs in with their current number X, the WHERE clause matches X, but COALESCE selects Y. The OTP sent to X therefore verifies and promotes Y instead

services/auth/go/controller/workflows.go
skoveit: fixed: https://github.com/nhost/nhost/pull/4576/changes/c9787ce726f42ccb380bb5193399195df13fd232

services/auth/go/sql/query.sql
skoveit: fixed: https://github.com/nhost/nhost/pull/4576/changes/108eb541b6438e687658012c2718c9fb0d98d9e8

services/auth/go/notifications/sms/dev_internal_test.go
```diff
@@ -0,0 +1,98 @@
...
+func TestDevNoOutputDirIsNoop(t *testing.T) {
+	t.Parallel()
+
+	dev := &Dev{
+		logger:    slog.Default(),
+		outputDir: "",
+	}
+
+	if err := dev.SendSMS("+1234567890", "Your code is 123456."); err != nil {
+		t.Fatalf("SendSMS failed: %v", err)
```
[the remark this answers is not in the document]
skoveit: Intentionally not changing this one. The four tests are not variations on one input — they have materially different setups: `TestDevRewritesModeOfExistingFile` seeds a stale `0o600` file to assert the `os.Chmod` rewrite, `TestDevNoOutputDirIsNoop` constructs `Dev` with an empty `outputDir` and asserts no file work happens at all, and `TestDevSanitizesPhone` feeds a path-traversal string and asserts on a different filename than it passed in.

Collapsing those into a table needs per-case branching in the body plus fields that only apply to one case each, which reads worse than four named tests and makes a failure harder to locate. Keeping them separate.

docs/src/content/docs/reference/javascript/nhost-js/auth.md
```diff
@@ -4323,6 +4402,28 @@ signInMethod: UserDeanonymizeRequestSignInMethod;
...
 ---
 
+## UserDeanonymizeSMSRequest
+
+### Properties
+
+#### options?
+
+```ts
+optional options?: SignUpOptions;
```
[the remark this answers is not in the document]
skoveit: Skipping: options is a $ref to SignUpOptions, whose own properties already carry descriptions, and a sibling description next to a $ref is ignored in OpenAPI 3.0, so this would need an allOf wrapper plus an SDK/docs regen to have any visible effect.

services/auth/go/notifications/sms/dev.go
```diff
@@ -28,5 +38,51 @@ func NewDev(
 
 func (s *Dev) SendSMS(to string, body string) error {
 	s.logger.Info("Dev SMS sent", slog.String("to", to), slog.String("body", body))
+
+	if s.outputDir == "" {
+		return nil
+	}
+
+	name := sanitizePhoneForFilename(to)
+	if name == "" {
```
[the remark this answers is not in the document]
skoveit: Skipping: the auth container writes as root while the host test reads as a different uid, so `filePerm = 0o644` is load-bearing (dropping it to `0o600` breaks `readSMSCode`), and the `dirPerm = 0o700` never actually applies because Docker pre-creates the bind-mount point at `0755`

services/auth/go/notifications/sms/dev.go
```diff
@@ -28,5 +38,51 @@ func NewDev(
...
+	if err := os.WriteFile(path, []byte(body), filePerm); err != nil {
+		return err //nolint:wrapcheck
+	}
+
+	// os.WriteFile only applies the mode when it creates the file, so a file left
+	// behind by an older build keeps its original mode forever.
+	if err := os.Chmod(path, filePerm); err != nil {
+		return err //nolint:wrapcheck
+	}
+
```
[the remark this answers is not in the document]
skoveit: skip

services/auth/go/notifications/sms/dev.go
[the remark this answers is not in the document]
skoveit: Skipping: the premise here is inverted — the auth container writes as root while the host test reads as a different uid, so `filePerm = 0o644` is load-bearing (dropping it to `0o600` breaks `readSMSCode`), and the `dirPerm = 0o700` never actually applies because Docker pre-creates the bind-mount point at `0755`.

services/auth/go/notifications/sms/dev.go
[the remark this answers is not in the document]
skoveit: Skipping: `os.MkdirAll`, `os.WriteFile` and `os.Chmod` all return a `*PathError` that already carries both the operation and the path, so wrapping would add no call-site information beyond what the error already reports.

docs/src/content/docs/reference/javascript/nhost-js/auth.md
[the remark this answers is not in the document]
skoveit: Skipping: `options` is a `$ref` to `SignUpOptions`, whose own properties already carry descriptions — and a sibling `description` next to a `$ref` is ignored in OpenAPI 3.0, so this would need an `allOf` wrapper plus an SDK/docs regen to have any visible effect.

services/auth/go/middleware/ratelimit/rate_limit.go
```diff
@@ -36,6 +36,8 @@ func sendsSMS(path string) bool {
 	return slices.Contains([]string{
 		"/signin/passwordless/sms",
 		"/signup/passwordless/sms",
+		"/user/deanonymize/sms",
```
dbarrosop, line 39: I think we are missing documenting this in the docs

services/auth/go/controller/workflows.go
```diff
@@ -277,6 +277,30 @@ func (wf *Workflows) GetUser(
 	return user, nil
 }
 
+func (wf *Workflows) validateUserIsAnonymous(
```
dbarrosop, line 280: do we also need to check if the user is disabled or is that covered somewhere else?

services/auth/go/controller/workflows.go
```diff
@@ -754,6 +824,92 @@ func (wf *Workflows) ChangeEmail(
 	return user, nil
 }
 
+// PhoneNumberClaimedByOtherUser returns true only if another non-disabled user
```
dbarrosop: Why exclude disabled users here? The database uniqueness constraint also applies to disabled users, so this pre-check can report the number as available and send an SMS, only for verification to fail with a conflict

services/auth/go/controller/verify_sign_in_passwordless_sms_test.go
```diff
@@ -189,16 +196,55 @@ func TestVerifySignInPasswordlessSms(t *testing.T) { //nolint:maintidx
...
+					).Return(sql.AuthUser{}, "", errors.New("invalid OTP")) //nolint:err113
 
 					return mock
 				}),
 			},
 		},
 
 		{
-			name:   "user is disabled",
+			name:   "disabled non-anonymous user is ineligible for SMS verification",
```
dbarrosop: can you confirm the tests:

- `"disabled non-anonymous user is ineligible for SMS verification"` (lines 206-238)
- `"disabled anonymous user is ineligible for SMS deanonymization"` (lines 582-614)
- `"anonymous passwordless OTP without staged deanonymization is rejected before promotion"` (lines 721-753)

are correct, not duplicated, and test what they claim to test?

services/auth/go/sql/query.sql
```diff
@@ -439,6 +660,69 @@ WHERE user_id = $1;
 DELETE FROM auth.refresh_tokens
 WHERE expires_at < now();
 
+-- name: ReleaseExpiredStagedPhoneNumbers :exec
```
dbarrosop: This is very very very very scary. If we introduce some new mechanism and forget to update this we risk deleting those users. I would suggest leaving this out for now. I think a better thing to do would be to have some more generic clean up tasks user can enable to delete "orphaned" users after X days (out of scope though). For instance, something like:

```sql
   DELETE FROM auth.users
   WHERE last_seen IS NULL
     AND created_at < now() - interval '60 days';
```

This would be a more generic solution that could clean users and that project admins could simply add to their pg_cron

services/auth/go/sql/query.sql
```diff
@@ -439,6 +660,69 @@ WHERE user_id = $1;
...
+  AND NOT EXISTS (
+      SELECT 1 FROM auth.pkce_authorization_codes
+      WHERE user_id = users.id
+  )
+  AND NOT EXISTS (
+      SELECT 1 FROM auth.oauth2_clients
+      WHERE created_by = users.id
+  );
+
+-- name: ReleaseExpiredStagedPhoneNumberChanges :exec
```
dbarrosop: should we do something similar for the deanonymization part?

services/auth/go/sql/query.sql
```diff
@@ -388,6 +552,44 @@ SET (email, new_email) = (new_email, null)
...
+-- name: UpdateUserChangePhoneNumber :exec
+UPDATE auth.users
+SET
+    new_phone_number = @new_phone_number,
+    otp_hash = crypt(@otp, gen_salt('bf')),
+    otp_hash_expires_at = @otp_hash_expires_at,
+    otp_method_last_used = 'sms-change'
+WHERE id = @id;
+
+-- name: UpdateUserConfirmChangePhoneNumber :one
```
dbarrosop, line 625: do we need the same counting mechanism we use in VerifyEmailOTP in both verify endpoints related to SMS?

services/auth/go/sql/query.sql
```diff
@@ -93,15 +182,89 @@ SELECT (
...
+      AND candidate.is_anonymous = true
+    RETURNING users.*
+), verified AS (
+    UPDATE auth.users AS users
+    SET
+        phone_number = @phone_number,
+        new_phone_number = NULLIF(users.new_phone_number, @phone_number),
+        phone_number_verified = true,
+        otp_hash = NULL,
+        otp_hash_expires_at = now()
```
dbarrosop: do we need to clean pending_sms_deanonymize_options here too?

docs/src/content/docs/products/auth/otp/sms.mdx
```diff
@@ -78,11 +101,55 @@ await nhost.auth.signIn(
...
 
 :::note
 Phone numbers should start with `+` (not `00`) to follow the [E.164 formatting standard](https://en.wikipedia.org/wiki/E.164)
 :::
+
+## Change Phone Number
+
+An authenticated user can change the phone number on their account. This is a two-step flow: the user requests the change, an OTP is sent via SMS to the new number, and the user submits the OTP to confirm. The current phone number is left unchanged until verification succeeds.
+
+This endpoint requires [elevated permissions](/products/auth/elevated-permissions).
```
dbarrosop:
```suggestion
Both endpoints require [elevated permissions](/products/auth/elevated-permissions) when `auth.elevatedPrivileges.mode` is `required`, or when it is `recommended` and the user has a security key. With the default `disabled` mode no elevated claim is needed.```

File: docs/src/content/docs/products/auth/index.mdx [hunk 16]
```diff
@@ -16,7 +16,7 @@ Nhost Auth is a ready-to-use authentication service seamlessly integrated with t
 Use the Nhost SDK to sign in users. For example, with email and password:
 
 ```js
-await nhost.auth.signIn({
+await nhost.auth.signInEmailPassword({
```
dbarrosop, line 19: good catch, I think the signout method below is also wrong, could you check and fix as well? Thanks!

services/auth/test/routes/user/deanonymize-sms.test.ts
```diff
@@ -0,0 +1,516 @@
+import { StatusCodes } from 'http-status-codes';
```
dbarrosop: nice work, something we should test here as well is that denaonymizing invalidates the existing tokens

services/auth/test/utils.ts
```diff
@@ -1,14 +1,47 @@
...
-import { request } from './server';
-import { StatusCodes } from 'http-status-codes';
-import { generateTicketExpiresAt } from './src/ticket';
 import { hashPassword } from './src/password';
-import { ClientBase } from 'pg';
-import { v4 as uuidv4 } from 'uuid';
+import { generateTicketExpiresAt } from './src/ticket';
+
+// Kept in sync by hand with the bind mount in build/dev/docker/docker-compose.yaml.
+// The nix check derivation scrubs the environment, so this can't be configurable
```
dbarrosop: actually, not correct. Nix runs `bun test --env-file .env.example` so this can be made configurable by adding the value there.

services/auth/go/notifications/sms/dev.go
```diff
@@ -28,5 +38,51 @@ func NewDev(
...
+	name := sanitizePhoneForFilename(to)
+	if name == "" {
+		return errors.New("invalid phone number for output file") //nolint:err113
+	}
+
+	// The body is world-readable because the process that reads it back is not
+	// the one that wrote it, and the two don't share a uid. This is a dev-only
+	// provider and the same body is already logged in plaintext above.
+	const (
+		dirPerm  = 0o700
```
dbarrosop: this should be 0o755 to be consistent with `filePerm = 0o644`

services/auth/test/routes/user/deanonymize-sms.test.ts
```diff
@@ -0,0 +1,516 @@
...
+    expect(claims).toBeTruthy();
+    expect(
+      claims?.['https://hasura.io/jwt/claims']['x-hasura-user-is-anonymous'],
+    ).toBe('false');
+
+    // After OTP verification the user is non-anonymous and the OLD anonymous
+    // refresh token is revoked.
+    await request
+      .post('/token')
+      .send({ refreshToken: anonRefreshToken })
```
dbarrosop: I think this is not correct, you are using anonRefreshToken which was already consumed on line 67 so you need to grab the new refresh token there and use it here

services/auth/go/controller/workflows.go
skoveit: yes, during the OTP verification, but I’ll check it here too. that's better

services/auth/go/controller/workflows.go
skoveit: I removed the disabled = false filter

services/auth/go/sql/query.sql
skoveit: removed

docs/src/content/docs/products/auth/otp/sms.mdx
```diff
@@ -78,11 +101,55 @@ await nhost.auth.signIn(
...
 
 :::note
 Phone numbers should start with `+` (not `00`) to follow the [E.164 formatting standard](https://en.wikipedia.org/wiki/E.164)
 :::
+
+## Change Phone Number
+
+An authenticated user can change the phone number on their account. This is a two-step flow: the user requests the change, an OTP is sent via SMS to the new number, and the user submits the OTP to confirm. The current phone number is left unchanged until verification succeeds.
+
+Both endpoints require [elevated permissions](/products/auth/elevated-permissions) when `auth.elevatedPrivileges.mode` is `required`, or when it is `recommended` and the user has a security key. With the default `disabled` mode no elevated claim is needed.```
```
dbarrosop:
```suggestion
Both endpoints require [elevated permissions](/products/auth/elevated-permissions) when `auth.elevatedPrivileges.mode` is `required`, or when it is `recommended` and the user has a security key. With the default `disabled` mode no elevated claim is needed.
```

services/auth/go/migrations/postgres/00026_user_new_phone_number_data.up.sql
```diff
@@ -0,0 +1,13 @@
+UPDATE auth.users AS users
```
dbarrosop, line 4: I think we should narrow this one down a bit further with:

```sql
   last_seen IS NULL
   AND otp_method_last_used = 'sms'
```

or even further with every auth mechanism:

```sql
   UPDATE auth.users AS users
   SET new_phone_number = users.phone_number,
       phone_number = NULL
   WHERE users.phone_number IS NOT NULL
     AND users.new_phone_number IS NULL
     AND users.phone_number_verified = false
     AND users.email IS NULL
     AND users.new_email IS NULL
     AND users.email_verified = false
     AND users.password_hash IS NULL
     AND users.is_anonymous = false
     AND users.last_seen IS NULL
     AND users.otp_method_last_used = 'sms'
     AND users.pending_sms_deanonymize_options IS NULL
     AND users.totp_secret IS NULL
     AND users.active_mfa_type IS NULL
     AND users.ticket IS NULL
     AND users.webauthn_current_challenge IS NULL
     AND NOT EXISTS (
         SELECT 1
         FROM auth.user_providers AS provider
         WHERE provider.user_id = users.id
     )
     AND NOT EXISTS (
         SELECT 1
         FROM auth.user_security_keys AS security_key
         WHERE security_key.user_id = users.id
     )
     AND NOT EXISTS (
         SELECT 1
         FROM auth.refresh_tokens AS refresh_token
         WHERE refresh_token.user_id = users.id
     )
     AND NOT EXISTS (
         SELECT 1
         FROM auth.oauth2_auth_requests AS auth_request
         WHERE auth_request.user_id = users.id
     )
     AND NOT EXISTS (
         SELECT 1
         FROM auth.oauth2_refresh_tokens AS refresh_token
         WHERE refresh_token.user_id = users.id
     )
     AND NOT EXISTS (
         SELECT 1
         FROM auth.pkce_authorization_codes AS authorization_code
         WHERE authorization_code.user_id = users.id
     )
     AND NOT EXISTS (
         SELECT 1
         FROM auth.oauth2_clients AS client
         WHERE client.created_by = users.id
     );
```

I would rather "under migrate" to the new columns as that just means business as usual than "over migrate" as that might lead to users losing their phone claim. Basically it should align with `UpdateStagedSMSUser`

services/auth/go/migrations/postgres/00026_user_new_phone_number_data.down.sql
```diff
@@ -0,0 +1,33 @@
+-- The up migration moved pre-existing unverified phone_number values only from
+-- non-anonymous users without an email, password, or linked provider. Subsequent
+-- account changes and newly staged values make that original population
+-- indistinguishable at rollback time, so this best-effort inverse restores only
+-- the oldest claimant for each number and only when no user currently owns that
+-- phone_number.
+WITH restoration_candidates AS (
```
dbarrosop, line 4: I think we should do something similar to the proposed up migration here as I believe we may be over-fixing, potentially over-restoring. For instance:

```sql
   -- Restore only credential-free, unverified SMS-signup placeholders.
   -- Pending deanonymization and phone-change targets must not become ownership.
   WITH restoration_candidates AS (
       SELECT
           users.id,
           users.new_phone_number,
           row_number() OVER (
               PARTITION BY users.new_phone_number
               ORDER BY users.created_at, users.id
           ) AS claimant_order
       FROM auth.users AS users
       WHERE users.phone_number IS NULL
         AND users.new_phone_number IS NOT NULL
         AND users.phone_number_verified = false
         AND users.email IS NULL
         AND users.new_email IS NULL
         AND users.email_verified = false
         AND users.password_hash IS NULL
         AND users.is_anonymous = false
         AND users.last_seen IS NULL
         AND users.otp_method_last_used = 'sms'
         AND users.pending_sms_deanonymize_options IS NULL
         AND users.totp_secret IS NULL
         AND users.active_mfa_type IS NULL
         AND users.ticket IS NULL
         AND users.webauthn_current_challenge IS NULL
         AND NOT EXISTS (
             SELECT 1
             FROM auth.user_providers AS provider
             WHERE provider.user_id = users.id
         )
         AND NOT EXISTS (
             SELECT 1
             FROM auth.user_security_keys AS security_key
             WHERE security_key.user_id = users.id
         )
         AND NOT EXISTS (
             SELECT 1
             FROM auth.refresh_tokens AS refresh_token
             WHERE refresh_token.user_id = users.id
         )
         AND NOT EXISTS (
             SELECT 1
             FROM auth.oauth2_auth_requests AS auth_request
             WHERE auth_request.user_id = users.id
         )
         AND NOT EXISTS (
             SELECT 1
             FROM auth.oauth2_refresh_tokens AS refresh_token
             WHERE refresh_token.user_id = users.id
         )
         AND NOT EXISTS (
             SELECT 1
             FROM auth.pkce_authorization_codes AS authorization_code
             WHERE authorization_code.user_id = users.id
         )
         AND NOT EXISTS (
             SELECT 1
             FROM auth.oauth2_clients AS client
             WHERE client.created_by = users.id
         )
   ),
   restorable AS (
       SELECT candidate.id, candidate.new_phone_number
       FROM restoration_candidates AS candidate
       WHERE candidate.claimant_order = 1
         AND NOT EXISTS (
             SELECT 1
             FROM auth.users AS owner
             WHERE owner.phone_number = candidate.new_phone_number
         )
   )
   UPDATE auth.users AS target
   SET phone_number = restorable.new_phone_number,
       new_phone_number = NULL
   FROM restorable
   WHERE target.id = restorable.id;
```

services/auth/go/notifications/sms/sms_test.go
```diff
@@ -0,0 +1,208 @@
...
+
+	"github.com/google/uuid"
+	"github.com/jackc/pgx/v5/pgtype"
+	"github.com/nhost/nhost/services/auth/go/notifications/sms"
+	"github.com/nhost/nhost/services/auth/go/sql"
+)
+
+// stubVerifyDB implements sms.DB, records what it was called with, and returns
+// canned results so CheckVerificationCode's branches can be exercised without a
+// database. A same-package mock/ subdir cannot be imported from a black-box test
```
dbarrosop: I think this comment is bogus, there is no circular import here. Please, check and use mockgen.

services/auth/docs/openapi.yaml
```diff
@@ -501,7 +509,10 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/ErrorResponse"
-          description: "An error occurred while processing the request"
+          description: |
+            Authentication failed. Errors include `invalid-otp` (400) for invalid, expired, or ineligible verification
```
dbarrosop: We don't typically enumerate "manually" errors in the default response and having to manage this by hand is just going to be problematic. We are already missing some like `otp-too-many-attempts`. I would suggest just avoiding this enumeration.

services/auth/go/sql/query.sql
```diff
@@ -388,6 +599,87 @@ SET (email, new_email) = (new_email, null)
 WHERE id = $1
 RETURNING *;
 
+-- name: GetVerifiedUserByPhoneNumberOtherThanSelf :one
```
dbarrosop: I think we should call this `GetUserByPhoneNumberOtherThanSelf` and drop `phone_number_verified = true`. Otherwise using this we may think we are safe to write phone_number but the fact is that we aren't because there is a UNIQUE constraint on the column regardless of the verification column's value.

services/auth/test/routes/user/deanonymize-sms.test.ts
```diff
@@ -0,0 +1,1084 @@
+import { readFileSync } from 'node:fs';
```
dbarrosop, line 1: can we test a correct code that expired for the deanonimization process?

services/auth/test/routes/user/phone-squat.test.ts
```diff
@@ -0,0 +1,664 @@
...
+    const otp = readSMSCode(phoneNumber);
+
+    await request
+      .post('/user/phone-number/change/verify')
+      .set('Authorization', `Bearer ${accessToken}`)
+      .send({ newPhoneNumber: phoneNumber, otp })
+      .expect(StatusCodes.OK);
+
+    // Y owns the verified phone; the unverified signup row remains staged until
+    // its OTP expires and the debris sweep reclaims it.
```
dbarrosop: I am not sure this comment is accurate looking at the SQL and the test. We are basically testing the new_phone_number is still set.

Also, this cleaning is nice but might be unnecessary. Let's do the same we do for change-email-password. If we are not cleaning new_email and if needed we can tackle cleaning that column and this one.

services/auth/go/sql/query.sql
skoveit: Good catch! this also covers the migration case where existing users can keep an unverified `phone_number`, which still reserves the value through the unique constraint. [fixed](https://github.com/nhost/nhost/pull/4576/changes/71168133aa80c5da37c533e36923ec92d2c80734#diff-97f747248104c1546792d48460adc342691cf9541b0f0970a773c5835668d41a)

services/auth/docs/openapi.yaml
```diff
@@ -478,8 +478,16 @@ paths:
   /signin/passwordless/sms/otp:
     post:
       operationId: verifySignInPasswordlessSms
-      summary: Verify SMS OTP
-      description: Complete passwordless SMS authentication by verifying the one-time password. Returns a session if validation is successful.
+      summary: Verify SMS OTP and complete authentication
+      description: |
+        Complete passwordless SMS authentication by verifying the one-time password and returning a session.
```
dbarrosop: Let's not leak internal details and enumerate errors to avoid drift

skoveit (approved):

PR #4587, opened 2026-06-23
fix(cli): route container-to-container traffic via traefik aliases

meh:
### Checklist

- [x] No breaking changes
- [x] Tests pass

On Linux, `host-gateway` resolves to the default docker0 bridge
gateway, which is unroutable from containers attached to a user-defined
project bridge. As a result, services in the nhost stack could not
reach each other over the public `local.<service>.local.nhost.run`
hostnames: HTTPS calls hung and timed out after ~30s. The same names
worked from the host (where /etc/hosts pins them to 127.0.0.1 and
traefik is published on the configured port), masking the issue on
Docker Desktop where the VM's NAT happens to route the gateway
correctly across bridges.

Replace the per-service `extra_hosts` mappings with network aliases on
the traefik service. Docker's embedded DNS on the project bridge now
resolves these hostnames to traefik's container IP. Same-bridge
routing works, TLS validates (Traefik's certificates already cover
these SANs), and the existing ingress labels handle the rest.

The standalone hasura-cli helper (`docker run` without --network)
continues to use the legacy host-gateway mapping via a separate
`hostGatewayHosts` helper, since it is not attached to the project
bridge and the default docker0 gateway can reach the host-published
traefik port from there.

The console container keeps its <subdomain>.hasura.local.nhost.run →
0.0.0.0 override so hasura-cli's API discovery loops back to itself
instead of round-tripping through traefik.

Closes the long-standing gap that forced users to introduce a
`BACKEND_FUNCTIONS_URL=http://functions:3000` indirection to invoke
functions from Hasura actions locally; `{{NHOST_FUNCTIONS_URL}}` now
works uniformly across local and cloud.


cli/dockercompose/compose.go
```diff
@@ -126,7 +126,49 @@ type Volume struct {
...
 }
 
-func extraHosts(subdomain string) []string {
+// extraHosts returns the /etc/hosts entries injected into every bridge
+// service. Public local.nhost.run hostnames are intentionally absent: on
+// Linux, `host-gateway` resolves to the default docker0 bridge gateway,
+// which is unroutable from containers attached to the user-defined
+// project bridge. Resolution for those hostnames is provided by network
+// aliases on the traefik service (see traefikAliases).
+func extraHosts(_ string) []string {
```
dbarrosop: remove arg if unused

meh: That ends up removing the `subdomain` parameter up to the root of the call tree, I think it's better to keep it even if unused for now.

dbarrosop: and isn't that good? why would you pass parameters around you don't need? you wouldn't be passing it if you were writing this from scratch or am I missing something?

dbarrosop (approved):

PR #4598, opened 2026-06-25
fix(dashboard): fallback to getFile download for storage files in private buckets

### LINKED ISSUE #4552: FR: make image preview viewable without presigned urls
> **Note:** Feature requests that are clearly AI-generated will not be accepted and will be closed immediately. Please write your feature request in your own words.

Currently when I go to the storage section in the dashboard I can see all the images with their small preview. However when I click on them the preview popup will not load because they aren't in a bucket where presigned urls are enabled.

Since I'm in the admin dashboard I'd like to be able to see the image in the preview popup, regardless of if presigned urls is enabled or not.

**Describe alternatives you've considered**
Copying the uuid and fetching the image with the admin secret manually. Not that pleasant of an experience :/

**Additional context**
-

Dakshx07:
### Checklist

- [x] No breaking changes
- [x] Tests pass

---

### Description

Fixes #4552.

Currently, when files (images, videos, audio, or general documents) reside in a private bucket with presigned URLs disabled, the dashboard is unable to render previews or open/download the files, as `getFilePresignedURL` calls fail.

This PR implements a frontend fallback mechanism for both the preview dialog and the data grid file opening cell:
1. **Fallback fetching:** If retrieving the presigned URL fails, the dashboard catches the error and downloads the file as a `Blob` using `appClient.storage.getFile` with the `x-hasura-admin-secret` header.
2. **Object URLs:** The blob is converted to a local Object URL (`blob:`) to render image/video/audio previews inside the dialog or to open non-previewable files (like PDFs or ZIPs) in a new tab via `window.open`.
3. **Memory management:** Generated Object URLs are automatically revoked when the preview dialog unmounts or when the URL changes. For files opened in new tabs, the Object URL is revoked after a 60-second timeout to allow the browser tab sufficient time to load the content.
4. **Test Environment Fixes:** Resolved testing environment limitations in JSDOM where `localStorage` and `matchMedia` were unavailable at import-time during vitest execution.

### Tests

- Added new unit tests in `DataGridPreviewCell.test.tsx` verifying both the preview dialog and handleOpenPreview fallback paths.
- Ran all storage tests locally (`pnpm test:vitest src/features/orgs/projects/storage`), with all 176 tests passing successfully.


dashboard/src/setupTests.ts
```diff
@@ -21,4 +21,41 @@ class ResizeObserverMock {
 // Stub the global ResizeObserver
 vi.stubGlobal('ResizeObserver', ResizeObserverMock);
 
+// Stub the global matchMedia to prevent matchMedia warnings/errors
+vi.stubGlobal(
```
robertkasza: Could you tell me please why we need to add these mocks globally?

dashboard/src/features/orgs/projects/storage/dataGrid/components/DataGridPreviewCell/FilePreviewDialog.tsx
```diff
@@ -40,17 +41,47 @@ export default function FilePreviewDialog({
...
-      }
+        if (!body?.url) {
+          throw new Error('Presigned URL could not be fetched.');
+        }
+
+        return body.url;
+      } catch (presignedError) {
+        // Fallback: download the file as a Blob using the admin secret
+        try {
+          const { res } = await appClient.storage.getFile(
```
robertkasza: Res does not exist on getFile return type. This is the return type:
```
export interface FetchResponse<T> {
  /** The parsed response body */
  body: T; <- this is Blob
  /** HTTP status code of the response */
  status: number;
  /** Response headers */
  headers: Headers;
}
```

dashboard/src/features/orgs/projects/storage/dataGrid/components/DataGridPreviewCell/FilePreviewDialog.tsx
```diff
@@ -40,17 +41,47 @@ export default function FilePreviewDialog({
   } = useQuery({
     queryKey: ['file-presigned-url', id],
     queryFn: async () => {
-      const { body } = await appClient.storage.getFilePresignedURL(id, {
-        headers: {
-          'x-hasura-admin-secret': project!.config!.hasura.adminSecret,
-        },
-      });
+      try {
+        const { body } = await appClient.storage.getFilePresignedURL(id, {
```
robertkasza: We are doing the same logic in two places I think we should move the logic in to a common function. What do you think?

dashboard/src/features/orgs/projects/storage/dataGrid/components/DataGridPreviewCell/FilePreviewDialog.tsx
```diff
@@ -40,17 +41,47 @@ export default function FilePreviewDialog({
...
-        throw new Error('Presigned URL could not be fetched.');
-      }
+        if (!body?.url) {
+          throw new Error('Presigned URL could not be fetched.');
+        }
+
+        return body.url;
+      } catch (presignedError) {
+        // Fallback: download the file as a Blob using the admin secret
+        try {
```
robertkasza: We know whether the presigned URLs are enabled or not at this point, I think we should use that to decide how we want to fetch the images.

dashboard/src/features/orgs/projects/storage/dataGrid/components/FilesDataGrid/FilesDataGrid.test.tsx
```diff
@@ -67,7 +67,10 @@ const server = setupServer(
   getFilesAggregateHandler,
 );
 
-beforeAll(() => server.listen());
+beforeAll(() => {
+  server.listen();
+  global.localStorage = localStorageMock();
```
robertkasza, line 74: Could you please tell me why this was added?

dashboard/src/setupTests.ts
Dakshx07: We stubbed `matchMedia` and `localStorage` globally because they are referenced during module evaluation/render-time by UI component libraries (like Radix UI) and storage/theme hook utilities. Stubbing them globally in `setupTests.ts` prevents import-time errors and warnings across all dashboard test suites.

Also, I have removed the now-redundant manual `global.localStorage = localStorageMock()` setup from `FilesDataGrid.test.tsx` since it's covered by the global stub.

dashboard/src/features/orgs/projects/storage/dataGrid/components/DataGridPreviewCell/FilePreviewDialog.tsx
Dakshx07: Good catch! You are absolutely right. I've updated the implementation of the fallback download. We now destructure `{ body: blob, status }` from the response (which is of type `FetchResponse<Blob>`) and verify the `status` and `blob` directly instead of assuming it has a `.res` property. The unit tests have also been updated to match this API shape.

dashboard/src/features/orgs/projects/storage/dataGrid/components/DataGridPreviewCell/FilePreviewDialog.tsx
Dakshx07: Agreed! I've moved the fetching and fallback logic to a new common utility function `getFileUrlOrFallback` in `dashboard/src/features/orgs/projects/storage/dataGrid/components/DataGridPreviewCell/utils.ts`. Both `FilePreviewDialog` and `DataGridPreviewCell` now import and share this helper.

dashboard/src/features/orgs/projects/storage/dataGrid/components/DataGridPreviewCell/FilePreviewDialog.tsx
Dakshx07: Excellent point. I've updated the `getFileUrlOrFallback` helper to accept `presignedUrlsEnabled` (which is passed down from the `bucket` metadata in `FilesDataGrid`). If it's explicitly set to `false`, we now skip the `getFilePresignedURL` attempt entirely and directly fetch using the admin secret, avoiding unnecessary network round-trips and console errors.

dashboard/src/features/orgs/projects/storage/dataGrid/components/FilesDataGrid/FilesDataGrid.test.tsx
Dakshx07: This was originally added locally before we moved the `localStorage` stub to `setupTests.ts`. Since the mock is now handled globally, I have removed this redundant block.

dashboard/src/setupTests.ts
robertkasza: I understand why matchmedia was mocked, but not localstorage. It is probably better to mock match media globally, but right now there is two matchmedia mocks and localStorage too. 

I would like to ask you that remove both global mocks and add the matchmedia mock to the DataGridPreviewCell.test.tsx, like in [TransferProjectDialog.test.tsx](https://github.com/nhost/nhost/blob/main/dashboard/src/features/orgs/components/common/TransferProjectDialog/TransferProjectDialog.test.tsx#L26).

We should add the global mock later, if needed. I do not think adding the global mocks and cleaning up belongs to this PR.

dashboard/src/features/orgs/projects/storage/dataGrid/components/DataGridPreviewCell/DataGridPreviewCell.test.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+    mocks.useAppClient.mockReturnValue({
+      storage: {
+        getFilePresignedURL: getFilePresignedURLMock,
+        getFile: getFileMock,
+      },
+    });
+
+    const createObjectURLSpy = vi.spyOn(URL, 'createObjectURL').mockReturnValue('blob:http://localhost/test-blob-url');
+    const revokeObjectURLSpy = vi.spyOn(URL, 'revokeObjectURL').mockImplementation(() => {});
```
robertkasza: I believe this is not used.

dashboard/src/features/orgs/projects/storage/dataGrid/components/DataGridPreviewCell/DataGridPreviewCell.test.tsx
```diff
@@ -0,0 +1,130 @@
...
+    });
+
+    mocks.useAppClient.mockReturnValue({
+      storage: {
+        getFilePresignedURL: getFilePresignedURLMock,
+        getFile: getFileMock,
+      },
+    });
+
+    const createObjectURLSpy = vi.spyOn(URL, 'createObjectURL').mockReturnValue('blob:http://localhost/pdf-blob-url');
```
robertkasza: These two mocks are not used.

dashboard/src/features/orgs/projects/storage/dataGrid/components/DataGridPreviewCell/DataGridPreviewCell.test.tsx
```diff
@@ -0,0 +1,130 @@
+import { vi } from 'vitest';
+import { render, screen, waitFor } from '@/tests/testUtils';
+import userEvent from '@testing-library/user-event';
```
robertkasza: For user events please use [TetsUserEvent from testUtils.tsx](https://github.com/Dakshx07/nhost/blob/64959c2f1b3c1b87ddaeaee08aab3d69f39bc023/dashboard/src/tests/testUtils.tsx#L208).

robertkasza (commented):
Had a few comments on the tests, please take a look. I also pushed a commit to fix biome errors.
FYI you can run biome in the dashboard using npx:  npx @biomejs/biome@2.4.15 check (--write)
Just make sure that you use that same version as the dashboard, which you can find here: [biome version](https://github.com/Dakshx07/nhost/blob/64959c2f1b3c1b87ddaeaee08aab3d69f39bc023/nixops/overlays/js.nix#L4)

dashboard/src/features/orgs/projects/storage/dataGrid/components/DataGridPreviewCell/DataGridPreviewCell.test.tsx
Dakshx07: Removed! I have cleaned up the unused variables and only mock the function return value directly now.

dashboard/src/features/orgs/projects/storage/dataGrid/components/DataGridPreviewCell/DataGridPreviewCell.test.tsx
Dakshx07: Removed! I have cleaned up the unused variables and only mock the function return value directly now.

dashboard/src/features/orgs/projects/storage/dataGrid/components/DataGridPreviewCell/DataGridPreviewCell.test.tsx
Dakshx07: Updated! I replaced the `userEvent` import with `TestUserEvent` from `@/tests/testUtils` and used `TestUserEvent.fireClickEvent(button)` to trigger the click actions.

dashboard/src/setupTests.ts
Dakshx07: Done! I have reverted `setupTests.ts` completely to remove the global mocks. I've also added the local `matchMedia` mock inside `DataGridPreviewCell.test.tsx` using `mockMatchMediaValue` to keep it clean and localized. Additionally, I restored `localStorageMock` locally inside `DataGridPreviewCell.test.tsx` and `FilesDataGrid.test.tsx` to prevent any import-time errors.

robertkasza (approved):
LGTM, Thanks!

PR #4601, opened 2026-06-26
fix(dashboard): align additional headers input fields

lakshitsoni26:
### Checklist

- [x] No breaking changes
- [x] Tests pass

--- 

### Description

This PR fixes a vertical alignment bug in the "Additional Headers" form section inside the dashboard's Event Triggers page. 

**Root Cause:**
- `FormSelect` was unconditionally rendering an empty `<FormLabel>`, which was triggering a default `space-y-2` margin from its parent `FormItem` wrapper. This caused the "Value" dropdown (and the adjacent "Header value" input) to be pushed down by 8px.
- The colon `:` separator and the "Trash" icon button had hardcoded top-padding (`pt-2` and `pt-3`) which forced them out of vertical alignment with the `h-10` form inputs.

**Changes:**
1. Updated `FormSelect.tsx` to only render `<FormLabel>` if a label is actually provided (matching the behavior of `FormInput.tsx`).
2. Removed the hardcoded `pt-2` and `pt-3` padding on the colon and trash icon in `HeadersFormSection.tsx`.
3. Added `h-10 items-center` to the trash button wrapper to ensure perfect vertical centering with the standard form inputs.

### Before
Screenshot 2026-06-26 at 11 25 24 PM

### After
Screenshot 2026-06-26 at 11 21 23 PM


robertkasza (approved):

robertkasza:
Thanks for the contribution!

PR #4609, opened 2026-06-29
feat(internal/lib): add screenshot tool

PVUL:
Screenshot 2026-06-30 at 2 32 41 AM

This is a screenshot tool designed to help us take better and more consistent screenshots for our docs, but honestly can be used on any website. It's a chrome extension, so only runs on chromium compatible browsers, like brave, etc. Must build the extension locally and load it manually to install. Details in the [`README.md`](https://github.com/nhost/nhost/tree/screenshot-tool/tools/screenshot-extension#readme). 

The thing that makes this different than anything else I've used, is the spotlight feature. When in `select` mode, it allows you to select any element on the page, and it adds an outline to that element, while masking out the background with dim. I packed a lot of cool features in here actually, so hopefully this is useful to y'all.

I'll show this off on July 1 during team coffee hour, but if you want to see what all the fuss is about:
Demo: https://drive.google.com/file/d/1hkXG29Z5PswaRmePoxRNvnV4wQ8j8Bxs/view


dbarrosop:
I think we should probably create a dedicated repo for this. I am closing it, I think you have permissions to create repos, otherwise, let me know.

PR #4632, opened 2026-07-08
fix(docs): align specs with reality

meh:
### Checklist

- [x] No breaking changes
- [x] Tests pass

The auth server returns user.metadata: null for users without metadata (a nil Go map marshals to null), but the OpenAPI spec declared metadata as a required, non-nullable object. Clients that validate responses at runtime (e.g. the Python SDK via pydantic) reject the null; the TS SDK never noticed because its types are erased at runtime.

Mark metadata nullable: true so the spec matches actual server behaviour, and regenerate the oapi-codegen server/types (User.Metadata becomes a pointer). Wrap the controller assignments in a nil-preserving pointer helper (mapPtr) so a nil map still marshals to null, keeping JSON output identical. Update the affected controller test expectations accordingly.

Alternative is to always return a default `{}` object but that would change the schema.

Also UploadFiles' metadata[] parts are JSON objects, but the multipart body had no
encoding declaration, so the per-part content type fell back to the OpenAPI
default that several client generators ignore, they sent the part as
text/plain and hasura-storage's openapi3filter rejected it ('value must be an
object'). Declare the content type explicitly via an encoding block so the
requirement lives in the contract rather than in each SDK's codegen. file[] is
left unset so binary parts keep their detected content type.

---

Not sure about either but while working on the new SDKs the actual behavior showed inconsistencies with the declared specs.


services/auth/go/controller/controller.go
```diff
@@ -40,6 +40,18 @@ func deptr[T any](x *T) T { //nolint:ireturn
 	return *x
 }
 
+// mapPtr returns a pointer to m, or nil when m is nil, so a nil metadata map
```
dbarrosop: nitpick but the comment isn't correct. `map[string]any` is already a pointer and thus an empty `map[string]any` is already null (same for slices) so json.Marshal without `omitempty` will actually emit `"field": null`. This is mostly needed to satisfy the weird type used by the codegen.

dbarrosop: Looking a bit more into this:

https://github.com/oapi-codegen/oapi-codegen/blob/main/docs/configuration.md

I think using `prefer-skip-optional-pointer-on-container-types: false` might be cleaner as this pointer on a type that is already a pointer is unnecessary and a bit ugly.

If changing it globally has some unintended consequences we can also target the fix with something like:

```
   # in the openapi schema
   metadata:
     type: object
     additionalProperties: true
     nullable: true
     x-go-type-skip-optional-pointer: true
```

or the other way around, change it globally and add `x-go-type-skip-optional-pointer: false` on existing fields where we already use the indirect pointer (that way future optional fields will have the correct behavior by default).

dbarrosop (commented):
For HS256/HS384/HS512, `decodeJWTSecret` returns a nil JWK slice. `GET /.well-known/jwks.json` passes it through and emits `{"keys":null}`, while `JWKSet.keys` is a required, non-nullable array in the OpenAPI spec. `null` is also not a valid JWK Set representation under RFC 7517 §5.1; deployments with symmetric signing should publish `{"keys":[]}`.

services/auth/docs/openapi.yaml
```diff
@@ -3229,6 +3229,8 @@ components:
         metadata:
           type: object
           additionalProperties: true
+          nullable: true
+          x-go-type-skip-optional-pointer: true
```
dbarrosop, line 3336: **Warning — this nullability does not reach the TypeScript SDK.**

The spec change is correct, but the repository's TypeScript generator does not read `nullable`. Regenerating from this spec leaves `packages/nhost-js/src/auth/client.ts` byte-identical, with `metadata: Record<string, unknown>` rather than `Record<string, unknown> | null`. Consumers can therefore still hit a runtime error with no type-level warning when the service returns `metadata: null`.

Please either teach `tools/codegen` to emit `T | null` and regenerate the SDK, or explicitly defer that broader type-level change (it affects the other nullable fields too) and track it as a follow-up.

services/storage/controller/openapi.yaml
```diff
@@ -64,6 +64,14 @@ paths:
...
               required:
                 - file[]
+            encoding:
+              # metadata[] parts are JSON objects. Declaring the per-part
+              # content type explicitly (OpenAPI's default for object-typed
+              # multipart parts, which some client generators ignore) so tools
+              # send them as application/json rather than text/plain. file[] is
+              # intentionally left unset so binary parts keep their own type.
+              metadata[]:
+                contentType: application/json
```
dbarrosop, line 69: **Warning — the same multipart encoding is missing from `replaceFile`.**

This fixes `POST /files`, but `PUT /files/{id}` has the same object-typed multipart part (`metadata` at lines 480-481) without an `encoding` entry. Both paths parse JSON metadata in the same way, so generators that ignore the OpenAPI default can still send replacement metadata as `text/plain`.

Add the corresponding `encoding.metadata.contentType: application/json` to the `replaceFile` request body and regenerate the embedded and published specs.

dbarrosop (approved):

PR #4644, opened 2026-07-10
feat(engine): add a single binary including all services

meh:
### Checklist

- [x] No breaking changes
- [x] Tests pass

Nhost's Go services (auth, storage, and the constellation GraphQL engine) could only run as separate processes/containers. This adds a single `nhost-engine` binary that runs any subset of them in one process behind one shared listener, with unified `NHOST_*` configuration, one logger, and a single lifecycle — while the standalone binaries keep working unchanged.

___

### **Change Type**
Enhancement

___

### **Description**
- Extracts the per-service logger setup into a shared `internal/lib/serve` package (`NewLogger`/`LogFlags`); each service's `logger.go` becomes a thin delegator.
- Adds `internal/lib/serve/service.go` with a reusable `Service{Handler, Background, Close}` abstraction, and gives each service an exported `NewService(...)` that builds its handler + lifecycle without owning the listener.
- Adds `services/nhost-engine/` — CLI grammar (`[shared] SERVICE [flags] -- SERVICE ...`), a supervisor (`internal/runner`), service registry, and `compose.go` that mounts every selected service under a path prefix (`/auth`, `/storage`, `/graphql`, `/healthz`) behind one `http.Server`.
- Introduces unified `NHOST_*` shared configuration (bind/port, debug, log format, HTTP timeouts, admin/JWT secret, database + migrations URLs, CORS) injected onto each service with service-wins precedence.
- Wires the new module into `flake.nix` (check, devShell, package, docker image) reusing storage's libvips.
- Adds README plus an illustrative docker-compose example; keeps the standalone auth/storage/constellation binaries as a supported compatibility path.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  CLI["nhost-engine CLI<br/>[shared] a -- s -- g"] --> CFG["parseSharedFlags<br/>NHOST_* + precedence"]
  CFG --> BLD["buildService x N<br/>NewService()"]
  BLD --> MUX["shared mux<br/>/auth /storage /graphql /healthz"]
  MUX --> SRV["single http.Server"]
  BLD --> BG["Background goroutines"]
  SRV --> SUP["runner.Supervise"]
  BG --> SUP
```

<tr><td><strong>Shared runtime</strong></td><td><details><summary>3 files</summary><table>
<tr><td><strong>internal/lib/serve/logger.go</strong><dd><code>New shared logger setup (NewLogger/LogFlags/isSecret)</code></dd></td><td>+120/-0</td></tr>
<tr><td><strong>internal/lib/serve/service.go</strong><dd><code>Reusable Service{Handler,Background,Close} + Run/Shutdown helpers</code></dd></td><td>+46/-0</td></tr>
<tr><td><strong>internal/lib/serve/logger_internal_test.go</strong><dd><code>Logger test relocated from constellation</code></dd></td><td>+4/-4</td></tr>
</table></details></td></tr>
<tr><td><strong>nhost-engine</strong></td><td><details><summary>9 files</summary><table>
<tr><td><strong>services/nhost-engine/main.go</strong><dd><code>Entry point: parse shared flags, run/supervise services</code></dd></td><td>+95/-0</td></tr>
<tr><td><strong>services/nhost-engine/compose.go</strong><dd><code>Shared config, override mapping, mux composition, supervisor wiring</code></dd></td><td>+457/-0</td></tr>
<tr><td><strong>services/nhost-engine/registry.go</strong><dd><code>Service registry, prefixes, usage text</code></dd></td><td>+137/-0</td></tr>
<tr><td><strong>services/nhost-engine/internal/runner/runner.go</strong><dd><code>CLI split grammar + sync-based supervisor</code></dd></td><td>+184/-0</td></tr>
<tr><td><strong>services/nhost-engine/compose_internal_test.go</strong><dd><code>Routing/prefix-stripping tests</code></dd></td><td>+90/-0</td></tr>
<tr><td><strong>services/nhost-engine/config_internal_test.go</strong><dd><code>Bind/port resolution + override precedence tests</code></dd></td><td>+174/-0</td></tr>
<tr><td><strong>services/nhost-engine/internal/runner/runner_internal_test.go</strong><dd><code>Split + Supervise tests</code></dd></td><td>+230/-0</td></tr>
<tr><td><strong>services/nhost-engine/project.nix</strong><dd><code>Nix package/devShell/check/docker image</code></dd></td><td>+142/-0</td></tr>
<tr><td><strong>services/nhost-engine/Makefile</strong><dd><code>Delegates to shared makefile</code></dd></td><td>+2/-0</td></tr>
</table></details></td></tr>
<tr><td><strong>Service refactors</strong></td><td><details><summary>6 files</summary><table>
<tr><td><strong>services/auth/go/cmd/serve.go</strong><dd><code>Exported NewService; serve builds it then runs listener</code></dd></td><td>+63/-18</td></tr>
<tr><td><strong>services/constellation/cmd/serve.go</strong><dd><code>Exported NewService; getHandler; extracted helpers</code></dd></td><td>+111/-46</td></tr>
<tr><td><strong>services/storage/cmd/serve.go</strong><dd><code>Exported NewService; getHandler returns http.Handler</code></dd></td><td>+85/-35</td></tr>
<tr><td><strong>services/auth/go/cmd/logger.go</strong><dd><code>Thin delegator to serve pkg</code></dd></td><td>+8/-89</td></tr>
<tr><td><strong>services/constellation/cmd/logger.go</strong><dd><code>Thin delegator to serve pkg</code></dd></td><td>+8/-92</td></tr>
<tr><td><strong>services/storage/cmd/logger.go</strong><dd><code>Thin delegator to serve pkg</code></dd></td><td>+8/-89</td></tr>
</table></details></td></tr>
<tr><td><strong>Build & docs</strong></td><td><details><summary>2 files</summary><table>
<tr><td><strong>flake.nix</strong><dd><code>Wire engine check/devShell/package/docker image</code></dd></td><td>+14/-0</td></tr>
<tr><td><strong>services/nhost-engine/README.md</strong><dd><code>Usage, routing, config, docker-compose example</code></dd></td><td>+132/-0</td></tr>
</table></details></td></tr>
</tbody></table>

</details>

___


services/nhost-engine/README.md
```diff
@@ -0,0 +1,132 @@
...
+deployment can co-host these services without running (and networking) a
+separate container per service.
+
+The standalone `auth`, `storage`, and `constellation` binaries remain fully
+supported; the engine is an additional way to run the same code, not a
+replacement.
+
+## Command grammar
+
+Services are selected on the command line and separated by `--`. Shared flags
```
dbarrosop: Let's KISS it, the point of the nhost-engine is to run all services so we shoudln't need to complicate things allowing enabling/disabling stuff, we already have the individual services for that and if we want to disable something I would expect to have something simple like `--disable-auth`. Also, regarding shared and explicit flags, I would expect something like `nhost-engine --bind :8080 --auth-some-flag --storage-some-flag --graphql-some-other-flag ...`. Keep in mind the goal is to REPLACE the individual binaries with this, not support both paths so this command line should look like what we want to have by the end of this work.

services/auth/go/cmd/logger.go
```diff
@@ -3,101 +3,20 @@ package cmd
...
-		strings.Contains(name, "license") ||
-		strings.Contains(name, "postgres") ||
-		strings.Contains(name, "client-id") ||
-		strings.Contains(name, "client-secret")
+	return serveutil.NewLogger(debug, formatTEXT)
 }
 
+// logFlags logs the resolved flag values at startup, redacting secrets. It
+// delegates to the shared serve package.
 func logFlags(ctx context.Context, logger *slog.Logger, cmd *cli.Command) {
```
dbarrosop: this method feels unnecessary

services/auth/go/cmd/logger.go
```diff
@@ -3,101 +3,20 @@ package cmd
...
-	"github.com/gin-gonic/gin"
-	"github.com/lmittmann/tint"
+	serveutil "github.com/nhost/nhost/internal/lib/serve"
 	"github.com/urfave/cli/v3"
 )
 
+// getLogger builds the structured logger for the auth server. It delegates to
+// the shared serve package so logger construction stays identical across every
+// Nhost service binary.
 func getLogger(debug bool, formatTEXT bool) *slog.Logger {
```
dbarrosop: this method feels unnecessary

services/constellation/cmd/logger.go
```diff
@@ -3,104 +3,20 @@ package cmd
...
+	serveutil "github.com/nhost/nhost/internal/lib/serve"
 	"github.com/urfave/cli/v3"
 )
 
-const redactedSecret = "********"
-
+// getLogger builds the structured logger for the constellation server. It
+// delegates to the shared serve package so logger construction stays identical
+// across every Nhost service binary.
 func getLogger(debug bool, formatTEXT bool) *slog.Logger {
```
dbarrosop: this method feels unnecessary

services/constellation/cmd/logger.go
```diff
@@ -3,104 +3,20 @@ package cmd
...
-		strings.Contains(name, "postgres") ||
-		strings.Contains(name, "client-id") ||
-		strings.Contains(name, "client-secret") ||
-		strings.Contains(name, "metadata-database-url")
+	return serveutil.NewLogger(debug, formatTEXT)
 }
 
+// logFlags logs the resolved flag values at startup, redacting secrets. It
+// delegates to the shared serve package.
 func logFlags(ctx context.Context, logger *slog.Logger, cmd *cli.Command) {
```
dbarrosop: this method feels unnecessary

services/storage/cmd/logger.go
```diff
@@ -3,101 +3,20 @@ package cmd
...
-	"github.com/gin-gonic/gin"
-	"github.com/lmittmann/tint"
+	serveutil "github.com/nhost/nhost/internal/lib/serve"
 	"github.com/urfave/cli/v3"
 )
 
+// getLogger builds the structured logger for the storage server. It delegates
+// to the shared serve package so logger construction stays identical across
+// every Nhost service binary.
 func getLogger(debug bool, formatTEXT bool) *slog.Logger {
```
dbarrosop: this method feels unnecessary

services/storage/cmd/logger.go
```diff
@@ -3,101 +3,20 @@ package cmd
...
-		strings.Contains(name, "license") ||
-		strings.Contains(name, "postgres") ||
-		strings.Contains(name, "client-id") ||
-		strings.Contains(name, "client-secret")
+	return serveutil.NewLogger(debug, formatTEXT)
 }
 
+// logFlags logs the resolved flag values at startup, redacting secrets. It
+// delegates to the shared serve package.
 func logFlags(ctx context.Context, logger *slog.Logger, cmd *cli.Command) {
```
dbarrosop: this method feels unnecessary

dbarrosop:
As noted on a previous comments, I think we need some "product" improvements here and if we need to change flags / env vars on the main services to better support that that's fine. I would expect the CLI to be simpler and more ergonomic, without the need for anything complex. For instance `nhost-engine serve` should work out of the box without anything else needed. Then we should be able to pass things like `--disable-auth` to not start auth (although this is optional at this point) and then have global flags and prefixed flags i.e. `nhost-engine --bind :8080 --auth-some-feature --storage-some-feature --graphql-some-feature ...`. Similar for env vars (`BIND`, `AUTH_SOMETHING`, STORAGE_SOMETHING`, etc...). Exposing all flags might not be necessary (i.e. postgres connection) and new might need to be introduced.

Ideally we would replace the CLI already to use this nhost-engine as that's going to be a good test for this work.

go.mod
```diff
@@ -230,3 +229,5 @@ require (
 )
 
 replace github.com/buger/jsonparser => github.com/dbarrosop/jsonparser v1.1.2
+
+replace github.com/nhost/be => /home/meh/work/nhost/be-nhost-engine-config
```
dbarrosop: I was going to test it as part of the review but unfortunately I couldn't because of this replace : (

.github/workflows/cli_wf_e2e.yaml
```diff
@@ -0,0 +1,106 @@
...
+
+      - name: "Get artifacts"
+        uses: actions/download-artifact@v8
+        with:
+          path: ~/artifacts
+
+      - name: "Load CLI image (configserver)"
+        run: |
+          skopeo copy --insecure-policy \
+            "dir:$HOME/artifacts/cli-docker-image-x86_64-0.0.0-dev" \
```
dbarrosop: is it possible we are missing the OS?

.github/workflows/ci_final_status_check.yaml
dbarrosop: add engine here

.github/workflows/ci_final_status_check.yaml
```diff
@@ -27,7 +27,7 @@ jobs:
           echo "Validating PR title: $PR_TITLE"
 
           VALID_TYPES="feat|fix|chore|release"
-          VALID_PKGS="auth|ci|cli|codegen|constellation|dashboard|deps|docs|examples|functions|ghactivity|govulncheck-wrapper|internal\/lib|mcp|nhost-js|nixops|postgres|storage|observability|stripe-graphql-js"
+          VALID_PKGS="auth|ci|cli|codegen|constellation|dashboard|deps|docs|examples|functions|ghactivity|govulncheck-wrapper|internal\/lib|mcp|nhost-js|nixops|postgres|storage|observability|stripe-graphql-js|engine"
```
dbarrosop: nitpick, alphabetically is easier to read/update

services/engine/compose.go
```diff
@@ -0,0 +1,468 @@
...
+
+	var built *serveutil.Service
+
+	app := &cli.Command{ //nolint:exhaustruct
+		Name:    name,
+		Version: version,
+		Usage:   serveCmd.Usage,
+		Flags:   serveCmd.Flags,
+		Action: func(ctx context.Context, c *cli.Command) error {
+			if err := applySharedConfig(c, name, cfg); err != nil {
```
dbarrosop, line 392: I don't think this is working as you expect but I couldn't test it due to the issue in `go.mod` so I may be wrong or I may be misunderstanding things. You are running this apply magic in the "Action" step but at this point all the flags have already been validated so any required flag should have triggered an error

.github/workflows/cli_checks.yaml
```diff
@@ -1,12 +1,17 @@
 ---
 name: "cli: check and build"
 on:
-  pull_request_target:
```
dbarrosop, line 4: adding a comment here to revert this as it's a bit more visible :P

.github/workflows/engine_wf_release.yaml
```diff
@@ -0,0 +1,69 @@
...
+      DOCKER_USERNAME:
+        required: true
+      DOCKER_PASSWORD:
+        required: true
+
+jobs:
+  build_artifacts:
+    uses: ./.github/workflows/wf_build_artifacts.yaml
+    with:
+      # `engine` matches the nix attr (services/engine/project.nix `name`), the
```
dbarrosop: let's remove this comment

cli/dockercompose/engine.go
```diff
@@ -0,0 +1,413 @@
...
+		); err != nil {
+			return nil, err
+		}
+	}
+
+	// Constellation is merged before auth: both set HASURA_GRAPHQL_DATABASE_URL
+	// (constellation to the postgres superuser, auth to nhost_hasura). Only
+	// auth reads it (via --postgres); constellation reads only
+	// CONSTELLATION_METADATA_DATABASE_URL, so letting auth's value win is
+	// correct.
```
dbarrosop: this claim isn't correct, is it? From a test case from real data:

https://github.com/nhost/nhost/blob/main/services/constellation/metadata/internal/hasura/testdata/TestFromJSON_RealMetadata/golden.json#L8

I don't think we won't trigger this issue here but it might get confusing and problematic as a general rule if we have different services competing for the same env vars. So two things related to this:

1. Can we ensure env vars don't overlap?
2. Shouldn't we be using `appconfig.NhostEngine` or similar instead of reusing appconfig.{Auth,Storage,Constellation}? Otherwise we are doing different things in the cloud and here

As a result I would expect to see only either global variables that affect the entire sevice (i.e. bind, log-level, debug, etc) or prefixed env vars (plus the user-defined ones and the NHOST_* ones related to the environment), right now there seems to be a bit of a mixture.

services/engine/compose.go
```diff
@@ -0,0 +1,537 @@
...
+	// headers, mirroring the per-service standalone servers.
+	readHeaderTimeout = 5 * time.Second
+	// shutdownTimeout bounds the graceful shutdown of the shared server once the
+	// process context is cancelled.
+	shutdownTimeout = 30 * time.Second
+
+	// The shared server's request timeouts default to constellation's values,
+	// since it is the service most sensitive to them (long-lived GraphQL
+	// responses); auth and storage previously left them unset.
+	defaultHTTPReadTimeout  = 30 * time.Second
```
dbarrosop: this might actually be problematic for storage as uploading a file might take longer than 30s. Maybe we can ship for now without `defaultHTTPReadTimeout` and `defaultHTTPWriteTimeout` leaving that unbounded and leave these protections to the cloud instead (the LB already sets sane values for these things)

services/engine/compose.go
```diff
@@ -0,0 +1,537 @@
...
+	disabled map[string]bool
+}
+
+// globalFlags defines the engine's shared flag surface. These consolidate the
+// settings common to every service: the listener, logging, shared secrets,
+// database URLs, and CORS origins, plus the --disable-<service> opt-outs. They
+// use bare env vars (BIND, ADMIN_SECRET, ...) because the engine replaces the
+// individual service binaries rather than running alongside them.
+func globalFlags() []cli.Flag {
+	flags := make([]cli.Flag, 0, numSharedGlobalFlags+len(serviceOrder()))
```
dbarrosop: I suspect we will forget about `numSharedGlobalFlags` as this is a bit of a strange pattern. Alternatively you could do:

```go
flags := []cli.Flag{
   ...
}

serviceOrderFlags := make([]cli.Flag, len(serviceOrder())
for i, name := range serviceOrder() {
   serviceOrderFlags[i] = ...
}

flags = append(flags, serviceOrderFlags...)
```

Or just append directly to flags in the in the for loop. This is a one time operation on startup so saving a few nanoseconds in garbage collection and/or growing slices is basically negligible.

PR #4645, opened 2026-07-10
feat(nhost-sdk-run): Nhost Run helpers for all supported languages

meh:
### Checklist

- [x] No breaking changes
- [x] Tests pass

Not sure how worth this is, but at least makes it easier to extend things in the future.


dbarrosop:
I am not sure about this one tbh. It is nice to save a bit of boilerplate but then you are missing a lot of functionality compared to using something like FastAPI, Gin, Express.js, etc... Feels like it might be better to have templates/tutorials/quickstarts instead which also allows us to adapt quickly to any future HTTP API framework without having to maintain another package.

meh:
Yeah I wasn't sure either, in practice it's usable with fastapi or express and whatnot (the python example uses fastapi), it only provides a thin layer with the foundational stuff.

PR #4682, opened 2026-07-20
fix(ci): tag local docker images with nhost/ registry prefix

meh:
Running `make docker-build-image` would make a locally installed  `cli` complain there is no container for this version of the CLI. Not sure this is the right fix but at least it makes it work.

### Checklist

- [x] No breaking changes
- [x] Tests pass

### DIFF
--- a/build/makefiles/general.makefile
+++ b/build/makefiles/general.makefile
@@ -30,6 +30,11 @@ ifeq ($(CI),true)
   docker-build-options=--option system $(ARCH)-linux --extra-platforms ${ARCH}-linux
 endif
 
+# Container registry namespace used to tag locally-built images so they match
+# the `nhost/<name>` references used in CI (see wf_docker_push_image.yaml) and
+# by the CLI runtime (e.g. `nhost dev up` looks for `nhost/cli:<version>`).
+CONTAINER_REGISTRY?=nhost
+
 
 .PHONY: help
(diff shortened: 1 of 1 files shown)


build/makefiles/general.makefile
```diff
@@ -30,6 +30,11 @@ ifeq ($(CI),true)
   docker-build-options=--option system $(ARCH)-linux --extra-platforms ${ARCH}-linux
 endif
 
+# Container registry namespace used to tag locally-built images so they match
+# the `nhost/<name>` references used in CI (see wf_docker_push_image.yaml) and
```
dbarrosop, line 34: `wf_docker_push_image.yaml` doesn't use the docker daemon at all, it loads the image straight from `./result` with skopeo.

build/makefiles/general.makefile
```diff
@@ -30,6 +30,11 @@ ifeq ($(CI),true)
...
+# by the CLI runtime (e.g. `nhost dev up` looks for `nhost/cli:<version>`).
+CONTAINER_REGISTRY?=nhost
```
dbarrosop, line 36: I would rename this to `LOCAL_CONTAINER_REGISTRY`, we already use `CONTAINER_REGISTRY` as part of the release process to declare where we want to push to so there is a bit of a collision there (even though it has no impact right now).

PR #4720, opened 2026-07-23
fix(cli): SELinux bind mount label

sepples:
### Checklist

- [x] No breaking changes
- [x] Tests pass

Adds a bind mount label to generated docker-compose configuration so that mount works on SELinux-enabled host environments (e.g. Podman machine, Red Hat/Fedora-based systems)

### DIFF
--- a/cli/dockercompose/compose.go
+++ b/cli/dockercompose/compose.go
@@ -120,7 +120,7 @@ type Port struct {
 	Protocol  string `yaml:"protocol"`
 }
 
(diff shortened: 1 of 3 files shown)


sepples:
While this does fix SELinux issues (e.g. on a Red Hat-based system), in practice it does not fix the issue for users of Podman as the relabel does not propagate through podman's compatibility symlink. This can be resolved in a different PR.

dbarrosop (approved):

PR #4772, opened 2026-08-04
fix(cli): sanitize nhost init pg_dump migration for consumption by hasura

meh:
### Checklist

- [ ] No breaking changes
- [ ] Tests pass

### DIFF
--- a/cli/dockercompose/docker.go
+++ b/cli/dockercompose/docker.go
@@ -193,56 +193,9 @@ func (d *Docker) HasuraWrapper( //nolint:funlen
 // `syntax error at or near "\"`. Dropping them keeps the dump valid for Hasura
 // while leaving the actual DDL untouched.
 //
-// It runs on both the `init --remote` generation path and the `up` apply path
-// so migrations that reach disk by any route (older CLIs, checked-in tutorial
-// repos, out-of-band dumps) apply cleanly.
-// MigrationsNeedSanitizing reports whether any .sql migration under
-// migrationsDir carries pg_dump's \restrict/\unrestrict psql meta-commands.
-// It lets the apply path stay on the fast, direct hasura-cli route unless
-// stripping is actually required.
-func MigrationsNeedSanitizing(migrationsDir string) (bool, error) {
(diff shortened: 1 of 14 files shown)


cli/dockercompose/docker.go
```diff
@@ -118,15 +164,89 @@ func (d *Docker) HasuraWrapper( //nolint:funlen
...
+// to harden restores done through psql. Hasura applies migrations by sending
+// the raw SQL to Postgres over its query API rather than piping it through
+// psql, so those backslash meta-commands are seen as SQL and fail with
+// `syntax error at or near "\"`. Dropping them keeps the dump valid for Hasura
+// while leaving the actual DDL untouched.
+//
+// It runs only on the `init --remote` generation path, so the initial
+// migration reaches disk already clean. Migrations applied by `nhost up` are
+// left untouched and applied as-is.
+func SanitizeMigrations(migrationsDir string) error {
```
dbarrosop, line 199: I would suggest moving all this code to `cmd/project/init.go`. It isn't necessarily a `dockercompose` action in any way and the only caller according to the docstring is `init --remote`, that way we don't need to export this either.

cli/dockercompose/docker.go
```diff
@@ -23,6 +25,50 @@ func NewDocker() *Docker {
...
+func (b *tailBuffer) String() string {
+	return string(b.buf)
+}
+
+// pgDumpVersionMismatch inspects hasura-cli output for the error Hasura
+// returns when its bundled pg_dump is older than the remote PostgreSQL
+// server (e.g. dumping a PG 18 database with pg_dump 17 during
+// `nhost init --remote`). The raw error is opaque, so translate it into
+// actionable guidance while leaving the original output visible above.
+func pgDumpVersionMismatch(output string) error {
```
dbarrosop, line 58: this made me realize I think we have a general bug where we are calling docker asynchronously in many places but we are not calling `Wait()`, which is why you ended up having to implement this instead of properly relying on the status code. Can you take a look, please? I think the underlying issue is in this code path but also in others.

PR #4777, opened 2026-08-05
feat(dashboard): native queries relationships

dbm03:
### Checklist

- [X] No breaking changes
- [X] Tests pass

## Goal

Let users manage object and array relationships between native queries, so one native query's results can be joined to another native query in the GraphQL API.

**Type:** Enhancement

## Changes

- Add a relationships view with create, edit, and delete flows for native queries.
- Add a form for native-query targets with configurable column mappings.
- Persist relationship changes through the existing native-query metadata mutation path.
- Open relationship management from native-query sidebar actions without navigating away.

## Flow

```mermaid
flowchart LR
  A[Native query sidebar] --> B[Relationships drawer]
  B --> C[List relationships]
  C --> D[Create or edit mapping]
  C --> E[Delete relationship]
  D --> F[Native-query metadata mutation]
  E --> F
```

## Key files

- `.../components/NativeQueryRelationships/` (+357/-0) — relationship list, target display, and persistence flow
- `.../components/RelationshipFormDialog/` (+430/-0) — create and edit form for native-query targets
- `.../components/EditNativeQueryRelationships/` and sidebar integration (+112/-3) — drawer entry point and actions
- `.../types/` and `.../utils/updateNativeQueryRelationship/` (+175/-0) — relationship inputs and immutable metadata transforms
- `6 test files` (+1674/-0) — drawer, sidebar, form, list, and transform coverage


dashboard/src/features/orgs/projects/database/native-queries/components/EditNativeQueryRelationships/EditNativeQueryRelationships.test.tsx
```diff
@@ -0,0 +1,317 @@
...
+
+    expect(await screen.findByText('1 object · 0 array')).toBeInTheDocument();
+    expect(screen.getByText('manager')).toBeInTheDocument();
+    expect(screen.getByText('· 1 mapping(s)')).toBeInTheDocument();
+    expect(screen.getByRole('link', { name: 'authors' })).toHaveAttribute(
+      'href',
+      '/orgs/test/projects/local/database/native-queries/default/queries/authors',
+    );
+
+    fireEvent.click(screen.getByRole('button', { name: 'Back' }));
```
robertkasza, line 166: Why are we using fireevent? instead of user?

dashboard/src/features/orgs/projects/database/native-queries/components/EditNativeQueryRelationships/EditNativeQueryRelationships.test.tsx
```diff
@@ -0,0 +1,317 @@
...
+    expect(onCancel).toHaveBeenCalledOnce();
+  });
+
+  it('keeps Back available when the native query no longer exists', async () => {
```
robertkasza, line 170: When is this case can happen in the real world?

dashboard/src/features/orgs/projects/database/native-queries/components/EditNativeQueryRelationships/EditNativeQueryRelationships.test.tsx
```diff
@@ -0,0 +1,317 @@
...
+    chooseOption('Target Native Query', 'authors');
+    fireEvent.click(screen.getByRole('button', { name: 'Add New Mapping' }));
+    expect(
+      screen
+        .getAllByRole('combobox')
+        .filter((element) => element.textContent === 'analytics_id'),
+    ).toHaveLength(2);
+  });
+
+  it('uses refreshed metadata for a second relationship mutation', async () => {
```
robertkasza, line 243: What are we testing int his case?

dashboard/src/features/orgs/projects/database/native-queries/components/EditNativeQueryRelationships/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './EditNativeQueryRelationships';
```
robertkasza, line 1: I think this is not used outside.

dashboard/src/features/orgs/projects/database/native-queries/components/NativeQueriesBrowserSidebar/LogicalModelListItem/LogicalModelListItem.tsx
```diff
@@ -10,8 +10,8 @@ import type { LogicalModelItem } from '@/utils/hasura-api/generated/schemas';
 const EditLogicalModelPermissionsForm = dynamic(
   () =>
     import(
-      '@/features/orgs/projects/database/native-queries/components/EditLogicalModelPermissionsForm'
-    ).then((mod) => mod.EditLogicalModelPermissionsForm),
+      '@/features/orgs/projects/database/native-queries/components/EditLogicalModelPermissionsForm/EditLogicalModelPermissionsForm'
```
robertkasza: This import does not seem right.

dashboard/src/features/orgs/projects/database/native-queries/components/NativeQueryRelationships/NativeQueryRelationships.test.tsx
```diff
@@ -0,0 +1,832 @@
+import { HttpResponse, http } from 'msw';
+import { setupServer } from 'msw/node';
+import { act } from 'react';
+import { toast } from 'react-hot-toast';
+import { useDialog } from '@/components/common/DialogProvider';
+import { EXPORT_METADATA_QUERY_KEY } from '@/features/orgs/projects/common/hooks/useExportMetadata';
+import { NativeQueryRelationships } from '@/features/orgs/projects/database/native-queries/components/NativeQueryRelationships';
+import { mockMatchMediaValue } from '@/tests/mocks';
+import {
+  fireEvent,
```
robertkasza, line 10: Let's keep using UserEvent

dashboard/src/features/orgs/projects/database/native-queries/components/NativeQueryRelationships/NativeQueryRelationships.test.tsx
```diff
@@ -0,0 +1,832 @@
...
+    act(() => toast.remove());
+    expect(unexpectedRequests).toEqual([]);
+    if (!mocks.isPlatform) {
+      expect(metadataBodies).toEqual([]);
+    }
+  });
+
+  afterAll(() => server.close());
+
+  it('awaits one atomic migration on add and preserves all other native query data', async () => {
```
robertkasza, line 324: Maybe it is me, but I cannot tell what this test supposed to test just based on description.

dashboard/src/features/orgs/projects/database/native-queries/components/NativeQueryRelationships/NativeQueryRelationships.test.tsx
```diff
@@ -0,0 +1,832 @@
...
+  it('edits through one atomic migration while changing only the target relationship', async () => {
+    render(
+      <NativeQueryRelationships
+        source={selectedSource}
+        query={query}
+        queries={[query]}
+        models={[model]}
+      />,
+    );
+    fireEvent.click(
```
robertkasza, line 448: let's use userEvent or whatever it is called.

dashboard/src/features/orgs/projects/database/native-queries/components/NativeQueryRelationships/NativeQueryRelationships.test.tsx
```diff
@@ -0,0 +1,832 @@
...
+        screen.getByRole('button', { name: 'Edit relationship owner' }),
+      ).toBeDisabled();
+      expect(
+        screen.getByRole('button', { name: 'Delete relationship owner' }),
+      ).toBeEnabled();
+      expect(getQueryHref).not.toHaveBeenCalled();
+    },
+  );
+
+  it('edits through one atomic migration while changing only the target relationship', async () => {
```
robertkasza, line 439: Again, it might be just me, but what edits through one atomic means? or in general, because there are other description like this.

dashboard/src/features/orgs/projects/database/native-queries/components/RelationshipFormDialog/index.ts
```diff
@@ -0,0 +1 @@
+export * from './RelationshipFormDialog';
```
robertkasza, line 1: This is not how we usually export in index.

dashboard/src/features/orgs/projects/database/native-queries/components/RelationshipFormDialog/RelationshipFormDialog.tsx
```diff
@@ -0,0 +1,429 @@
...
+  }));
+
+  return (
+    <>
+      <Form {...form}>
+        <form
+          className="flex flex-col gap-5"
+          onSubmit={form.handleSubmit(async (nextValues) => {
+            if (
+              hasNativeQueryRelationshipName(
```
robertkasza, line 187: Why are we validating this way. Doesn't zod can do this out of the box?

dashboard/src/features/orgs/projects/database/native-queries/components/RelationshipFormDialog/RelationshipFormDialog.tsx
```diff
@@ -0,0 +1,429 @@
...
+                <div
+                  key={mapping.id}
+                  className="grid grid-cols-12 items-center gap-2"
+                >
+                  <FormSelect
+                    control={control}
+                    name={`fieldMappings.${index}.sourceField`}
+                    placeholder="Select source field"
+                    label={
+                      <span className="sr-only">Source field {index + 1}</span>
```
robertkasza, line 259: sr only label? Why? just curios.

dashboard/src/features/orgs/projects/database/native-queries/utils/nativeQueryRelationships/index.ts
```diff
@@ -0,0 +1 @@
+export * from './nativeQueryRelationships';
```
robertkasza, line 1: not the usual export that we do.

dashboard/src/features/orgs/projects/database/native-queries/components/RelationshipFormDialog/RelationshipFormDialog.tsx
```diff
@@ -0,0 +1,429 @@
...
+      </Form>
+
+      <DiscardChangesDialog
+        open={showDiscardDialog}
+        onOpenChange={setShowDiscardDialog}
+        onDiscardChanges={() => {
+          setShowDiscardDialog(false);
+          onClose();
+        }}
+        // Keeps Escape from reaching the MUI drawer hosting this dialog.
```
robertkasza, line 377: I am not a big fan of jsx comments. And why not add this stopPropogation to the component it self? Based on how our forms work feels like we should add this every time we use this DiscardChangesDialog.

Since we are talking about discard. Shouldn't we alert the user of unsaved changes when we are editing something and haven't saved it yet. I mean this is how discard alerts work in general? Just a question really.

dashboard/src/features/orgs/projects/database/native-queries/components/NativeQueryRelationships/NativeQueryRelationships.test.tsx
dbm03: The atomic is just the type of operation from the Hasura API, it's a `bulk_atomic`. https://hasura.io/docs/2.0/api-reference/metadata-api/index/#request-types

If any of the two steps that make the operation fail, then it fails completely and recovers the state "nicely". It's like, either both steps make the change, or no change is made at all. If it was a `bulk` it could happen that the first step is done and second fails, and you're left with an inconsistent state

dashboard/src/features/orgs/projects/database/native-queries/components/RelationshipFormDialog/RelationshipFormDialog.tsx
dbm03: I'm going to move the `stopPropagation` to the DiscardChangesDialog directly, so I don't need to pass the `onEscapeKeyDown` prop.

I was checking first because I didn't want to change a component that is shared in other drawer forms; but it seems fine in others too.

> Since we are talking about discard. Shouldn't we alert the user of unsaved changes when we are editing something and haven't saved it yet. I mean this is how discard alerts work in general? Just a question really.

What do you mean by this? like adding something to the Save button that says something like "Draft, click to save your changes"? I guess we would have to speak about this with the designers if that's the case

dashboard/src/features/orgs/projects/database/native-queries/types/index.ts
```diff
@@ -0,0 +1 @@
+export * from './nativeQueryRelationships';
```
dbm03: I'm now doing
```ts
export type LogicalModelTypeNode =
  | { kind: 'scalar'; scalar: string; nullable: boolean }
  | { kind: 'logical_model'; logicalModel: string; nullable: boolean }
  | { kind: 'array'; item: LogicalModelTypeNode; nullable: boolean };

// ... other types ...
```

dashboard/src/features/orgs/projects/database/native-queries/components/RelationshipFormDialog/RelationshipFormDialog.tsx
dbm03: It's good for screenreader users, and also it lets me use `getByRole('combobox', { name: 'Source field 1' })` in the tests.


If I removed the `sr-only` it would make it very visually redundant I think. It's better to hide it and keep it only for screen readers in my opinion
Screenshot 2026-09-18 at 16 00 49

dashboard/src/features/orgs/projects/database/native-queries/components/RelationshipFormDialog/index.ts
dbm03: Changed this into:
```
export * from './RelationshipFormDialog';
export { default as RelationshipFormDialog } from './RelationshipFormDialog';
```

dashboard/src/features/orgs/projects/database/native-queries/components/EditNativeQueryRelationships/index.ts
dbm03: I've changed this to just 
```
export { default as EditNativeQueryRelationships } from './EditNativeQueryRelationships';
```

dashboard/src/features/orgs/projects/database/native-queries/components/EditNativeQueryRelationships/EditNativeQueryRelationships.test.tsx
dbm03: This was testing the path here where a native query no longer exists. It could happen if you click the Edit relationships in one browser tab and you delete that query in another tab.
```tsx
{query ? (
          <NativeQueryRelationships
            query={query}
            queries={queries}
            models={models}
            getQueryHref={(targetQueryName) =>
              `/orgs/${orgSlug}/projects/${appSubdomain}/database/native-queries/default/queries/${encodeURIComponent(targetQueryName)}`
            }
          />
        ) : (
          <p className="rounded-md bg-muted p-4 text-muted-foreground text-sm">
            Native query {queryName} no longer exists.
          </p>
        )}
```

I deleted this test because it's very unlikely to happen and I'd rather not have an extra test. Also the test wasn't accurate because it was checking that the Back button is clickable, and it's always enabled

dashboard/src/features/orgs/projects/database/native-queries/components/EditNativeQueryRelationships/EditNativeQueryRelationships.test.tsx
dbm03: I changed this to use `user.click` instead. However, i had to keep `fireEvent.submit()` to submit the forms, since the user.click on a `type="submit"` button doesn't submit in jsdom for some reason

dashboard/src/features/orgs/projects/database/native-queries/components/NativeQueryRelationships/NativeQueryRelationships.test.tsx
dbm03: I've changed this test description to "sends a single metadata request when adding and preserves other query config"

that's to avoid having duplicate fetching, and also guard against the regression of not fetching right before adding a new relationship. 

```ts
        const { data: latestQueries, error } = await refetchNativeQueries();
```
If we removed that line from NativeQueryRelationships.tsx the test would fail, since adding two relationships in a row would make the first one disappear from the sent metadata update

dashboard/src/features/orgs/projects/database/native-queries/components/NativeQueryRelationships/NativeQueryRelationships.test.tsx
dbm03: I've changed this into user.click

dashboard/src/features/orgs/projects/database/native-queries/components/RelationshipFormDialog/RelationshipFormDialog.tsx
dbm03: I've moved this logic inside the validation schema's `superRefine` that we already had above

dashboard/src/features/orgs/projects/database/native-queries/components/NativeQueryRelationships/NativeQueryRelationships.test.tsx
dbm03: I had to keep using fireEvent but only for submitting the forms, since jsdom doesn't submit them with user.click for some reason. 
I've replaced all the rest with `user.click`, `user.keyboard`, etc.

dashboard/src/features/orgs/projects/database/native-queries/components/EditNativeQueryRelationships/EditNativeQueryRelationships.test.tsx
dbm03: I renamed this test to "rebuilds each save from refetched metadata so an earlier relationship is not dropped, same refetch guard I described in the other reply (https://github.com/nhost/nhost/pull/4777#discussion_r4049705199)

If the guard wasn't there and we forgot the refetch, adding a relationship and editing a relationship in a row would make the first one disappear

PR #4538, opened 2026-06-15
fix(cli): make sure local files belong to the user running nhost

meh:
### **What this PR solves**
On Linux, `nhost up` containers run as root and write migrations, metadata, generated config, and a default `tsconfig.json` into the user's bind-mounted project — leaving root-owned files the developer can't edit or delete. This change runs the file-producing services (and the `hasura-cli` wrapper) as the host user and pre-creates bind-mount sources as the caller, so everything in the repo stays owned by the user who ran the command.

___

### **PR Type**
Bug fix

___

### **Description**
- Added a `User` field to the compose `Service` model and an `applyHostUserID` pass that stamps `user: <uid>:<gid>` on `console`, `configserver`, and `constellation` (Linux only; no-op on Docker Desktop).
- Pre-create bind-mount source directories the daemon would otherwise create as root: `nhost/{metadata,migrations,seeds,emails}` for the local and cloud stacks, and the functions `node_modules`/`functions` mountpoints plus a default `functions/tsconfig.json` (matching `services/functions/tsconfig.json`, no-clobber).
- Gave the console service a writable, host-owned `HOME` (`/home/cli`) backed by a bind mount under `.nhost/hasura`, threading `dotNhostFolder` through `console`/`consoleCloud`.
- Ran the `hasura-cli` wrapper container as the host user with `HOME=/tmp` on Linux so exports/squashes land owned by the caller.
- Updated `console` tests for the new signature and added `TestApplyHostUserID`.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["ComposeFileFromConfig"] --> B["prepareNhostFolderSubdirs<br/>(mkdir as host user)"]
  A --> C["applyHostUserID<br/>(user: uid:gid, linux)"]
  A --> D["functions()"]
  D --> E["prepareFunctionsHostFiles<br/>(mountpoints + tsconfig)"]
  C --> F["console / configserver / constellation"]
  G["Docker.HasuraWrapper"] --> H["--user uid:gid, HOME=/tmp"]
```

<details> <summary><h3> File Walkthrough</h3></summary>

  | Relevant files
-- | --
[...]

PR #4778, opened 2026-08-06
fix(cli): Fix broken rootless docker/podman because uid:gid is set

CaspianChaharom:
PR #4538 adds `user: {uid}:{gid}` to the docker-compose on the console/constellation services and the hasura-cli wrapper on Linux hosts. When running with a rootless daemon, that uid gets remapped to a different host user that cannot access the bind-mounted nhost folder, breaking `nhost up`. With the changes here, `docker info` is checked and skips the mapping when the daemon already maps container root to the caller (rootless Docker/Podman, Docker Desktop).

### Checklist

- [x] No breaking changes
- [x] Tests pass


dbarrosop:
Thanks for the PR. To be honest I wasn't a fan of the original changes as I was expecting things to break for someone and I think this exacerbates a bit my concerns. I think the main issue is that the original heuristics were too simplistic and these are fragile and complex and are bound to break for someone else. I would suggest keeping things slightly simpler and more flexible.

1. We add a `--user auto|none|$uid:$gid` flag. `auto` being the default.
2. If the user passes `auto` we run the heuristics (see below).
3. If the user passes `none` we don't set anything.
4. If the user passes some $uid:$gid we just set that

Hopefully this will keep it simple and fine for most use cases and provide an alternative for the rest. Re the heuristics for `auto`, I would suggest something like this:

```go
func dockerEndpoint(ctx context.Context) string {
	if h := os.Getenv("DOCKER_HOST"); h != "" {
		return h
	}
	out, err := exec.CommandContext(ctx, "docker", "context", "inspect",
		"--format", "{{.Endpoints.docker.Host}}").Output()
	if err != nil {
		return "unix:///var/run/docker.sock"
	}
	return strings.TrimSpace(string(out))
}

func autoUser(hostOS string, endpoint string, uid string, gid string) string {
	if hostOS != osLinux {
		return ""
	}
	path, ok := strings.CutPrefix(endpoint, "unix://")
	if !ok {
		return "" // ssh://, tcp://, npipe://
	}
	if path != "/var/run/docker.sock" && path != "/run/docker.sock" {
		return "" // $XDG_RUNTIME_DIR, ~/.docker/desktop, ~/.colima, ...
	}
	return fmt.Sprintf("%d:%d", uid, gid)
}

func somewhere(ctx context.Context) {
   ug := autoUser(runtime.GOOS, dockerEndpoint(ctx), os.Getuid(), os.Getgid())
   ...
}
```

I think this should cover your case in the `auto` mode but even if it doesn't now you have a flag to override the behavior. How does this sound?

/cc @meh as the original implementor of this

CaspianChaharom:
That makes sense. It would definitely work for our use case since there's the manual override, so we'll be able to run it rootless. I implemented and pushed a fix. Let me know what you think.

CaspianChaharom:
@dbarrosop Let me know if there's any feedback

dbarrosop:
Sorry, I think it's mostly fine, if there is anything it should be minor so I will just push the fixes myself and merge, I am hoping today or tomorrow. I just got stuck with a million other things : ( Sorry again for the delay.

dbarrosop (approved):

CaspianChaharom:
Thanks!

dbarrosop:
No worries, thanks for the PR and sorry for the delay. We are waiting to release a new version of the functions service and then we will release the CLI, I would guess mid-next week.

Issue #3961, opened 2026-02-27
dashboard: add a way to refetch the resource version/metadata


dbm03:
**Describe the bug**
The dashboard caches metadata (including its resource version) locally. If the metadata changes externally (e.g. from another tab or user), the cached resource version becomes stale. When the dashboard then tries to update metadata using the outdated version, there's version mismatch error. The resulting error message is unhelpful, and the user has to refresh the page to fix it.

**To Reproduce**
Example with tracking and untracking a table
1. Open the database page in two tabs with the same Nhost project
2. On the first tab, Click on the three dots next to a table
3. Click on "Edit GraphQL"
4. Click on the track/untrack table button
5. Try to do the same in the second tab
6. You should see a resource version mismatch error message

**Expected behavior**
It should either:
1) Catch the error, refetch the metadata in the background, and ask the user to try again in the error message.
2) Or, add a "Refetch metadata version" button inside the error toast.

**Screenshots**
<img width="613" height="102" alt="Image" src="https://github.com/user-attachments/assets/7e925d04-9f9b-42e4-9a5f-c1a9e589bf9b" />

PR #4792, opened 2026-08-10
feat(dashboard): manage logical models and native queries

dbm03:
### Checklist

- [X] No breaking changes
- [X] Tests pass

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

## Goal

Let users create, edit, and delete logical models and native queries from the dashboard's Database section for the project's `default` data source, with unsaved-change protection and metadata writes that use a single local migration in local development and ordinary metadata calls on the platform.

**Type:** Enhancement

## Changes

- Add a browsable Database > Native Queries section: sidebar listing logical models and native queries, detail views, and routes scoped to the `default` data source.
- Add create/edit/delete flows for both object types, including an embedded "create logical model" dialog inside the native query form.
- Add `useLogicalModelMetadataMutation` and `useNativeQueryMetadataMutation`, which write a single migration locally and ordinary metadata on platform, and invalidate the metadata cache after success.
- Warn about dependents before deleting a logical model, and replay select permissions when a logical model is replaced.
- Register the feature in project navigation, command palette, and route state; extract `getGraphQLIdentifierSchema` into `projects/common/utils` for reuse.

## Flow

```mermaid
flowchart LR
  A[Sidebar / detail view] --> B[Create or Edit form]
  B --> C[buildLogicalModelDTO / buildNativeQueryDTO]
  C --> D{useIsPlatform}
  D -- local --> E[single migration]
  D -- platform --> F[ordinary metadata write]
  E & F --> G[invalidate exportMetadata]
```

## Key files

- `.../native-queries/hooks/use{LogicalModel,NativeQuery}MetadataMutation/` (+1708/-0) — metadata writes, migration up/down steps, permission replay
- `.../native-queries/components/Base{LogicalModel,NativeQuery}Form/` (+1012/-0) — shared form bodies, validation schemas, dirty reporting
- `.../native-queries/components/{Create,Edit}{LogicalModel,NativeQuery}Form/` (+1314/-0) — create and edit wrappers
- `.../native-queries/components/NativeQueriesBrowserSidebar/` (+1204/-0) — sidebar and list items
- `.../native-queries/components/` detail and field views (+1640/-0) — details, sections, typed-field editors, empty states
- `.../native-queries/utils/` (+898/-0) — DTO builders, logical-model type converters, dependent lookup
- `src/pages/.../database/native-queries/` (+178/-0) and nav/route registration (+47/-0) — routes and navigation entries
- `28 test files` (+4222/-0 total) — unit, hook, and page-level coverage


dashboard/src/features/orgs/projects/database/dataGrid/components/BaseRelationshipDialog/BaseRelationshipFormTypes.tsx
```diff
@@ -1,4 +1,5 @@
 import { z } from 'zod';
+import { getGraphQLIdentifierSchema } from '@/features/orgs/projects/graphql/common/utils/getGraphQLIdentifierSchema';
```
robertkasza, line 2: This file is in the wrong folder. You should not use things from the grapql feature in the database feature.
Let's move the function some where else. `projects/common` maybe, what do you think?

dashboard/src/features/orgs/projects/database/native-queries/components/BaseLogicalModelForm/BaseLogicalModelForm.tsx
```diff
@@ -0,0 +1,646 @@
+import { zodResolver } from '@hookform/resolvers/zod';
```
robertkasza, line 1: This is quite large file. Let's split it up into smaller files.

dashboard/src/features/orgs/projects/database/native-queries/components/BaseNativeQueryForm/BaseNativeQueryForm.tsx
```diff
@@ -0,0 +1,624 @@
+import { PostgreSQL, sql } from '@codemirror/lang-sql';
```
robertkasza, line 1: I guess the same is true for this. Let's split it up into smaller files

dashboard/src/features/orgs/projects/database/native-queries/components/CreateNativeQueryForm/CreateNativeQueryForm.test.tsx
```diff
@@ -0,0 +1,245 @@
...
+      .mockResolvedValue({ message: 'success' });
+    mocks.logicalModelMutateAsync
+      .mockReset()
+      .mockResolvedValue({ message: 'success' });
+    mocks.router.asPath =
+      '/orgs/test-org/projects/test-app/database/native-queries/default';
+    mocks.router.push.mockReset().mockResolvedValue(true);
+  });
+
+  it('waits for initial metadata without resetting the mounted draft on refetch', async () => {
```
robertkasza, line 126: Why are we refetching during draft? I am not sure that I follow why this test is here.

dashboard/src/features/orgs/projects/database/native-queries/components/CreateNativeQueryForm/CreateNativeQueryForm.tsx
```diff
@@ -0,0 +1,120 @@
...
+    if (!result || router.asPath !== submissionPath) {
+      return;
+    }
+
+    reportDirtyState(false);
+    const { orgSlug, appSubdomain } = router.query;
+    await router.push(
+      `/orgs/${orgSlug}/projects/${appSubdomain}/database/native-queries/${encodeURIComponent(nextValues.source)}/queries/${encodeURIComponent(nextValues.rootFieldName)}`,
+    );
+    onCancel?.();
```
robertkasza, line 102: Why are we calling `onCancel` in the submit handler? That feels like a contradiction.

dashboard/src/features/orgs/projects/database/native-queries/hooks/useGetNativeQueries/useGetNativeQueries.ts
```diff
@@ -0,0 +1,16 @@
...
+} from '@/utils/hasura-api/generated/schemas';
+
+export const selectNativeQueries = (
+  data: ExportMetadataResponse,
+  sourceName: string,
+): NativeQueryItem[] =>
+  data.metadata.sources?.find((source) => source.name === sourceName)
+    ?.native_queries ?? [];
+
+export default function useGetNativeQueries(source: string) {
```
robertkasza, line 14: Why not get the source inside the hook?

dashboard/src/features/orgs/projects/database/native-queries/components/DeleteNativeQueryDialog/DeleteNativeQueryDialog.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const source =
+    typeof dataSourceSlug === 'string' ? dataSourceSlug : 'default';
+  const { mutateAsync: deleteNativeQuery, isPending: isDeletingNativeQuery } =
+    useNativeQueryMetadataMutation({ type: 'delete' });
+  const { data: nativeQueries = [] } = useGetNativeQueries(source);
+  const dependents = query
+    ? getNativeQueryDependents(nativeQueries, query.root_field_name)
+    : [];
+
+  const handleConfirm = async (): Promise<boolean> => {
```
robertkasza, line 53: Why arror function. I believe we usually use function expression

dashboard/src/features/orgs/projects/database/native-queries/components/DeleteNativeQueryDialog/DeleteNativeQueryDialog.tsx
```diff
@@ -0,0 +1,103 @@
...
+    if (!query) {
+      return false;
+    }
+    const submissionPath = router.asPath;
+    const result = await execPromiseWithErrorToast(
+      async () => {
+        await deleteNativeQuery({ source, original: query });
+        if (router.asPath !== submissionPath) {
```
robertkasza, line 61: Can you tell me the scenario where  `router.asPath` and `submissionPath` is not the same. And why are we checking this after the fact?

dashboard/src/features/orgs/projects/database/native-queries/components/EditLogicalModelForm/EditLogicalModelForm.tsx
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    reportDirtyState(false);
+    const { orgSlug, appSubdomain, modelSlug } = router.query;
+    if (modelSlug === model.name && nextValues.name !== model.name) {
+      await router.push(
+        `/orgs/${orgSlug}/projects/${appSubdomain}/database/native-queries/${encodeURIComponent(nextValues.source)}/models/${encodeURIComponent(nextValues.name)}`,
+      );
+    }
+    onCancel?.();
```
robertkasza, line 69: Same calling onCancel in a submit handler, feels wrong.

dashboard/src/features/orgs/projects/database/native-queries/components/LogicalModelDetails/LogicalModelDetails.tsx
```diff
@@ -0,0 +1,177 @@
...
+import { Skeleton } from '@/components/ui/v3/skeleton';
+import { EditLogicalModelForm } from '@/features/orgs/projects/database/native-queries/components/EditLogicalModelForm';
+import { NativeQueriesDetailsHeader } from '@/features/orgs/projects/database/native-queries/components/NativeQueriesDetailsHeader';
+import { NativeQueriesDetailsSection } from '@/features/orgs/projects/database/native-queries/components/NativeQueriesDetailsSection';
+import { NativeQueriesEmptyState } from '@/features/orgs/projects/database/native-queries/components/NativeQueriesEmptyState';
+import { useGetLogicalModels } from '@/features/orgs/projects/database/native-queries/hooks/useGetLogicalModels';
+import { useGetNativeQueries } from '@/features/orgs/projects/database/native-queries/hooks/useGetNativeQueries';
+import type { LogicalModelType } from '@/utils/hasura-api/generated/schemas';
+
+export function formatLogicalModelType(type: LogicalModelType): string {
```
robertkasza, line 15: Does this need to be exported?

dashboard/src/features/orgs/projects/database/native-queries/components/NativeQueriesBrowserSidebar/NativeQueriesBrowserSidebar.test.tsx
```diff
@@ -0,0 +1,747 @@
...
+    expect(
+      screen.queryByRole('button', { name: 'New native query' }),
+    ).not.toBeInTheDocument();
+    expect(
+      screen.queryByRole('button', { name: 'New logical model' }),
+    ).not.toBeInTheDocument();
+  });
+
+  it.each(guardedDrawerSurfaces)(
+    'guards dirty Cancel and preserves the %s draft until discard',
```
robertkasza: The sidebar is guarding a dirty cancel? Could you explain this case to me?

dashboard/src/features/orgs/projects/database/native-queries/components/NativeQueriesBrowserSidebar/NativeQueriesBrowserSidebar.test.tsx
```diff
@@ -0,0 +1,747 @@
...
+      await user.click(
+        within(
+          await screen.findByRole('dialog', { name: 'Unsaved changes' }),
+        ).getByRole('button', { name: 'Discard' }),
+      );
+      await waitFor(() => expect(description).not.toBeInTheDocument());
+    },
+  );
+
+  it('guards a route change while a drawer form is dirty', async () => {
```
robertkasza, line 394: Do we really need this? How can we navigate away when a drawer is open? Could you explain this to me?

dashboard/src/features/orgs/projects/database/native-queries/components/NativeQueriesBrowserSidebar/NativeQueriesBrowserSidebar.test.tsx
```diff
@@ -0,0 +1,747 @@
...
+    await screen.findByText('search_authors');
+    await user.click(screen.getByRole('button', { name: 'New native query' }));
+    expect(screen.getByText('Create native query')).toBeInTheDocument();
+
+    const rootFieldName = await screen.findByLabelText('Root field name');
+    fireEvent.change(rootFieldName, {
+      target: { value: 'search_authors' },
+    });
+    fireEvent.submit(
+      screen.getByRole('button', { name: 'Create' }).closest('form')!,
```
robertkasza, line 627: Why aren't we clicking on the create button instead?

dashboard/src/features/orgs/projects/database/native-queries/components/NativeQueriesBrowserSidebar/NativeQueriesBrowserSidebar.test.tsx
```diff
@@ -0,0 +1,747 @@
...
+    );
+    expect(
+      await screen.findByText(
+        'A native query with this root field name already exists.',
```
robertkasza, line 631: This feels like we are testing the mocks rather then the component itself

dashboard/src/features/orgs/projects/database/native-queries/components/NativeQueriesBrowserSidebar/NativeQueriesBrowserSidebar.test.tsx
```diff
@@ -0,0 +1,747 @@
...
+  it('opens native query creation and submits the form', async () => {
+    const user = new TestUserEvent();
+    render(<NativeQueriesBrowserSidebar />);
+
+    await screen.findByText('search_authors');
+    await user.click(screen.getByRole('button', { name: 'New native query' }));
+    expect(screen.getByText('Create native query')).toBeInTheDocument();
+
+    const rootFieldName = await screen.findByLabelText('Root field name');
+    fireEvent.change(rootFieldName, {
```
robertkasza, line 623: Why are we using user and fireEvent at the same time?

dashboard/src/features/orgs/projects/database/native-queries/components/NativeQueriesBrowserSidebar/NativeQueriesBrowserSidebar.tsx
```diff
@@ -0,0 +1,319 @@
...
+function NativeQueriesBrowserSidebarContent({
+  source,
+}: NativeQueriesBrowserSidebarContentProps) {
+  const modelsResult = useGetLogicalModels(source);
+  const queriesResult = useGetNativeQueries(source);
+  const models = modelsResult.data ?? [];
+  const queries = queriesResult.data ?? [];
+  const { openDrawer } = useDialog();
+  const [searchQuery, setSearchQuery] = useState('');
+  const [modelToDelete, setModelToDelete] = useState<LogicalModelItem | null>(
```
robertkasza, line 72: The sidebar should not have any knowledge about what is about to be deleted.

dashboard/src/features/orgs/projects/database/native-queries/components/NativeQueryDetails/NativeQueryDetails.tsx
```diff
@@ -0,0 +1,149 @@
...
+  if (isLoading || !querySlug) {
+    return (
+      <div className="space-y-4 p-6">
+        <Skeleton className="h-14 w-72" />
+        <Skeleton className="h-64 w-full" />
+      </div>
+    );
+  }
+
+  const query = queries.find((item) => item.root_field_name === querySlug);
```
robertkasza, line 37: After scrolling to the end of this file, I am sure that this files should not filter the queries and display error/loading state. It should just receive the query and display it.

dashboard/src/features/orgs/projects/database/native-queries/components/TypedFieldRow/TypedFieldRow.tsx
```diff
@@ -0,0 +1,93 @@
...
+  nameInputProps: InputProps;
+  descriptionInputProps: ComponentProps<'textarea'>;
+  descriptionValue?: string;
+  nameError?: string;
+  typeEditor: ReactNode;
+  continuationRows?: ReactNode;
+  onRemove: VoidFunction;
+}
+
+export default function TypedFieldRow({
```
robertkasza, line 23: I do not like that the rows are too wide for the drawer. I think it's bad that we cannot see the X(remove) button.

dashboard/src/features/orgs/projects/database/native-queries/hooks/useGetLogicalModels/index.ts
```diff
@@ -0,0 +1,4 @@
+export {
+  default as useGetLogicalModels,
+  selectLogicalModels,
```
robertkasza, line 3: This is only used in the test file. I do not think that we need to export it here.

dashboard/src/features/orgs/projects/database/native-queries/hooks/useGetNativeQueries/index.ts
```diff
@@ -0,0 +1,4 @@
+export {
+  default as useGetNativeQueries,
+  selectNativeQueries,
```
robertkasza, line 3: same for this.

dashboard/src/features/orgs/projects/database/native-queries/hooks/useLogicalModelMetadataMutation/index.ts
```diff
@@ -0,0 +1,6 @@
+export type {
```
robertkasza, line 1: Are these need to be exported? They are only used in the hook, right?

dashboard/src/features/orgs/projects/database/native-queries/hooks/useNativeQueryMetadataMutation/index.ts
```diff
@@ -0,0 +1,6 @@
+export type {
+  NativeQueryMutationArgs,
```
robertkasza, line 2: These are not used outside of this folder. I don't think it is necessary to export them.

dashboard/src/features/orgs/projects/database/native-queries/hooks/useLogicalModelMetadataMutation/useLogicalModelMetadataMutation.ts
```diff
@@ -0,0 +1,166 @@
...
+    unknown,
+    LogicalModelMutationVariables<T>
+  >(
+    async (variables) => {
+      if (!adminApi) {
+        throw new Error('Project metadata connection is unavailable.');
+      }
+
+      const { source } = variables;
+      if (isEmptyValue(source)) {
```
robertkasza, line 72: We are in a mutation. Validating at this point is too late. This should just fire the mutation maybe do some operation on the data.

And I am quite sure that this is a no-op, meaning that we are always comparing to the same data that rendered the UI, even if somebody deleted the source, we would not know.

dashboard/src/features/orgs/projects/database/native-queries/hooks/useLogicalModelMetadataMutation/useLogicalModelMetadataMutation.ts
```diff
@@ -0,0 +1,166 @@
...
+      }
+
+      const base = { adminSecret: adminApi.adminSecret } as const;
+
+      const { data: resourceVersion, error: resourceVersionError } =
+        await refetchResourceVersion();
+      if (resourceVersionError) {
+        throw resourceVersionError;
+      }
+      if (resourceVersion === undefined) {
```
robertkasza, line 86: The same goes for the resource version validation too. It is too late to validate inside the mutation. It should be done before.

dashboard/src/features/orgs/projects/database/native-queries/hooks/useNativeQueryMetadataMutation/useNativeQueryMetadataMutation.ts
```diff
@@ -0,0 +1,165 @@
...
+    NativeQueryMutationResponse,
+    unknown,
+    NativeQueryMutationVariables<T>
+  >(
+    async (variables) => {
+      if (!adminApi) {
+        throw new Error('Project metadata connection is unavailable.');
+      }
+
+      const { source } = variables;
```
robertkasza, line 70: Same as the other comment.

dashboard/src/features/orgs/projects/database/native-queries/utils/createEmptyTypeNode/index.ts
```diff
@@ -0,0 +1 @@
+export { default } from './createEmptyTypeNode';
```
robertkasza, line 1: this is the wrong import

dashboard/src/features/orgs/projects/database/native-queries/utils/formFieldsToLogicalModelFields/index.ts
```diff
@@ -0,0 +1 @@
+export { default } from './formFieldsToLogicalModelFields';
```
robertkasza, line 1: same

dashboard/src/features/orgs/projects/database/native-queries/utils/formTypeToLogicalModelType/index.ts
```diff
@@ -0,0 +1 @@
+export { default } from './formTypeToLogicalModelType';
```
robertkasza, line 1: same

dashboard/src/features/orgs/projects/database/native-queries/utils/logicalModelFieldsToForm/index.ts
```diff
@@ -0,0 +1 @@
+export { default } from './logicalModelFieldsToForm';
```
robertkasza, line 1: same

dashboard/src/features/orgs/projects/database/native-queries/utils/logicalModelTypeToForm/index.ts
```diff
@@ -0,0 +1 @@
+export { default } from './logicalModelTypeToForm';
```
robertkasza, line 1: same

dashboard/src/features/orgs/projects/graphql/common/utils/getGraphQLIdentifierSchema/index.ts
```diff
@@ -0,0 +1 @@
+export { default as getGraphQLIdentifierSchema } from '@/features/orgs/projects/graphql/common/utils/getGraphQLIdentifierSchema/getGraphQLIdentifierSchema';
```
robertkasza, line 1: from './getGraphQLIdentifierSchema'

dashboard/src/pages/orgs/[orgSlug]/projects/[appSubdomain]/database/native-queries/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { NativeQueriesEmptyState } from '@/features/orgs/projects/database/native-queries/components/NativeQueriesEmptyState';
+import { useProject } from '@/features/orgs/projects/hooks/useProject';
+
+export default function NativeQueriesLandingPage() {
+  const router = useRouter();
+  const { orgSlug, appSubdomain } = router.query;
+  const { project } = useProject();
+  const isPlatform = useIsPlatform();
+  const { data: sources = [], isLoading, error } = useGetDataSources();
+  const hasDefault = sources.includes('default');
```
Comment (robertkasza) [inline dashboard/src/pages/orgs/[orgSlug]/projects/[appSubdomain]/database/native-queries/index.tsx:17]: The routing for native queries should be the same or similar to database tables, shouldn't it?

Why are we paying this much attention to sources? Can they change? Is default not always there?

If I understand this correctly in this case the data source is the same for native queries and tables. Why are we adding extra complexity around it when they are the same?

dashboard/src/features/orgs/projects/database/native-queries/components/DeleteNativeQueryDialog/DeleteNativeQueryDialog.tsx
dbm03: I think we have both, or maybe older code has more arrow functions and newer code has more function expressions? I'm going to switch the arrow functions by `function handleConfirm` because I think it's more common in our code, thx

We could look into some Biome rule to enforce one or the other so we don't have two styles for declaring functions (not taking into account differences like hoisting or the use of `this`)?

dashboard/src/features/orgs/projects/database/dataGrid/components/BaseRelationshipDialog/BaseRelationshipFormTypes.tsx
dbm03: I moved it to `projects/common/utils/getGraphQLIdentifierSchema`, thx

dashboard/src/features/orgs/projects/database/native-queries/components/BaseLogicalModelForm/BaseLogicalModelForm.tsx
dbm03: I'm moving most of the types and the validationSchema into a file called `BaseLogicalModelFormTypes.tsx`. I remember you mentioned in the past this isn't perfect because you don't expect to find a validation schema in a file called "...*Types". 

I went with it since we it's the convention pretty much everywhere: same in ServiceFormTypes.ts, BaseActionFormTypes.ts, etc. 

Happy to rename to something like *FormSchema.ts if you'd prefer

dashboard/src/features/orgs/projects/database/native-queries/components/BaseNativeQueryForm/BaseNativeQueryForm.tsx
dbm03: I did the same split as above, thx.

dashboard/src/features/orgs/projects/database/native-queries/components/CreateNativeQueryForm/CreateNativeQueryForm.test.tsx
dbm03: I've removed the async-validation, so this is no longer necessary. It was testing that refetching the metadata dind't delete your draft in the form with default values

I changed the test to check only that it shows the loading skeleton until metadata resolves

dashboard/src/features/orgs/projects/database/native-queries/components/CreateNativeQueryForm/CreateNativeQueryForm.tsx
dbm03: I added `onSubmit` to the forms instead  and now it calls `onSubmit`. i was only passing onCancel by mistake

dashboard/src/features/orgs/projects/database/native-queries/hooks/useGetNativeQueries/useGetNativeQueries.ts
dbm03: I removed multisoure so now it's just 
```ts
export default function useGetNativeQueries() {
  return useExportMetadata(selectNativeQueries);
}
```
Thx

dashboard/src/features/orgs/projects/database/native-queries/components/EditLogicalModelForm/EditLogicalModelForm.tsx
dbm03: now it's calling `onSubmit`, thanks

dashboard/src/features/orgs/projects/database/native-queries/components/LogicalModelDetails/LogicalModelDetails.tsx
dbm03: It wasn't necessary to be exported, however I've moved this into its own utility function folder, `utils/formatLogicalModelType` so I export it from there.
I think that way is cleaner and reduces complexity from LogicalModelDetails.tsx.

dashboard/src/features/orgs/projects/database/native-queries/components/NativeQueriesBrowserSidebar/NativeQueriesBrowserSidebar.test.tsx
dbm03: The sidebar isn't clickable, so I think it's better to just delete the test. You could navigate away when a drawer is open if you press the back navigation arrow in the browser. 
But this is something that happens to any drawer, not just this one so I think it's safe to delete the test

dashboard/src/features/orgs/projects/database/native-queries/components/NativeQueriesBrowserSidebar/NativeQueriesBrowserSidebar.test.tsx
dbm03: I've changed this to `user.click` thx

```ts
    await user.click(screen.getByRole('button', { name: 'Create' }));
```

dashboard/src/features/orgs/projects/database/native-queries/components/NativeQueriesBrowserSidebar/NativeQueriesBrowserSidebar.tsx
dbm03: I'm moving this state inside LogicalModelListItem/NativeQueryListItem. The sidebar has no delete knowledge anymore

dashboard/src/features/orgs/projects/database/native-queries/components/NativeQueryDetails/NativeQueryDetails.tsx
dbm03: I splitted this to NativeQueryOverview and NativeQueryDetailsSkeleton, like the same pattern that we have for ActionDetails in the past for example
Thanks

dashboard/src/features/orgs/projects/database/native-queries/components/TypedFieldRow/TypedFieldRow.tsx
dbm03: I made the columns adapt to the size of the drawer so you can see the X button. Also the ellipsis in the "Type" field works fine if there's a really long data type, like "timestamp without ti..."

I tried with a sticky X button, but I didn't like it so I left just the adaptive size

dashboard/src/features/orgs/projects/database/native-queries/hooks/useGetLogicalModels/index.ts
dbm03: I'm now only exporting this
```
export { default as useGetLogicalModels } from './useGetLogicalModels';
```

dashboard/src/features/orgs/projects/database/native-queries/hooks/useLogicalModelMetadataMutation/useLogicalModelMetadataMutation.ts
dbm03: I'm now passing the `resourceVersion` as a prop, like we do in other hooks like for instance `useSetTableTrackingMutation`.

e.g.
```ts
createLogicalModel({
          ...buildLogicalModelDTO(values),
          resourceVersion: resourceVersion!,
        }),
```


So there's no validation inside the hook anymore, it just sends the operation with the last metadata fetch.

If the resource version is outdated, that's another issue, not for this PR 
https://github.com/nhost/nhost/issues/3961 (dashboard: add a way to refetch the resource version/metadata)

dashboard/src/pages/orgs/[orgSlug]/projects/[appSubdomain]/database/native-queries/index.tsx
Comment (dbm03) [reply dashboard/src/pages/orgs/[orgSlug]/projects/[appSubdomain]/database/native-queries/index.tsx]: I removed entirely the multiple source handling, now I'm just using the literal `'default'`, like what the database tables grid does. Should be much simpler

dashboard/src/features/orgs/projects/graphql/common/utils/getGraphQLIdentifierSchema/index.ts
dbm03: replaced with
```
export { default as getGraphQLIdentifierSchema } from './getGraphQLIdentifierSchema';
```

dashboard/src/features/orgs/projects/database/native-queries/utils/logicalModelTypeToForm/index.ts
dbm03: replaced with `export { default as logicalModelTypeToForm } from './logicalModelTypeToForm';`

dashboard/src/features/orgs/projects/database/native-queries/utils/logicalModelFieldsToForm/index.ts
dbm03: replaced with `export { default as logicalModelFieldsToForm } from './logicalModelFieldsToForm';`

dashboard/src/features/orgs/projects/database/native-queries/utils/formTypeToLogicalModelType/index.ts
dbm03: replaced with `export { default as formTypeToLogicalModelType } from './formTypeToLogicalModelType';`

dashboard/src/features/orgs/projects/database/native-queries/utils/formFieldsToLogicalModelFields/index.ts
dbm03: replaced with `export { default as formFieldsToLogicalModelFields } from './formFieldsToLogicalModelFields';`

dashboard/src/features/orgs/projects/database/native-queries/utils/createEmptyTypeNode/index.ts
dbm03: now it's `export { default as createEmptyTypeNode } from './createEmptyTypeNode';`

dashboard/src/features/orgs/projects/database/native-queries/components/DeleteNativeQueryDialog/DeleteNativeQueryDialog.tsx
dbm03: It was meant to cover that you can't navigate away while the delete is loading, (that's why the check for difference between before and after). But you can't change the page unless you click on the Back browser, since you can't click outside the delete dialog while it's loading. So it wasn't necessary

I'm just checking this now:
```ts
        if (querySlug === query.root_field_name) {
          await router.push(
            `/orgs/${orgSlug}/projects/${appSubdomain}/database/native-queries/default`,
          );
        }
```

dashboard/src/features/orgs/projects/database/native-queries/components/NativeQueriesBrowserSidebar/NativeQueriesBrowserSidebar.test.tsx
dbm03: I had to keep this fireEvent in the file though:

```
function chooseOption(comboboxName: string, optionName: string) {
  fireEvent.keyDown(screen.getByRole('combobox', { name: comboboxName }), {
    key: 'Enter',
  });
  fireEvent.click(screen.getByRole('option', { name: optionName }));
}
```
The combobox inside a drawer doesn't seem to work well with jsdom and the TestUserEvent

dashboard/src/features/orgs/projects/database/native-queries/components/NativeQueriesBrowserSidebar/NativeQueriesBrowserSidebar.test.tsx
dbm03: I'm now only using fireEvent for the `<Select>` elements and one of the submits in #4777 . The rest are all `user.something()`
the Select wouldn't open with user.click because of it being inside a Radix dialog
I looked this issue on Google and I think this is related: https://github.com/radix-ui/primitives/issues/2122

The submit doesn't work either inside a dialog with `user.click`

dashboard/src/features/orgs/projects/database/native-queries/hooks/useGetLogicalModels/index.ts
dbm03: I'm now only exporting this
```
export { default as useGetLogicalModels } from './useGetLogicalModels';
```

dashboard/src/features/orgs/projects/database/native-queries/hooks/useGetNativeQueries/index.ts
dbm03: I'm now exporting
```
export { default as useGetNativeQueries } from './useGetNativeQueries';
```
Thx

dashboard/src/features/orgs/projects/database/native-queries/hooks/useLogicalModelMetadataMutation/index.ts
dbm03: Yeah, I'm now exporting just
```
export { default as useLogicalModelMetadataMutation } from './useLogicalModelMetadataMutation';
```

dashboard/src/features/orgs/projects/database/native-queries/hooks/useNativeQueryMetadataMutation/index.ts
dbm03: I'm now doing
```
export { default as useNativeQueryMetadataMutation } from './useNativeQueryMetadataMutation';
```

dashboard/src/features/orgs/projects/database/native-queries/hooks/useLogicalModelMetadataMutation/useLogicalModelMetadataMutation.ts
dbm03: i'm no longer checking resourceVersion here

dashboard/src/features/orgs/projects/database/native-queries/hooks/useNativeQueryMetadataMutation/useNativeQueryMetadataMutation.ts
dbm03: i'm no longer checking resourceVersion here. i' passing it as a variable instead from outside like the other hooks in the codebase

dashboard/src/features/orgs/projects/database/native-queries/components/NativeQueriesBrowserSidebar/NativeQueriesBrowserSidebar.test.tsx
dbm03: If you refer to the `mutationBodies` assertion part:
```
    expect(mutationBodies[0]?.up[0]?.args?.[1]).toMatchObject({
      type: 'pg_track_logical_model',
      args: { name: 'author_result', description: 'Updated model' },
    });
```

That actually comes from the migration that would be sent to the backend, so it's testing the form submit, the migration builder and that the hook sends it

```
function migrationHandler(
  response: () => Response | Promise<Response> = () =>
    HttpResponse.json({ name: '0_update_native_query_metadata' }),
) {
  return http.post(
    'https://local.hasura.local.nhost.run/apis/migrate',
    async ({ request }) => {
      mutationBodies.push((await request.json()) as RecordedMetadataMigration);
      return response();
    },
  );
}
```

I'm now only using fireEvent for the <Select> elements and one of the submits inside a dialog. That helps with testing more real-like functionality

dashboard/src/features/orgs/projects/database/native-queries/components/NativeQueriesBrowserSidebar/NativeQueriesBrowserSidebar.test.tsx
robertkasza: The combobox inside a drawer doesn't seem to work well with jsdom and the TestUserEvent. 
There should be a fix for that, You should just mock the pointer event. What was the issue?

dashboard/src/features/orgs/projects/database/native-queries/components/DeleteNativeQueryDialog/DeleteNativeQueryDialog.tsx
robertkasza: https://biomejs.dev/linter/rules/use-consistent-function-style/javascript/

There is this rule. But we need to make sure that what I said true. I believe that we are using function declarations(looks like this is the correct term🤷‍♂️). If we have some down time we can look into it.

dashboard/src/features/orgs/projects/database/native-queries/components/NativeQueriesBrowserSidebar/NativeQueriesBrowserSidebar.test.tsx
robertkasza: One more thing. Am I blind or why can't I find any fireEvent int this PR? Or the chooseOption function?

dashboard/src/features/orgs/projects/database/native-queries/components/NativeQueriesBrowserSidebar/NativeQueriesBrowserSidebar.test.tsx
robertkasza: Why fireEvent though? For the other test we are using TestUserEvent. Now I am really lost. Where do you use fireEvents?

robertkasza: Nevermind, found it. Will look into that on the PR.

robertkasza (approved):

dashboard/src/features/orgs/projects/database/native-queries/components/DeleteNativeQueryDialog/DeleteNativeQueryDialog.tsx
dbm03: Absolutely. I think we 99% use function declarations so it'd be better to just pin that rule

Might be worth it to look at this one too?
 https://biomejs.dev/linter/rules/no-nested-ternary/javascript/

PR #4823, opened 2026-08-13
fix(dashboard): prepare shared UI for native queries

dbm03:
- [X] No breaking changes
- [x] Tests pass

## Goal

Prepare the shared dashboard UI primitives that the native queries feature depends on, so each native-queries feature branch does not carry its own changes to common components.

**Type:** Refactoring

## Changes

- `DiscardChangesDialog` stops Escape from propagating to a hosting drawer/dialog by default, with an `onEscapeKeyDown` override for callers that need different behavior.
- `DiscardChangesDialog` restores focus to the element that was focused before it opened instead of relying on Radix's default close-focus target.
- `Combobox` forwards a new optional `onCloseAutoFocus` to its popover content so a caller can control where focus lands after the popover closes.
- `FreeCombobox` accepts an `aria-describedby` prop in its public props type.
- Bumps `lucide-react` to `^0.564.0` and `@radix-ui/react-alert-dialog` to `^1.1.15`.

## Key files

- `dashboard/src/components/common/DiscardChangesDialog/DiscardChangesDialog.tsx` (+33/-1) — Escape isolation and focus restoration
- `dashboard/src/components/common/DiscardChangesDialog/DiscardChangesDialog.test.tsx` (+78/-0) — new test for Escape isolation and focus restore inside a drawer
- `dashboard/src/components/ui/v3/combobox.tsx`, `free-combobox.tsx` (+4/-0) — new optional props
- `dashboard/package.json`, `dashboard/pnpm-lock.yaml` (+6/-115) — dependency bumps

### Test plan

- `pnpm exec vitest run src/components/common/DiscardChangesDialog/DiscardChangesDialog.test.tsx`


dashboard/src/components/common/DiscardChangesDialog/DiscardChangesDialog.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+        onOpenChange={setConfirmationOpen}
+        onDiscardChanges={onDiscardChanges}
+        onEscapeKeyDown={(event) => event.stopPropagation()}
+      />
+    </>
+  );
+}
+
+describe('DiscardChangesDialog', () => {
+  it('isolates Escape and restores focus inside a parent drawer', async () => {
```
robertkasza, line 49: Exactly what logic is that isolates the Escape button?

dbm03: The "isolates Escape" part is only testing that we pass the `onEscapeKeyDown` prop down in this part of the setup above:
```tsx
      <DiscardChangesDialog
        open={confirmationOpen}
        onOpenChange={setConfirmationOpen}
        onDiscardChanges={onDiscardChanges}
        onEscapeKeyDown={(event) => event.stopPropagation()} // <---
      />
```

The point of the test is checking that `onDrawerClose` is never called: without that handler, one Escape can close the confirmation and the MUI drawer behind it, losing everything typed in the form. That was happening before with the embedded "Create logical model" dialog in the create native query form.

robertkasza: What I do not understand is that why this `event.stopPropagation()` logic is part of the `DiscardChangesDialog`. Are there going to be cases where we do not want to stop the escape button? to bubble up to the underlying drawer/dialog?

dbm03: > Are there going to be cases where we do not want to stop the escape button? to bubble up to the underlying drawer/dialog?

I don't think we want it to bubble up. If you press <kbd>Esc</kbd> once, the cancel dialog disappears from the DOM and then you can press <kbd>Esc</kbd> again if you want. I really doubt it'd be a footgun

Best case it avoids people from pressing Esc and closing everything at once (including their form values). Worst case you have to press Esc twice

I moved this logic here to the shared component because of this comment https://github.com/nhost/nhost/pull/4777#discussion_r4037971366

PR #4831, opened 2026-08-14
feat(auth)!: elevate sessions using SMS OTP

skoveit:


services/auth/go/controller/jwt.go
```diff
@@ -500,44 +508,60 @@ func (j *JWTGetter) verifyElevatedClaim(
...
-// email address, so the bypass only applies when none of those methods is
-// usable. When a method is disabled server-wide, a factor only usable through
-// it (a stale active_mfa_type, an email address) must not force an elevation
-// the user has no way to perform — its /elevate endpoint is disabled.
+// keys can still elevate via TOTP MFA, email OTP, or SMS OTP when the method is
+// enabled server-wide and the corresponding user factor is usable, so bypass
+// only applies when none of those methods is usable. When a method is disabled
+// server-wide, a factor only usable through it must not force an elevation the
+// user has no way to perform — its /elevate endpoint is disabled.
 func (j *JWTGetter) canBypassElevation(
```
dbarrosop, line 516: Actually, I am thinking this is no longer necessary. If I am not mistaken after all these changes all users have an elevation mechanism; webauthn, TOTP, SMS OTP, email OTP... So I think this is no longer needed. The elevation being optional makes no sense anymore, does it?

dbarrosop: We will need to change the flag to just be a regular boolean and adapt the backend too if we do this.

dbarrosop:
There seems to be some work here unrelated to elevation, capping attempts and whatnot, can you separate that work from this PR? another PR in the stack if there are dependencies is fine. Thanks!

services/auth/go/controller/jwt.go
skoveit: i don’t think we can make that assumption, since all elevation methods are configurable and can be disabled. also, a user’s factor state can change later (.e.g admin can mark email as unverified)

PR #4844, opened 2026-08-17
feat(dashboard): add search and select for foreign key columns in data grid (#4804)

### LINKED ISSUE #4804: Table (row) editor: provide search and select for foreign key id columns
When creating records with foreign keys to other tables in the web interface, I would expect to be able to search for the related record in place, as is often the case with these kind of editors. 

But in Nhost (also in de Hasura editors) you have to know and paste in the id of the record you want to link to.
This feature can be a huge time saver, and is also present in other similar offerings such as Supabase.

Dakshx07:
Fixes #4804

### Summary
Enhances the record creation and edit forms in the Database DataGrid by providing a searchable `FormFreeCombobox` (`ForeignKeyRecordField`) for foreign key columns, allowing users to search and select referenced records in-place rather than manually finding and copying UUIDs.

---

### 🔍 Changes Made
1. **`ForeignKeyRecordField` Component**:
   - Uses `useTableQuery` to fetch rows from the referenced table (`referencedSchema`, `referencedTable`).
   - Generates smart display labels showing the primary ID alongside secondary descriptive fields (e.g. `1 (Alice)` or `3f2a... (Project A)`) when `name`, `title`, `label`, `email`, `username`, or `slug` columns exist.
   - Powered by `FreeCombobox`, allowing users to either search & select existing rows or type/paste custom UUID values.
   - Includes `NULL` and `DEFAULT` action buttons when columns are nullable or have defaults.

2. **`DatabaseRecordInputGroup` Integration**:
   - Checks `column.foreignKeyRelation && !isArray` to render `ForeignKeyRecordField` instead of standard text inputs.

3. **Unit Tests**:
   - Added `ForeignKeyRecordField.test.tsx` verifying combobox rendering, option selection, NULL toggle, and DEFAULT toggle.
   - Added test in `DatabaseRecordInputGroup.test.tsx` verifying foreign key column rendering.

---

### ✅ Verification
- `pnpm test:vitest DatabaseRecordInputGroup ForeignKeyRecordField NullDefaultToggleField TemporalRecordField TemporalPicker` ➔ **33 / 33 tests passed**
- `tsc --noEmit -p tsconfig.json` ➔ **0 errors**
- `npx @biomejs/biome check` ➔ Clean


ebunders:
hello @Dakshx07 
Great that you implemented this. I looked over the code but am too unfamiliar with this kind of code to have an opinion there.
Question:
am I right to assume that this solution works if the record you want to relate to already exist? I wonder, what if you would want to create the record you want to link to inline along with the record that does the linking? Would you consider that out of scope for this ticket?

cheers, Ernst

dbarrosop:
> I wonder, what if you would want to create the record you want to link to inline along with the record that does the linking? Would you consider that out of scope for this ticket?

Yes, let's treat inline inserts out-of-scope for this PR. That's something that can be evaluated afterwards. Thanks!

dashboard/src/setupTests.ts
```diff
@@ -21,4 +21,31 @@ class ResizeObserverMock {
 // Stub the global ResizeObserver
 vi.stubGlobal('ResizeObserver', ResizeObserverMock);
 
+const storageMock = (() => {
```
robertkasza: Please remove this, this is not needed.

ebunders:
I don't think I can access the preview build https://dashboard-staging-kge3tvfxu-nhost.vercel.app/
?

Dakshx07:
Hi @ebunders 👋 The Vercel staging preview URL is restricted to Nhost team members with internal auth access, but all local tests and typechecks are fully passing.

@robertkasza Branch has been updated with latest `main` and is ready whenever you have a chance to review! 🚀

dbarrosop:
Hello,
thanks for the contribution. We need to look a bit into this and think about it. There are two main issues both related and unrelated to the changes here:

1. We have another PR improving support for composite keys that will basically break the changes here
2. We are unsure about this implementation. This allows people to pick by the foreign key relationship but that might not be what people expects as in most cases that will be an opaque uuid. Similarly we need to figure out which info to show

Anyway, just wanted to mention this is blocked right now by some other work, after we have that one out we will then look a bit into the UX of something like this.

Regards

Dakshx07:
Thanks for the transparent update @dbarrosop! Makes complete sense — handling composite foreign keys and displaying meaningful record identifiers (e.g., name/title alongside UUIDs) will definitely make for a much better user experience.

Happy to pause this for now and adapt/rebase the implementation once the composite keys PR lands. Let me know if there's anything else I can help with in the meantime! 👍

PR #4861, opened 2026-08-20
feat(docs): add Actions to docs

PVUL:
Preview: https://docs-fzvhpf9ch-nhost.vercel.app/products/graphql/actions

---
Adds a docs page for GraphQL Actions: custom queries and mutations backed by webhooks.

Covers concepts, the dashboard and metadata workflows, headers, transforms, permissions, relationships, the custom types editor, deploying, and troubleshooting. Adds a sidebar entry and an Actions card on the GraphQL index, plus dashboard screenshots.

---

### **What this change solves**
Nhost's GraphQL docs had no coverage of Actions, so users had no guidance on extending their GraphQL API with webhook-backed custom queries and mutations. This adds a full Actions reference page — dashboard and metadata-as-code paths, headers, transforms, relationships, permissions, and troubleshooting — and wires it into the GraphQL section navigation. The branch also carries an `fflate` security override and a `packageManager` pin for the functions pnpm example.

___

### **Change Type**
Documentation

___

### **Description**
- Add a new `products/graphql/actions` docs page covering use cases, the request/response flow, creating an Action via the dashboard and via `actions.graphql`/`actions.yaml` metadata, headers, Kriti transforms, viewing/editing/deleting, permissions, relationships, the Custom Types Editor, and a troubleshooting table.
- Register the new page in the Starlight sidebar (`docs/astro.config.mjs`) and surface it as a card on the GraphQL landing page.
- Extend the permissions docs with an "Action Permissions" section (call-or-deny per role, no column/row rules) and cross-link it from the permissions overview.
- Add 14 dashboard screenshots under `docs/public/images/products/graphql/actions/`.
- Pin `services/functions/example-pnpm` to `pnpm@11.24.0` and add a repo-wide `fflate@>=0.7.0 <0.7.5` → `>=0.7.5 <0.8.0` override for GHSA-px8p-9vwx-vf98, with the resulting lockfile updates.

___

<details> <summary><h3> File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody>
<tr><td><strong>Documentation</strong></td><td><details><summary>4 files</summary><table>
<tr>
  <td><strong>docs/src/content/docs/products/graphql/actions.mdx</strong><dd><code>New Actions reference page: flow diagram, handler payload, dashboard and metadata walkthroughs, headers, transforms, relationships, permissions, troubleshooting</code></dd></td>
  <td>+330/-0</td>
</tr>
<tr>
[...]

PR #4864, opened 2026-08-23
chore(cli): update graphql types

PVUL:
Regenerated `cli/nhostclient/graphql/models_gen.go` against the codegen control plane in `cli/gqlgenc.yaml`. It added the `DatabaseRestorePurpose` enum, so the committed file went stale and `go generate ./cli/...` fails the sha1sum check in the `gotests` derivation on any PR that invalidates the CLI cache.

Purely additive: one new enum plus its generated methods, 57 lines, no existing types changed.

Last seen on #4861, which only touches docs mdx, but those are inputs to the CLI derivation via `cli/project.nix`, so a docs change is enough to force a real rebuild and surface the drift.

Verified by generating twice from a clean tree and comparing `git hash-object` on the result, with `golines -w --base-formatter=gofumpt cli` applied after `go generate` to match what the check does before comparing checksums.

### DIFF
--- a/cli/nhostclient/graphql/models_gen.go
+++ b/cli/nhostclient/graphql/models_gen.go
@@ -5956,6 +5956,63 @@ func (e CheckoutStatus) MarshalJSON() ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
+type DatabaseRestorePurpose string
+
+const (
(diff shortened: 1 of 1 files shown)


PVUL:
Closing since these changes went out with: https://github.com/nhost/nhost/pull/4879/changes#diff-fde4132cfeba38a8b61c0b6da388706a05e25aa942a5d841eebf76eca6c02b41

PR #4870, opened 2026-08-24
fix(dashboard): preserve roles and omit the Admin user-id header

dbm03:
### Checklist

- [X] No breaking changes
- [x] Tests pass

Editing the GraphQL playground's Headers tab used to merge into the same state object as the User/Role dropdowns, so a Headers-tab edit could drop the selected role (and vice versa). This change keeps the dropdown selection and Headers-tab overrides as separate state and composes outgoing request headers deterministically, so the selected role survives header edits and the Admin pseudo-user no longer sends a bogus `x-hasura-user-id`.

___

### **Change Type**
Bug fix

___

### **Description**
- Added a `composeRequestHeaders` utility that merges base headers (`content-type`, `x-hasura-admin-secret`), the User/Role selection (`x-hasura-user-id`, `x-hasura-role`), and Headers-tab overrides in precedence order, stringifying/skipping non-string override values. Headers are accumulated in a `Headers` instance (normalizing names to lowercase and trimming values) and returned as a plain object, since GraphiQL spreads the result and graphql-ws JSON-serializes it — a `Headers` instance has no own enumerable properties and would silently reduce to `{}`.
- Added `withRequestHeaders`, a fetcher wrapper that recomposes headers per execution so HTTP requests always carry the current selection plus overrides; WebSocket subscriptions get the composed headers via `connectionParams`.
- Replaced the playground page's single `userHeaders` state with two independent pieces of state: `selection` (user + role) and `headersTabOverrides`, removing the `Record<string, any>` escapes.
- Collapsed `UserAndRoleSelect`'s `onUserChange`/`onRoleChange` callbacks into a single `onSelectionChange({ userId, role })`, keeping the current role when an admin-roles refresh returns the same user and resetting to the first available role when the user actually changes.
- `UserSelect` now emits an empty `userId` for the Admin pseudo-entry instead of the literal string `admin`, so no `x-hasura-user-id` header is sent for Admin.
- Added unit tests for `composeRequestHeaders`/`withRequestHeaders` (including a query-path test pinning that Headers-tab overrides merge with the selection instead of replacing it) and `UserAndRoleSelect`; updated the `UserSelect` test and an e2e helper doc comment.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  US["UserSelect<br/>(Admin => userId '')"] --> UARS["UserAndRoleSelect<br/>onSelectionChange({userId, role})"]
  UARS --> Page["graphql/index.tsx<br/>selection state"]
  HT["GraphiQLEditor Headers tab"] --> Page2["graphql/index.tsx<br/>headersTabOverrides state"]
  Page --> CRH["composeRequestHeaders"]
  Page2 --> CRH
  CRH --> HTTP["withRequestHeaders fetcher<br/>(HTTP queries/mutations)"]
  CRH --> WS["wsClient connectionParams<br/>(subscriptions)"]
```

<tr><td><strong>Bug fix</strong></td><td><details><summary>5 files</summary><table>
<tr>
  <td><strong>composeRequestHeaders/composeRequestHeaders.ts</strong><dd><code>New utility composing base, selection, and Headers-tab headers in precedence order; withRequestHeaders fetcher wrapper</code></dd></td>
  <td>+69/-0</td>
</tr>
<tr>
  <td><strong>composeRequestHeaders/index.ts</strong><dd><code>Barrel re-exports for the new utility and its types</code></dd></td>
  <td>+5/-0</td>
</tr>
<tr>
  <td><strong>graphql/index.tsx</strong><dd><code>Split userHeaders into selection + headersTabOverrides state; per-request header composition and ws connectionParams</code></dd></td>
  <td>+43/-48</td>
</tr>
<tr>
  <td><strong>UserAndRoleSelect/UserAndRoleSelect.tsx</strong><dd><code>Single onSelectionChange callback; keep role on same-user refresh, reset on user change</code></dd></td>
  <td>+24/-23</td>
</tr>
<tr>
  <td><strong>UserSelect/UserSelect.tsx</strong><dd><code>Admin pseudo-entry emits empty userId instead of 'admin'</code></dd></td>
  <td>+4/-3</td>
</tr>
</table></details></td></tr>
<tr><td><strong>Tests</strong></td><td><details><summary>3 files</summary><table>
<tr>
  <td><strong>composeRequestHeaders/composeRequestHeaders.test.ts</strong><dd><code>Unit tests for header composition, Admin selection, override normalization, query-path override merging, and ws subscription routing</code></dd></td>
  <td>+163/-0</td>
</tr>
<tr>
  <td><strong>UserAndRoleSelect/UserAndRoleSelect.test.tsx</strong><dd><code>Tests for role persistence across admin-roles refresh and role reset on user change</code></dd></td>
  <td>+152/-0</td>
</tr>
<tr>
  <td><strong>UserSelect/UserSelect.test.tsx</strong><dd><code>Expect empty userId for the Admin entry</code></dd></td>
  <td>+1/-1</td>
</tr>
</table></details></td></tr>
<tr><td><strong>Documentation</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>e2e/utils.ts</strong><dd><code>Updated the role-selection e2e helper comment for the renamed selection state</code></dd></td>
  <td>+1/-1</td>
</tr>
</table></details></td></tr>
</tbody></table>

</details>

___

### DIFF
--- a/dashboard/src/features/orgs/projects/graphql/common/utils/composeRequestHeaders/composeRequestHeaders.ts
+++ b/dashboard/src/features/orgs/projects/graphql/common/utils/composeRequestHeaders/composeRequestHeaders.ts
@@ -10,7 +10,7 @@ interface ComposeRequestHeadersOptions {
 }
 
 /**
- * Later layers win case-insensitive key collisions; `accept` is deliberately never emitted.
+ * Later layers win case-insensitive key collisions.
  */
 export default function composeRequestHeaders({
   adminSecret,
@@ -21,10 +21,6 @@ export default function composeRequestHeaders({
 
   const setHeader = (name: string, value: string) => {
     const canonicalName = name.toLowerCase();
-
-    if (canonicalName === 'accept') {
-      return;
-    }
 
     headers[canonicalName] = value;
   };
--- a/dashboard/src/features/orgs/projects/graphql/common/utils/composeRequestHeaders/composeRequestHeaders.ts
+++ b/dashboard/src/features/orgs/projects/graphql/common/utils/composeRequestHeaders/composeRequestHeaders.ts
@@ -1,3 +1,5 @@
+import type { Fetcher } from '@graphiql/toolkit';
+
 export type GraphQLPlaygroundSelection = {
   userId: string;
   role: string;
@@ -7,6 +9,12 @@ interface ComposeRequestHeadersOptions {
   adminSecret: string;
   selection: GraphQLPlaygroundSelection;
   headersTabOverrides?: Record<string, unknown>;
+}
+
+interface WithRequestHeadersOptions {
+  fetcher: Fetcher;
+  adminSecret: string;
+  selection: GraphQLPlaygroundSelection;
 }
 
 /**
@@ -37,12 +45,28 @@ export default function composeRequestHeaders({
   }
 
   for (const [name, value] of Object.entries(headersTabOverrides ?? {})) {
-    if (value === null || value === undefined) {
+    if (value == null) {
       continue;
     }
 
-    setHeader(name, typeof value === 'string' ? value : String(value));
+    setHeader(name, String(value));
   }
 
   return headers;
+}
+
+export function withRequestHeaders({
+  fetcher,
+  adminSecret,
+  selection,
+}: WithRequestHeadersOptions): Fetcher {
+  return (graphQLParams, fetcherOpts) =>
+    fetcher(graphQLParams, {
+      ...fetcherOpts,
+      headers: composeRequestHeaders({
+        adminSecret,
+        selection,
+        headersTabOverrides: fetcherOpts?.headers,
+      }),
+    });
 }
--- a/dashboard/src/features/orgs/projects/graphql/common/utils/composeRequestHeaders/composeRequestHeaders.ts
+++ b/dashboard/src/features/orgs/projects/graphql/common/utils/composeRequestHeaders/composeRequestHeaders.ts
@@ -26,23 +26,17 @@ export default function composeRequestHeaders({
   selection,
   headersTabOverrides,
 }: ComposeRequestHeadersOptions): Record<string, string> {
-  const headers: Record<string, string> = {};
+  const headers = new Headers();
 
-  const setHeader = (name: string, value: string) => {
-    const canonicalName = name.toLowerCase();
-
-    headers[canonicalName] = value;
-  };
-
-  setHeader('content-type', 'application/json');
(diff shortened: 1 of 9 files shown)


dashboard/src/features/orgs/projects/graphql/common/utils/composeRequestHeaders/composeRequestHeaders.ts
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Composes base, selection, and Headers-tab headers in precedence order,
+ * normalizing header names to lowercase.
+ */
+export default function composeRequestHeaders({
+  adminSecret,
+  selection,
+  headersTabOverrides,
+}: ComposeRequestHeadersOptions): Record<string, string> {
+  const headers: Record<string, string> = {};
```
robertkasza: Could you tell me why you did not use `Headers`?

dbm03: If we were to send the `Headers` object directly to the graphql fetcher, it doesn't have any enumerable property, and it'd read it as an empty object `{}` (even if it has set headers).

I'll use the `Headers` class inside this composeRequestHeaders method, but I still have to return a Object.fromEntries(headers) to make it work. I'm adding it to the latest commit.

I think it's nicer with Headers as you suggested because that way we no longer need the lowercase normalization (and define the `setHeader`) inside this method, but the return type needs to be something enumerable like `Record<string, string>` regardless

I've also checked and the tests fail if we forget to add this Object.fromEntries(), so there's no risk of breaking this in a refactor

It now looks like this:
```ts
export default function composeRequestHeaders({
  adminSecret,
  selection,
  headersTabOverrides,
}: ComposeRequestHeadersOptions): Record<string, string> {
  const headers = new Headers();

  headers.set('content-type', 'application/json');
  headers.set('x-hasura-admin-secret', adminSecret);

  if (selection.userId) {
    headers.set('x-hasura-user-id', selection.userId);
  }

  if (selection.role) {
    headers.set('x-hasura-role', selection.role);
  }

  for (const [name, value] of Object.entries(headersTabOverrides ?? {})) {
    if (value == null) {
      continue;
    }

    headers.set(name, String(value));
  }

  // GraphiQL spreads these into a plain object and graphql-ws JSON-serializes
  // them, so a Headers instance would silently reduce to {}.
  return Object.fromEntries(headers);
}
```

robertkasza (approved):

PR #4874, opened 2026-08-24
feat(templates): add the react starter template

PVUL:
Adds the `react` starter template for `nhost create`. Stacked on #4873 (react-native), which is stacked on #4865.

### Change Type

Enhancement

### Description

- New `templates/react`: Vite 7 + React 19 + TypeScript SPA, React Router, email OTP sign in, a protected route, and a typed todos list over the GraphQL API with TanStack Query.
- Auth state through a small `AuthProvider` over `@nhost/nhost-js`, no SSR or middleware, this is a pure client side app.
- Backend matches the other templates: one migration plus metadata for `public.todos` with `user` scoped permissions.
- Agent context bundle carried over: `AGENTS.md`, `CLAUDE.md`, `SKILLS.md`, `.mcp.json`, and the four Claude skills. `SKILLS.md` is generated from `.claude/skills/*/SKILL.md`, so the one-file and per-skill forms cannot drift.
- Registered in `cli/cmd/create/registry.go`, listed on the Getting Started page, CLI reference regenerated.
- CI: new `react` job in `templates_checks.yaml` running the codegen drift gate, biome, `tsc --noEmit`, vitest, and `pnpm build`.

### How to test

```sh
go build -o /tmp/nhost-create ./cli
cd $(mktemp -d)
/tmp/nhost-create create my-app --template-path <monorepo>/templates/react

cd my-app/backend && nhost up
cd ../frontend && pnpm dev     # http://localhost:5173, OTP sign in
```

### Checklist

- [x] No breaking changes
- [x] Tests pass
- [x] New features have new tests
- [x] Documentation is updated (if applicable)
- [x] Title of the PR is in the correct format

PR #4873, opened 2026-08-24
feat(templates): add the react-native starter template

PVUL:
Adds the `react-native` starter template for `nhost create`. Stacked on #4865, which adds the command and the first template.

### Change Type

Enhancement

### Description

- New `templates/react-native`: Expo SDK 55 with Expo Router, email OTP sign in, a protected route, and a typed todos list over the GraphQL API.
- Session persistence via `NhostAsyncStorage`, an AsyncStorage backed `SessionStorageBackend` with a write-through in-memory cache, because the Nhost client reads and writes the session synchronously while AsyncStorage is async. Covered by `app/lib/nhost/storage.test.ts`.
- Backend matches the other templates: one migration plus metadata for `public.todos` with `user` scoped permissions.
- Agent context bundle carried over: `AGENTS.md`, `CLAUDE.md`, `SKILLS.md`, `.mcp.json`, and the four Claude skills. `SKILLS.md` is generated from `.claude/skills/*/SKILL.md`, so the one-file and per-skill forms cannot drift.
- Registered in `cli/cmd/create/registry.go`, listed on the Getting Started page, CLI reference regenerated.
- CI: new `react-native` job in `templates_checks.yaml` running the codegen drift gate, biome, `tsc --noEmit`, and vitest. No `pnpm build` step, an Expo release build needs EAS.

### How to test

```sh
go build -o /tmp/nhost-create ./cli
cd $(mktemp -d)
/tmp/nhost-create create my-app --template-path <monorepo>/templates/react-native

cd my-app/backend && nhost up
cd ../frontend && pnpm start   # Expo Go or a simulator, OTP sign in
```

### Checklist

- [x] No breaking changes
- [x] Tests pass
- [x] New features have new tests
- [x] Documentation is updated (if applicable)
- [x] Title of the PR is in the correct format

PR #4865, opened 2026-08-23
feat(cli): add nhost create with starter templates

PVUL:
### **What this change solves**
Bootstrapping an Nhost project means wiring a backend config, a frontend, auth, and GraphQL codegen by hand. This adds `nhost create <name>`, which scaffolds a complete, agent-ready full-stack project from a versioned template in one step.

___

### **Change Type**
Enhancement

___

### **Description**
- New `nhost create` command that scaffolds a generated backend plus a starter frontend into a staging directory that is atomically renamed into place.
- Templates fetched via a shallow, blobless, sparse git checkout of `templates/<name>/` from `nhost/nhost`; repo/ref overridable via hidden flags and env vars. `--template-path` copies from a local directory with symlink refusal and traversal-safe copy.
- Interactive numbered prompts (template, project name, package manager, install confirmation) when in a TTY without `--yes`; fully flag-driven otherwise. `package.json` name patched, pnpm lockfile dropped for npm/bun.
- First template `nextjs-shadcn`: Next.js App Router, Tailwind v4, shadcn/ui todos app with email OTP auth, SSR session handling with middleware-based token refresh, typed GraphQL data layer, vitest tests, biome lint.
- Agent context bundled with the template: `AGENTS.md`, `CLAUDE.md`, `.mcp.json`, and four Claude skills.
- `project.InitProject` extracted from the `init` command for reuse by `create`.
- Docs: new Getting Started page `create-a-project`; `templates/README.md` for contributors.
- CI: new `templates_checks.yaml` (codegen drift gate, lint, build, vitest, scaffold smoke test), `templates` added to allowed PR title packages.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  U["nhost create"] --> R["resolve options<br/>(flags / prompts)"]
  R --> F["fetch template<br/>(git sparse / local path)"]
  F --> S["stage in tmp dir"]
  S --> P["postprocess<br/>(patch name, drop lockfile)"]
  P --> B["InitProject backend"]
  B --> M["atomic rename into place"]
```

<details> <summary><h3> File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody>
<tr><td><strong>CLI (Go)</strong></td><td><details><summary>11 files</summary><table>
<tr><td><strong>create.go</strong><dd><code>Command wiring, option resolution, staging/rename orchestration</code></dd></td><td>+392/-0</td></tr>
[...]

PR #4876, opened 2026-08-24
feat(templates): add the svelte starter template

PVUL:
Adds the `svelte` starter template for `nhost create`. Stacked on #4875 (vue), which is stacked on #4874, #4873, and #4865.

### Change Type

Enhancement

### Description

- New `templates/svelte`: SvelteKit with Svelte 5 runes, email OTP sign in, a protected route, and a typed todos list over the GraphQL API.
- Client-side only, matching the other SPA templates: `src/routes/+layout.ts` sets `ssr = false` and `prerender = false`, and `@sveltejs/adapter-static` builds with an `index.html` fallback.
- Auth state in `src/lib/nhost/auth.ts`: one `nhost` client plus `session`, `user`, and `isAuthenticated` stores, initialised once from the root layout. Protected routes `goto('/signin')` when unauthenticated.
- Backend matches the other templates: one migration plus metadata for `public.todos` with `user` scoped permissions.
- Agent context bundle carried over: `AGENTS.md`, `CLAUDE.md`, `SKILLS.md`, `.mcp.json`, and the four Claude skills. `SKILLS.md` is generated from `.claude/skills/*/SKILL.md`, so the one-file and per-skill forms cannot drift.
- Registered in `cli/cmd/create/registry.go`, listed on the Getting Started page, CLI reference regenerated.
- CI: new `svelte` job in `templates_checks.yaml` running the codegen drift gate, biome, `pnpm check`, vitest, and `pnpm build`.

Two Svelte specific notes:

- The job runs `pnpm check` (svelte-check) rather than `tsc --noEmit`, so components are type-checked and not just modules.
- Biome parses `<script>` blocks but not Svelte markup, so `noUnusedImports` and `noUnusedVariables` are turned off for `*.svelte` in `biome.json`, the same override the vue template uses for `*.vue`. svelte-check covers those cases instead.

### How to test

```sh
go build -o /tmp/nhost-create ./cli
cd $(mktemp -d)
/tmp/nhost-create create my-app --template-path <monorepo>/templates/svelte

cd my-app/backend && nhost up
cd ../frontend && pnpm dev     # http://localhost:5173, OTP sign in
```

### Checklist

- [x] No breaking changes
- [x] Tests pass


PVUL:
Reopening against the corrected base. The parent PR was split into a stack of smaller PRs and GitHub does not allow retargeting a PR that is part of a stack.

PR #4880, opened 2026-08-25
fix(docs): pin mermaid to an exact version

### LINKED ISSUE #4878: docs: sequence diagrams' subheadings are unreadable in light theme
Example: https://docs.nhost.io/products/ai/mcp/authentication#the-flow

Image

whereas in dark theme it's completely readable:
Image

RaghavShubham:
### Checklist

- [x] No breaking changes
- [x] Tests pass

---

The docs load mermaid from `mermaid@11`. jsDelivr resolves that to the newest 11.x at request time, so two visitors can load different builds of the diagram renderer, and rendering can change without any commit to this repo.

This pins the exact version. I checked that `@11` currently serves 11.17.1 and that `mermaid@11.17.1/dist/mermaid.esm.min.mjs` is byte-identical to what `@11` returns right now, so nothing about today's rendering changes — it only stops future 11.x releases landing on readers unannounced.

This came out of #4878, where sequence-diagram text is reported as unreadable in light theme. I could not reproduce that: on a light-theme load of both `products/ai/mcp/authentication` and `products/auth/oauth2-provider/authorization-flow`, participant headings measure 21:1 and message text 12.63:1 against white, both well above the 4.5:1 AA threshold. The floating version is my best explanation for why it reproduces for the reporter and not for me, since mermaid has changed sequence-diagram theme variables within minor releases before.

So this does not claim to fix #4878 — it makes the rendering deterministic so the report can be reproduced or ruled out against a known version. Measurements and details are in my comment on the issue.

### DIFF
--- a/docs/src/components/Head.astro
+++ b/docs/src/components/Head.astro
@@ -47,7 +47,9 @@ if (description) ogUrl.searchParams.set('description', description);
 )}
 
 <script type="module">
-    import mermaid from 'https://cdn.jsdelivr.net/npm/mermaid@11/dist/mermaid.esm.min.mjs';
+    // Exact version: `@11` resolves to the newest 11.x per request, so visitors can get
+    // different builds and diagram rendering drifts without a change here.
+    import mermaid from 'https://cdn.jsdelivr.net/npm/mermaid@11.17.1/dist/mermaid.esm.min.mjs';
 
     function getTheme() {
         // Starlight uses data-theme attribute


dbarrosop:
Thanks for the contribution. AFAICT mermaid.js adheres to semantic versioning so pinning to a major version instead of a specific one should give us the benefits of bugfixes and security patches without breaking anything. Unless we confirm they don't adhere to semantic versioning or that this issue is actually related to differences between different minor versions I think it would be preferable to keep this as is.

As a better alternative, if we could manage this with pnpm where we can leverage the lockfile and standard tooling for updates instead of loading the file directly from the CDN, that would certainly be preferable.

Thanks again, I am closing for now but don't hesitate to re-open if you confirm the issue is related to minor versions differences or to open a new one if you figure out how to manage this with pnpm.

PR #4944, opened 2026-09-04
fix(auth,storage,constellation): handle SIGTERM for graceful shutdown

### LINKED ISSUE #4943: auth, storage, constellation ignore SIGTERM, rolling updates hard-kill in-flight requests
On EKS-style rollouts kubelet sends SIGTERM and waits terminationGracePeriodSeconds before SIGKILL. I checked serve loops in auth, storage and constellation and they all use plain context.WithCancel with no signal handling, so SIGTERM is never observed and the process only exits on SIGKILL.

What I found:
- services/auth/go/cmd/serve.go serve() builds servCtx with context.WithCancel and blocks on <-servCtx.Done(), which only unblocks when ListenAndServe itself fails.
- services/storage/cmd/serve.go has the same pattern, and Shutdown uses a context without timeout handling.
- services/constellation/cmd/serve.go shadows ctx with context.WithCancel and never wires SIGINT/SIGTERM.
- services/mcp/server/server.go already does it right with signal.NotifyContext, so there is an in-repo pattern to copy.

Impact: every deployment/scale-down drops in-flight auth sign-ins, storage uploads and GraphQL queries as 502s, and skips pool.Close / imageTransformer.Shutdown / connector close paths.

Proposed fix: wrap the serve context with signal.NotifyContext for SIGINT/SIGTERM in all three services (same one-line pattern as mcp), keep the existing 30s shutdown timeout.

Ayush7614:
Fixes #4943.

I run multi-tenant workloads on Kubernetes and noticed these three services never observe SIGTERM. kubelet sends SIGTERM on every rollout/scale-down, then SIGKILL after the grace period. Without signal handling the process runs until SIGKILL, killing in-flight auth flows, uploads and GraphQL queries and skipping cleanup (pool.Close, transformer shutdown, connector close).

Change:
- auth, storage, constellation serve loops now use signal.NotifyContext for SIGINT/SIGTERM, same pattern already used in services/mcp/server/server.go.
- storage shutdown now uses a 30s timeout context like auth instead of the parent context.

Verified:
- go vet ./services/auth/go/cmd/... clean.
- storage vet only fails on missing system libvips (pre-existing, unrelated).
- gofmt clean on all three files.
- Compared against mcp positive control.

### DIFF
--- a/services/auth/go/cmd/serve.go
+++ b/services/auth/go/cmd/serve.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"log/slog"
 	"net/http"
+	"os/signal"
+	"syscall"
 	"time"
 
 	"github.com/bradfitz/gomemcache/memcache"
@@ -1602,8 +1604,8 @@ func serve(ctx context.Context, cmd *cli.Command) error {
 	logger.InfoContext(ctx, cmd.Root().Name+" v"+cmd.Root().Version)
 	logFlags(ctx, logger, cmd)
 
-	servCtx, cancel := context.WithCancel(ctx)
-	defer cancel()
+	servCtx, stop := signal.NotifyContext(ctx, syscall.SIGINT, syscall.SIGTERM)
+	defer stop()
 
 	pool, err := getDBPool(ctx, cmd)
 	if err != nil {
@@ -1627,7 +1629,7 @@ func serve(ctx context.Context, cmd *cli.Command) error {
 	}
 
 	go func() {
-		defer cancel()
+		defer stop()
 
 		logger.InfoContext(
 			ctx, "starting server", slog.String("port", cmd.String(flagPort)),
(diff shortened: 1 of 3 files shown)


Ayush7614:
This is ready for review. Could a maintainer please add the safe_to_test label so the checks can run? I verified locally with go vet, go build and gofmt (constellation needs GOEXPERIMENT=jsonv2 per flake.nix).

dbarrosop:
Thanks, I am leaving this one open for now as we are currently doing some refactoring on how services are initialized and I don't want to merge this, forget about it and lose the improvement once we merge the other branch with the refactoring.

Thanks again for the contribution, we will make sure it lands into the repo : )

dbarrosop:
We talked about this internally and I think we would rather merge now and leave the parent issue open until the refactoring we are doing is done. Do you think you could break this PR down into individual ones per service? Thanks!

PR #4946, opened 2026-09-04
fix(auth,storage): set http read/write/idle timeouts

### LINKED ISSUE #4945: auth and storage http.Server sets only ReadHeaderTimeout, slow clients can pin workers
auth and storage build their http.Server with only ReadHeaderTimeout: 5s. There is no ReadTimeout, WriteTimeout or IdleTimeout, so a slow-loris POST on login/signup or a slow reader on a storage download (including CPU-heavy image transforms) can hold a connection and worker indefinitely.

Constellation already sets all four timeouts in newHTTPServer (30s read, 5m write, 120s idle, 5s header cap), so auth/storage are inconsistent with the in-repo standard.

Impact: one noisy tenant can exhaust workers and memory for co-tenants, which is exactly the isolation problem a multi-tenant platform has to prevent.

Proposed fix: add the same Read/Write/Idle timeouts to auth getGoServer and storage getServer.

Ayush7614:
Fixes #4945.

auth and storage only set ReadHeaderTimeout, so slow POSTs and slow download readers can pin workers with no upper bound. Storage is the worst case because downloads can trigger CPU-bound image transforms. Constellation already sets full timeouts in newHTTPServer, this copies the same values (30s read, 5m write, 120s idle) for consistency.

Verified:
- go vet ./services/auth/go/cmd/... clean.
- gofmt clean.
- Values match services/constellation/cmd/serve.go newHTTPServer defaults.

### DIFF
--- a/services/auth/go/cmd/serve.go
+++ b/services/auth/go/cmd/serve.go
@@ -1514,7 +1514,10 @@ func getGoServer(
 	server := &http.Server{ //nolint:exhaustruct
 		Addr:              ":" + cmd.String(flagPort),
 		Handler:           router,
-		ReadHeaderTimeout: 5 * time.Second, //nolint:mnd
+		ReadHeaderTimeout: 5 * time.Second,   //nolint:mnd
+		ReadTimeout:       30 * time.Second,  //nolint:mnd
+		WriteTimeout:      5 * time.Minute,   //nolint:mnd
+		IdleTimeout:       120 * time.Second, //nolint:mnd
 	}
 
 	return server, nil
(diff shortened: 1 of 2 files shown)


Ayush7614:
This is ready for review. Could a maintainer please add the safe_to_test label so the checks can run? Verified locally with go vet, go build and gofmt.

dbarrosop:
Thanks for the contribution, I don't think we want to add this one though although we are still discussing it as part of some other refactoring we are doing so we will see. Current reasoning is we won't go beyond the basic ReadHeaders and IdleTimeout as the rest are better handled at the network layer (i.e. LB, API Gateway, Ingress, whatever)

PR #4948, opened 2026-09-04
feat(mcp): add unauthenticated /healthz for probes

### LINKED ISSUE #4947: mcp server exposes no health endpoint for kube probes
Every other service exposes GET /healthz for liveness/readiness, but services/mcp/server/server.go only registers /, /.well-known/oauth-protected-resource and /.well-known/oauth-authorization-server. There is nothing unauthenticated for kubelet or the monitoring stack to scrape.

Impact: orchestrators cannot gate rollouts or detect a wedged MCP process, so an outage is only visible after customer traffic fails.

Proposed fix: add an unauthenticated GET /healthz returning status ok, placed before the auth middleware like the other services, so liveness/readiness probes and up==0 alerting have something to check.

Ayush7614:
Fixes #4947.

MCP was the only service without a health endpoint, so there was nothing for liveness/readiness probes to check. This adds an unauthenticated GET /healthz returning status ok, registered before the auth middleware so probes do not need credentials.

Verified:
- go test ./services/mcp/... passes (auth, server, tools).
- New TestHealthzHandler covers the endpoint.
- go vet and gofmt clean.


Ayush7614:
This is ready for review. Could a maintainer please add the safe_to_test label so the checks can run? Verified locally: go test ./services/mcp/... passes including the new TestHealthzHandler, plus vet and gofmt clean.

dbarrosop (approved):

dbarrosop:
Thanks for the contribution!

PR #4970, opened 2026-09-08
feat(docs): make the docs machine-readable for LLMs and agents

maxnhost:
Three related additions so an LLM or a coding agent can read Nhost in a few requests instead of scraping 300+ HTML pages. Nothing existing changes: no page, no URL, no navigation. The sitemap still has the same 309 entries.

## What this adds

**`/llms.txt`** (3KB) — the entry point. A short description, six notes on how Nhost actually works, and links to everything below.

**`/llms-full.txt`** (1.7MB) — all 217 docs pages as plain text with code, no chrome.

**`/llms-small.txt`** (644KB) — the same minus the generated reference (25 pages) and the tutorial steps (26 pages), which restate the quickstarts. For smaller context windows.

**`/openapi/auth.json` · `auth.yaml`** (58 paths) and **`/openapi/storage.json` · `storage.yaml`** (11 paths) — the specs that already build our reference pages, now fetchable as files. Same idea as `public/install-mcp.md`.

**`/graphql/cloud.graphql`** and **`/graphql/cloud-with-mutations.graphql`** — the Cloud schema the CLI MCP server embeds, so an agent that isn't running the CLI can read it too.

`gen.sh` produces all six files, and the check job runs `pnpm generate` and sha1sums every file, so they can't drift from `services/auth`, `services/storage` or the CLI schema without failing CI.

## Why the OpenAPI and GraphQL files are here at all

`llms-full.txt` can only contain content collection entries. Our Auth and Storage reference pages are virtual routes injected by the local `starlight-openapi` plugin, so they can't be inlined — confirmed empirically, "Nhost Authentication API" appears zero times in the generated file. Linking the raw specs covers that gap, and a spec is a better source for an agent than scraped HTML anyway.

## On the llms.txt copy

The description and notes follow the docs as written. In particular the Constellation note matches `products/graphql/constellation`: our own engine, Hasura-compatible, ~90% less memory, alpha and opt-in today, running alongside Hasura and not yet the default. Every other claim is sourced the same way — per-user permissions from the Backend MCP Server page, OAuth2/OIDC from the auth provider pages, pgvector from the database extensions page. Happy to reword any of it.

## Two things worth a reviewer's call

1. **The Cloud schema files live under `cli/`.** They're copied, not modified, but someone on the CLI side should confirm serving them from the docs site is fine. They're already in this public repo and shipped inside the CLI binary, so it isn't a new disclosure. `project.nix` now lists them so the sandbox can see them, the same way it already lists the auth and storage specs.
2. **Generated files are committed** (~450KB). `gen.sh` isn't part of the Vercel build, so committing is the only way they ship — same pattern as `src/schemas/` and the typedoc output. An Astro route reading `src/schemas/` directly would avoid the duplication if you'd rather.

## Verification

- `pnpm build` succeeds from a clean `dist`.
- `pnpm test:check-links` passes with zero broken links.
- Biome 2.5.11 clean on `astro.config.mjs`; `bash -n gen.sh` clean.
- Both specs parse as OpenAPI 3.0; each `.json` is byte-equivalent in content to its `.yaml`; all six published files are byte-identical to their sources.
- Sitemap: 309 entries, unchanged. The `llms*` files are not in it.
- `starlight-llms-txt@^0.11.0` matches its peers exactly (Starlight 0.41.3, Astro 7.1.3). Lockfile diff is additions only.
- Content types on the preview deploy: `llms.txt` → `text/plain`, `openapi/*.json` → `application/json`, `openapi/*.yaml` → `text/yaml`, `graphql/*.graphql` → `application/octet-stream`. The linked formats are the first three; `.graphql` has no registered media type, which doesn't affect reading it.

An earlier revision of this description claimed nothing in CI detects a stale generated copy. That was wrong — `nixops/lib/js/js.nix` runs `pnpm generate` and sha1sums everything, and it is what caught the missing `project.nix` entry here.


maxnhost:
@nunopato CI is green now — both failures were mine: docs/project.nix didn't list the Cloud schema files, so gen.sh couldn't see them inside the sandbox, and Biome wanted to reformat the JSON/GraphQL files gen.sh writes into public/. Excluded those instead of reformatting, since reformatting would break the pnpm generate sha1sum check.

PVUL:
this is related to https://github.com/nhost/marketing/issues/27

PVUL (approved):
LGTM

PR #4904, opened 2026-08-31
fix(functions): normalize dependency installation

skoveit:
### Checklist

- [x] No breaking changes
- [x] Tests pass
- [x] New features have new tests
- [x] Documentation is updated (if applicable)
- [x] Title of the PR is in the correct format (see below)

PR #4975, opened 2026-09-09
fix(functions): pin Yarn Classic in shared dependency installer

skoveit:
Follow-up to: https://github.com/nhost/nhost/pull/4904

The shared dependency installer could end up running Yarn Berry even when the pre-install guard passed, because `corepack`/`yarn` resolution can pick a Berry version from a parent manifest or project config. This change pins Yarn Classic at the install call site, widens Berry detection to `devEngines.packageManager`, and fails loudly on an unreadable `package.json` instead of silently guessing the package manager.

___

### **Change Type**
Enhancement

___

### **Description**
- Introduce `NHOST_YARN_CLASSIC_SPEC` (`1.22.22+sha1.…`) and run Yarn installs as `corepack "yarn@$NHOST_YARN_CLASSIC_SPEC" install --frozen-lockfile --ignore-scripts`, so Corepack cannot resolve Berry from a project or parent manifest.
- Extend package-manager detection to fall back to `devEngines.packageManager` (`{name, version}`) when `packageManager` is absent, and update the Berry rejection message accordingly.
- Replace the previous swallow-all `try/catch` in the detection node snippet with a hard failure: an unparseable/unreadable `package.json` now returns a non-zero exit with `refusing to guess the package manager`.
- Rewrite the header and step comments to describe the current defense-in-depth model (Classic pinning is the primary Berry block; the manifest check is an early, clearer error).
- Update the pinned `sha256` parity checksum and the flag-assertion regex in the functions-side test, and add a temp-dir integration test that stubs `corepack`/`yarn` to assert the pinned Classic spec, install flags, and `YARN_IGNORE_PATH=1` when a parent manifest selects `yarn@4.9.1`.

___

  | Relevant files
-- | --
Enhancement | 1 files    nhost-install-deps.shPin Yarn Classic via Corepack, detect devEngines.packageManager, fail on unreadable manifest, rewrite comments   +31/-28 | nhost-install-deps.shPin Yarn Classic via Corepack, detect devEngines.packageManager, fail on unreadable manifest, rewrite comments | +31/-28
nhost-install-deps.shPin Yarn Classic via Corepack, detect devEngines.packageManager, fail on unreadable manifest, rewrite comments | +31/-28
Tests | 1 files    nhost-install-deps.test.jsUpdate pinned parity checksum and Yarn flag assertions; add stubbed end-to-end test for pinned Classic selection   +65/-4 | nhost-install-deps.test.jsUpdate pinned parity checksum and Yarn flag assertions; add stubbed end-to-end test for pinned Classic selection | +65/-4
nhost-install-deps.test.jsUpdate pinned parity checksum and Yarn flag assertions; add stubbed end-to-end test for pinned Classic selection | +65/-4

</details>

___


services/functions/nhost-install-deps.sh
```diff
@@ -53,40 +61,75 @@ nhost_install_deps() {
...
-  if (typeof value === "string") process.stdout.write(value)
-} catch {}
-' "$WORK_DIR/package.json")"
+const raw = require("fs").readFileSync(process.argv[1], "utf8").replace(/^\uFEFF/, "")
+const manifest = JSON.parse(raw)
+if (typeof manifest.packageManager === "string") {
+  process.stdout.write(manifest.packageManager)
+} else {
+  const declared = manifest.devEngines?.packageManager
+  const manager = Array.isArray(declared) ? declared[0] : declared
```
dbarrosop, line 89: wait, this is an array but we only process the first element? what about the rest? don't we need to?

services/functions/nhost-install-deps.sh
```diff
@@ -22,28 +29,29 @@
...
+# Wrapper-runtime versions. services/cd installs these into the user's project
+# (its Lambda wrapper.js requires them). services/functions declares its own
+# express in package.json; a parity test there checks only that the MAJOR
+# matches NHOST_EXPRESS_VERSION, so dev and prod can differ within a major.
+# serverless-http is prod-only (Lambda); dev runs a long-lived express server.
+# Exported: callers read these after sourcing (see services/cd's deploy task).
+export NHOST_EXPRESS_VERSION=5.2.1
+export NHOST_SERVERLESS_HTTP_VERSION=4.0.0
+
+export NHOST_YARN_CLASSIC_SPEC=1.22.22+sha1.ac34549e6aa8e7ead463a7407e1c7390f61a6610
```
dbarrosop, line 44: why are we pinning yarn to this specific version?

skoveit: to prevent corepack from using Yarn Berry. the guard above is just a nice error message but the main protection here is the pinning

services/functions/nhost-install-deps.sh
skoveit: yes, because corepack resolves exactly one package manager per project (even if the array contains only one manager, corepack will ignore it)

services/functions/test/nhost-install-deps.test.js
```diff
@@ -36,14 +46,168 @@ describe('shared install library (parity with nhost/be services/cd)', () => {
...
+      `,
+        'installer-test',
+        SCRIPT,
+      ];
+
+      try {
+        return {
+          status: 0,
+          stdout: execFileSync('sh', argv, opts),
+          stderr: '',
```
dbarrosop: shouldn't we be capturing the stderr so the callers can test it effectively?

dbarrosop (approved):

PR #4995, opened 2026-09-14
feat(landing): add Supabase vs Appwrite comparison blog post

PVUL:
## What this does

Adds a hands-on Supabase vs Appwrite comparison blog post. I built the same small app on both free tiers after Appwrite 2.0 shipped, worked through the dashboards rather than the docs, and wrote down where they actually differ on data models, permissions, auth, storage, functions, and pricing. It ends by positioning Nhost as the option that keeps Postgres at the center with a GraphQL API and role-based permissions.

## Changes

- New post at `landing/src/pages/blog/supabase-vs-appwrite.mdx` plus seven images under `landing/public/images/blog/supabase-vs-appwrite/`.
- Tailwind typography: style inline code in the prose theme and reset it inside fenced code blocks so `pre` blocks are unaffected. This applies to all prose pages, not only this post.
- `next-sitemap.config.js`: pin the lastmod offset to the `Z` form so the generated sitemap does not drift between git versions.
- Regenerated `sitemap-0.xml`. The diff is the new post URL plus a one-time reformat of existing lastmod offsets from `+00:00` to `Z`, which the offset pinning above makes stable going forward.

## Notes

- The post is dated 2026-09-18, a scheduled launch date. Nothing gates future-dated posts on the blog index, so it should not deploy before then or it goes live early as the featured post.


dbarrosop (changes requested):
on hold until further discussion

PR #5005, opened 2026-09-15
feat(cli): add support for the nhost-engine for local dev

meh:
### Checklist

- [x] No breaking changes
- [ ] Tests pass

## What this change solves

The CLI's local development environment did not match the topology the platform
deploys once `experimental.nhost` is enabled: it still ran separate auth and
storage containers, and `nhost software version` reported on components that are
not deployed. This change makes `nhost up` render one bundled `engine` container
in place of auth and storage, makes the version check topology-aware, and adds a
black-box e2e suite that runs identical assertions against both topologies so the
engine's equivalence to the standalone services is a tested claim.

## Change Type

Enhancement

## Description

- `cli/dockercompose` gains an `engine` service that replaces the `storage` and
  `auth` containers when `experimental.nhost` is set. Hasura keeps running; the
  engine's graphql service proxies to it.
- Traefik routes the existing `auth`, `storage` and `graphql` hosts to the
  engine's `/auth`, `/storage` and `/graphql` mount prefixes, so public hostnames
  and client-visible paths do not change and the engine needs no compatibility
  hosts locally.
- Engine environment is built with `appconfig.NhostEngineEnv`, the same builder
  factorio uses in the cloud, with a parity test asserting local and cloud
  variables cannot drift silently.
- `graphql.go` now treats the engine as implying constellation, so constellation
  takes over the public graphql host and hasura keeps only its console routes.
- The `ai` service resolves its dependency and storage URL from the topology, so
  it depends on `engine` and addresses `http://engine:8080/storage/v1` in engine
  mode instead of the `auth` and `storage` services that are no longer rendered.
- **Behaviour change:** `--auth-port` and `--storage-port` are rejected with an
  error when `experimental.nhost` is enabled. The bundled engine serves both
  behind one listener, so the flags cannot be honoured; they are refused rather
  than silently ignored. Standalone mode is unaffected.
- The post-metadata restart step includes the engine when it is the container
  bundling auth and storage, so metadata those services install at boot is
  reapplied after `hasura metadata apply` in both topologies.
- `nhost software version` selects components from config: the engine in place of
  auth and storage, hasura and postgres either way, and stays quiet when the
  cloud publishes no versions for a component. It no longer panics when run
  outside a project directory.
- The generated GraphQL client is regenerated against the engine-aware schema so
  `software_type_enum` includes `Engine`.
- A new `cli/e2e` package boots a real `nhost up` environment and exercises auth,
  storage and GraphQL through the public ingress, behind the `e2e` build tag,
  with failure-output redaction for admin secrets and JWT keys. The suite seeds a
  curated metadata fixture and authenticates as a real user, so a token that the
  bundled auth signs differently from what the GraphQL engine expects fails the
  run in either topology.
- CI runs the suite on CLI pull requests in both `standalone` and `engine` modes;
  engine mode builds and pins a local engine image.

## Diagram Walkthrough

```
standalone:   client ──traefik──> auth        (auth container)
                      ──traefik──> storage     (storage container)
                      ──traefik──> graphql     (hasura)

engine:       client ──traefik──> auth    ──rewrite /auth/$1    ─┐
                      ──traefik──> storage ──rewrite /storage/$1 ─┼─> engine:8080 ──> hasura (graphql)
                      ──traefik──> graphql ──rewrite /graphql/$1 ─┘                └─> postgres / minio
```

### Local environment (Go)

- `cli/dockercompose/engine.go` — <code>New bundled engine service: env via `appconfig.NhostEngineEnv`, traefik mount-prefix ingresses, healthcheck, volumes.</code> +238/-0
- `cli/dockercompose/compose.go` — <code>Render `engine` instead of `storage`/`auth` when the engine is enabled; reject `--auth-port`/`--storage-port` in that mode; extract `hasuraAuthUsable`.</code> +34/-15
- `cli/dockercompose/graphql.go` — <code>Treat engine mode as constellation-enabled for hasura env rendering.</code> +4/-1
- `cli/dockercompose/ai.go` — <code>Resolve the `depends_on` target and storage URL from the topology so `ai` points at the engine when auth and storage are not rendered.</code> +7/-2
- `cli/dockercompose/engine_test.go` — <code>Parity test asserting engine env matches the standalone services, plus ingress, rendering and dependency-target assertions.</code> +411/-0
- `cli/cmd/dev/up.go` — <code>Select the services to restart after applying metadata by topology, including the engine.</code> +12/-17
- `cli/cmd/dev/up_test.go` — <code>Cover the restart selection in both topologies.</code> +47/-0

### Version check (Go)

- `cli/cmd/software/version.go` — <code>Select checked components from config via `servicesToCheck`; skip empty recommendation lists; tolerate a nil config outside a project.</code> +76/-18
- `cli/cmd/software/version_test.go` — <code>Cover engine-replaces-auth/storage selection, the standalone default, the nil config, and the three `checkServiceVersion` output branches.</code> +270/-0

### End-to-end suite (Go, `e2e` tag)

- `cli/e2e/e2e_test.go` — <code>Harness entrypoint: env config, timeout budget, stack boot, curated metadata fixture, and the shared assertions run in both modes.</code> +1319/-0
- `cli/e2e/lifecycle_test.go` — <code>Project init, `nhost up`/`down`, ownership, port and stale-stack reclamation handling.</code> +925/-0
- `cli/e2e/redaction_test.go` — <code>Scrubber for admin secrets and JWTs in failure output, with unit coverage.</code> +419/-0
- `cli/e2e/command_test.go` — <code>Command execution helpers and `envConfig` parsing for the suite, with shell tracing so a failed harness invariant names itself.</code> +279/-0
- `cli/e2e/doc.go` — <code>Package doc recording which files carry the `e2e` tag, and why metadata fixtures must be curated rather than trusted to `hasura-cli`'s exit status.</code> +11/-0

### CI and build

- `.github/workflows/cli_wf_e2e.yaml` — <code>New reusable workflow running the e2e suite in a `standalone`/`engine` matrix, under `contents: read` without persisted credentials, within the aggregate status-check budget, skipped on `release/dashboard` PRs.</code> +120/-0
- `.github/workflows/cli_checks.yaml` — <code>Wire in `test_e2e`; fix stale path filter; add the engine image's source closure to both trigger lists; two TEMP(revert-before-merge) markers.</code> +45/-5
- `cli/project.nix` — <code>`extraCheck` linting and running the Docker-free `e2e`-tagged tests; e2e metadata fixtures in the source fileset.</code> +21/-0
- `go.mod` / `go.sum` — <code>Bump `github.com/nhost/be` for `appconfig.NhostEngineEnv`.</code> +1/-1, +4/-0

### Generated (excluded from review scope)

- `cli/nhostclient/graphql/models_gen.go` +42/-1, `vendor/.../appconfig/nhost_engine.go` +902/-0, `vendor/modules.txt` +1/-1


cli/cmd/software/version.go
```diff
@@ -104,29 +182,16 @@ func CheckVersions(
 		return fmt.Errorf("failed to get nhost client: %w", err)
 	}
 
+	// XXX(meh): Do not publish Engine rows in software_versions until a CLI
+	// release containing that enum value has shipped and been adopted. Older
+	// CLIs reject the entire response when strict enum decoding sees Engine.
 	swv, err := cl.GetSoftwareVersions(ctx)
```
meh, line 185: With `Engine` added to the software versions this is going to lead older versions of the CLI to explode due to `UnmarshalGQL` checking for the validity of the returned names, what's the right way to handle this?

Requiring everyone to instantly upgrade their CLIs for an experimental feature doesn't sound like a good solution.

Changing the enum type to a string doesn't solve older CLIs from exploding and loses out on type safety.

What was done when Graphite was added to the enum?

PR #5006, opened 2026-09-15
fix(functions): add glibc dynamic loader to dev runtime image

skoveit:
The `functions` development-runtime Docker image is built from Nix store paths only, so it has no `/lib64/ld-linux-x86-64.so.2` (or `/lib/ld-linux-*.so.*`). Any dynamically linked ELF binary that is not patched for the Nix store — for example native npm addons, prebuilt binaries downloaded by user projects at install time, or vendored tools — fails to start with a "no such file or directory" exec error. This change adds a tiny derivation that symlinks the glibc dynamic loader into the conventional FHS locations inside the image.

___

### **Change Type**
Bug fix

___

### **Description**
- Adds a `glibcLoader` derivation built with `pkgs.runCommand` that creates `/lib64/<loader>` and `/lib/<loader>` symlinks pointing at `pkgs.stdenv.cc.bintools.dynamicLinker`.
- Includes `glibcLoader` in the `copyToRoot` `buildEnv` paths for `mkDockerImage`, so all three image variants (Node 22 / 24 / 26) get the FHS loader paths.

___

  | Relevant files
-- | --
Build configuration | 1 files    project.nixAdd glibcLoader derivation symlinking the dynamic linker into /lib64 and /lib, and include it in the Docker image root   +8/-0 | project.nixAdd glibcLoader derivation symlinking the dynamic linker into /lib64 and /lib, and include it in the Docker image root | +8/-0
project.nixAdd glibcLoader derivation symlinking the dynamic linker into /lib64 and /lib, and include it in the Docker image root | +8/-0

</details>

___


services/functions/project.nix
```diff
@@ -85,6 +85,13 @@ let
     '';
   };
 
+  glibcLoader = pkgs.runCommand "glibc-loader" { } ''
+    loader="${pkgs.stdenv.cc.bintools.dynamicLinker}"
+    mkdir -p $out/lib64 $out/lib
+    ln -s "$loader" "$out/lib64/$(basename "$loader")"
+    ln -s "$loader" "$out/lib/$(basename "$loader")"
+  '';
+
```
sepples:
```suggestion
```

services/functions/project.nix
```diff
@@ -101,6 +108,7 @@ let
             name = "image";
             paths = [
               pkgs.fakeNss
+              glibcLoader
```
sepples:
```suggestion
              pkgs.bintools
```

sepples:
I'm preparing a pair of PRs to discuss differing approaches to this problem.

dbarrosop (approved):

sepples (approved):

PR #5023, opened 2026-09-17
fix(dashboard): stop unscheduled deployments hiding the live one

dbm03:
### Checklist

- [X] No breaking changes
- [X] Tests pass

## Goal

Deployments that are never scheduled have a null `startedAt`, and `NULLS FIRST` ordering on `DESC` piles them at the top of the deployments list until they push the live deployment off the first page. Order the list by the total, NOT NULL `createdAt` instead, and give the row's time label a matching fallback so row heights stay uniform.

**Type:** Bug fix

## Changes

- Order the deployments list subscription by `createdAt` instead of `startedAt`, so never-scheduled deployments no longer accumulate ahead of deployments that ran.
- Fall back to `createdAt` for the deployment row's relative time label, keeping row heights uniform and the displayed time consistent with the list order.
- Order `getUnifiedDeploymentByCommitSHA` by `createdAt` too. Several unified rows can share a commit SHA, and the old `startedAt: desc` was nulls-first with no tiebreaker among never-scheduled rows, so the serverless function commit badge could resolve to a different deployment between loads. It now deterministically resolves to the newest deployment for that commit.

## Key files

- `dashboard/src/gql/deployments/deployments.graphql` (+2/-2) — ordering for the list subscription and the commit-SHA lookup
- `dashboard/src/features/orgs/projects/deployments/components/DeploymentListItem/DeploymentListItem.tsx` (+3/-2) — row time label falls back to `createdAt`
- `dashboard/src/utils/__generated__/graphql.ts` (+2/-2) — regenerated documents for the ordering changes

## Notes

The commit badge still isn't guaranteed to point at the deployment that *shipped* the function you're viewing — that needs a `status` filter and a decision about what the badge should mean, left for a follow-up. `startedAt: desc_nulls_last` was considered and rejected: it only excludes deployments that never started, so a later deployment that started and then failed would still outrank the one that actually succeeded.


meh (approved):

dashboard/src/gql/deployments/deployments.graphql
```diff
@@ -195,7 +195,7 @@ subscription LatestLiveUnifiedDeploymentSub($appId: uuid!) {
 query getUnifiedDeploymentByCommitSHA($appId: uuid!, $commitSHA: String!) {
   unifiedDeployments(
     where: { appId: { _eq: $appId }, commitSHA: { _eq: $commitSHA } }
-    order_by: { startedAt: desc }
+    order_by: { startedAt: desc_nulls_last }
```
dbarrosop: that's actually going to trade one problem for another as it's going to hide failed deployments that failed to be scheduled. Change the orderby to be by created_at, no other change should be needed.

dbarrosop (changes requested):

dashboard/src/gql/deployments/deployments.graphql
dbm03: I just changed it to 

```gql
query getUnifiedDeploymentByCommitSHA($appId: uuid!, $commitSHA: String!) {
  unifiedDeployments(
    where: { appId: { _eq: $appId }, commitSHA: { _eq: $commitSHA } }
    order_by: { createdAt: desc } # <-----
    limit: 1
  ) {```

dbarrosop (approved):

PR #5028, opened 2026-09-17
fix(constellation): resolve foreign_key_constraint_on against the exact constraint

### LINKED ISSUE #5017: Constellation binds foreign_key_constraint_on: <column> to a composite FK when the same column is in two constraints
I have a table with two foreign keys that share a column. One is a normal single-column FK (org_id → orgs.id). The other is a composite FK ((user_id, org_id) → org_users(id, org_id)).

The object relationship is the Hasura shortcut:

using:
  foreign_key_constraint_on: org_id
Hasura binds that to orgs. Constellation binds it to org_users.

After that, a nested select permission like org.users.user_id fails while the GraphQL roots are built, and the whole Postgres connector shows up as inconsistent. The error looks like:

```
building database connector: failed to create postgres connector for default: failed to create sql connector: failed to build GraphQL roots: failed to initialize table gamemode.session_catches: failed to fix select permission columns for role user: failed to fix permission columns for relationship session: column or relationship not found in table: members not found in table session_members
```

or, with the generic repro below:

users not found in table org_users
What happened: Postgres default-names the composite constraint orders_user_id_org_id_fkey, which sorts before orders_org_id_fkey. Constellation introspects FKs ORDER BY conname and then takes the first row whose column name matches. So the composite key wins even though the metadata only named org_id.

### To Reproduce

Apply this schema:
```
CREATE TABLE public.orgs (
  id uuid PRIMARY KEY DEFAULT gen_random_uuid()
);
CREATE TABLE public.org_users (
  id uuid PRIMARY KEY DEFAULT gen_random_uuid(),
  org_id uuid NOT NULL,
  user_id uuid NOT NULL,
  UNIQUE (id, org_id),
  CONSTRAINT org_users_org_id_fkey
    FOREIGN KEY (org_id) REFERENCES public.orgs (id)
);
CREATE TABLE public.orders (
  id uuid PRIMARY KEY DEFAULT gen_random_uuid(),
  org_id uuid NOT NULL,
  user_id uuid NOT NULL,
  CONSTRAINT orders_org_id_fkey
    FOREIGN KEY (org_id) REFERENCES public.orgs (id),
  CONSTRAINT orders_user_id_fkey
    FOREIGN KEY (user_id) REFERENCES public.org_users (id),
  CONSTRAINT orders_user_id_org_id_fkey
    FOREIGN KEY (user_id, org_id)
    REFERENCES public.org_users (id, org_id)
);
```

Track the tables and add:
- orders.org → foreign_key_constraint_on: org_id
- orgs.users → array rel from org_users.org_id
- user select on orders with filter:
```
filter:
  org:
    users:
      user_id:
        _eq: X-Hasura-User-Id
```
Reload metadata / start Constellation against that database.

Connector for default is inconsistent. Permission walk is looking for users on org_users instead of orgs.

Same metadata works on Hasura.

### Expected behavior

foreign_key_constraint_on: org_id should use the constraint whose local columns are exactly {org_id}, i.e. orders_org_id_fkey → orgs. Nested filters on org.users should then resolve.

Screenshots
N/A. 

Desktop
OS: Windows 11
Version: Constellation current main (reproduced against local Nhost + Hasura metadata)

### Additional context

Workaround that works today: skip the shortcut and use manual_configuration with org_id → id on orgs.

If it helps: the first-match sits in LookupForwardFKTarget (connector/sql/introspection/object.go). Postgres introspection in populateForeignKeys unnests composite FKs per column and orders by conname, with no constraint identity on ForeignKey. pairForwardColumns has the same first-match for joins.

chiliec:
### Checklist

- [x] No breaking changes
- [x] Tests pass

---

Closes #5017

## Goal

Make Constellation resolve `foreign_key_constraint_on` relationships against the exact foreign key constraint—including shared single/composite columns and forward/reverse forms—across PostgreSQL and SQLite. Preserve the legacy first-match fallback while supporting SQLite schema-less and implicit foreign keys.

**Type:** Bug fix

## Changes

- Track foreign key constraint identity through PostgreSQL and SQLite introspection.
- Prefer an exact column-set and target match for forward and reverse relationship resolution, with compatibility fallback.
- Preserve schema-less SQLite relationships and resolve implicit references through ordered primary-key metadata.
- Add regression coverage for shared, composite, reordered, implicit, and backend-specific foreign keys.
- Document the supported metadata forms and constraint-selection behavior.

## Flow

```mermaid
flowchart LR
  A[Database introspection] --> B[Foreign keys grouped by constraint]
  C[Relationship metadata] --> D[Exact column-set and target match]
  B --> D
  D -->|matched| E[Reconcile and build join]
  D -->|no identity or exact match| F[Legacy per-column fallback]
  F --> E
```

## Key files

- `services/constellation/connector/sql/introspection/{object.go,object_test.go}` (+405/-17) — Defines constraint-aware foreign key lookup and its invariants.
- `services/constellation/connector/sql/graphql/{queries,schema}/` (+305/-34) — Applies exact lookup to relationship targets and joins.
- `services/constellation/connector/sql/{postgres,sqlite}/` (+576/-29) — Emits constraint identities, resolves SQLite implicit keys, and adds backend coverage.
- `services/constellation/connector/sql/{reconcile.go,reconcile_internal_test.go}` (+74/-1) — Keeps valid schema-less relationships during reconciliation.
- `services/constellation/docs/user/{hasura-metadata-support.md,postgres-features.md}` (+22/-7) — Documents supported forms and selection semantics.


dbarrosop (approved):

dbarrosop:
Thanks for this, I pushed a few changes and everything should be good now. Merging.

PR #5039, opened 2026-09-17
chore(mcp): publish the MCP service to the official MCP Registry

maxnhost:
## What is the MCP Registry?

[registry.modelcontextprotocol.io](https://registry.modelcontextprotocol.io) is the official, upstream index of MCP servers, run by the Model Context Protocol project itself. It holds **metadata only** — no code, no images. Each entry is a `server.json` that says what the server is called, where its source lives, and how to run it (here: which Docker image, which transport, which env vars).

It matters because it is the source most MCP client and directory surfaces read from, either directly or via downstream aggregators. Today we are not in it. Supabase (`com.supabase/mcp`) and Appwrite are. Anyone browsing for "which backends can my AI assistant talk to" does not currently see Nhost.

Publishing is free, reversible, and costs us one small metadata file plus a CI step.

## What this PR does

Three changes, all additive.

### 1. `services/mcp/server.json` (new)

The registry entry. Published under the name `io.github.nhost/mcp` — the `io.github.<org>` namespace is the one GitHub authentication grants us, and it is currently unclaimed.

It points at `docker.io/nhost/mcp`, declares the streamable-HTTP transport on port 3000 (the path the service actually serves MCP on is `/`, see `server/server.go`), and lists the two env vars `README.md` marks as required: `MCP_GRAPHQL_ENDPOINT` and `MCP_AUTH_URL`.

Validated against the current schema (`2025-12-11`) with `mcp-publisher validate` before opening this PR.

### 2. `services/mcp/project.nix` — image label

The registry will not accept an entry pointing at a Docker image unless it can prove we own that image. For OCI packages the proof is an `io.modelcontextprotocol.server.name` label baked into the image, whose value must equal the `name` in `server.json`. The registry pulls the image and reads it; there is no other ownership check for OCI.

The upstream docs describe this as an "annotation" and show it as a Dockerfile `LABEL`; the registry's validator reads the **image config labels** (`internal/validators/registries/oci.go`), which is what a `LABEL` produces and what nix2container's `config.Labels` produces too. So this goes in the existing `config` passthrough on `nixops-lib.go.docker-image`, which merges over the default config and leaves `Env` and `Entrypoint` untouched.

I built a minimal nix2container image with this exact config to confirm the label lands in the image config JSON that skopeo pushes. It does.

### 3. `.github/workflows/mcp_wf_release.yaml` — `publish-mcp-registry` job

Runs after `push-docker-hub` (the image must exist and be public before the registry can verify it). Installs `mcp-publisher`, rewrites the version and image tag in `server.json` from the release version, validates, logs in, publishes.

Auth is **GitHub OIDC** — `mcp-publisher login github-oidc` exchanges the Actions OIDC token for a registry token. **No new secrets.** The job declares `id-token: write`, the same pattern `wf_docker_push_image.yaml` already uses in this repo.

The committed `server.json` carries `0.2.0` (the current release) so the file is meaningful on its own; CI overwrites both `version` and the image tag with the version actually being released, so they can never drift.

Nothing else in the release path is touched — `build_artifacts`, `push-docker-hub` and `push-docker-ecr` are unchanged, and the new job is a leaf that nothing depends on.

## Reviewer notes

- **The first publish will be the next release, not `0.2.0`.** The `0.2.0` image on Docker Hub was built before this PR and has no labels at all (I checked the published manifest). The registry would reject it. The next tagged `mcp@x.y.z` release builds with the label and publishes cleanly.
- **A failed publish will fail the release run** (red check), though the Docker Hub and ECR pushes will already have completed by then. That seemed better than swallowing the error silently, but say the word if you'd rather it were `continue-on-error`.
- **Description is capped at 100 characters** by the schema, which is why it is terser than our usual copy.
- The registry is still labelled "preview" upstream — breaking changes and data resets are possible.

## Verification

- `mcp-publisher validate services/mcp/server.json` → valid
- `actionlint .github/workflows/mcp_wf_release.yaml` → clean
- `nixfmt --check services/mcp/project.nix` → clean
- `nix eval .#packages.<sys>.mcp-docker-image.drvPath` → evaluates
- Minimal nix2container build → `Labels` present in image config as the registry expects

🤖 Generated with [Claude Code](https://claude.com/claude-code)


maxnhost:
Pushed `f88393e` with both follow-ups.

**1. `mcp-publisher` is pinned.** The install step was pulling `releases/latest`, so the same commit could publish with a different tool over time. Upstream ships a `registry_<version>_checksums.txt` alongside each release that *does* cover the `mcp-publisher_*` assets, so this pins the version **and** verifies the download:

```yaml
env:
  MCP_PUBLISHER_VERSION: "1.8.1"
  MCP_PUBLISHER_SHA256: "a06c9096dcb9727c13555b6be26c7effa707b01f06a4c561ba7a3635443cf2cc"
```

The checksum is asset-specific, so the arch is now fixed at `linux_amd64` rather than derived from `uname` — that's what `ubuntu-latest` is, and hardcoding it is what makes a single checksum meaningful. Bumping the version means bumping the checksum. I verified the download matches upstream's published digest, and that `sha256sum --check --strict` exits non-zero on a wrong digest.

**2. The name is no longer written twice.** `services/mcp/project.nix` now derives the label from `server.json` instead of repeating the string:

```nix
mcpServerName = (builtins.fromJSON (builtins.readFile ./server.json)).name;
```

I went with deriving rather than adding a comparison check to `mcp_checks.yaml`, for three reasons: drift becomes structurally impossible instead of merely detected; `nixops/overlays/pi-agent.nix` already reads a JSON file from Nix this way, so it's an existing repo pattern rather than a new one; and it needs no new CI job — `mcp_checks.yaml` already builds the Docker image on every PR touching `services/mcp/**`, so a missing or malformed `server.json` fails there.

The derivation hash is byte-identical before and after (`66amcky…-image-as-dir.drv`), which confirms this is a pure refactor. I also checked the other direction: temporarily renaming `name` in `server.json` produces a new derivation whose label follows it.

**Re-ran the full check suite on the updated branch** — `mcp-publisher validate` valid, `actionlint` clean, `nixfmt --check` clean, `nix eval` of the image derivation evaluates.

One caveat: because Nix only sees git-tracked files, `server.json` must stay tracked or the Nix eval fails. It is, and the failure would be loud and immediate.

dbarrosop (approved):

